PR #4, opened 2015-01-16
PARQUET-169: Add support for reading nested and repeated data.

nongli:
Before this patch, only flat schemas were supported, that is values with def_level = 1 and rep_level = 0. This patches adds initial support for nested and repeated data. There is still more work to be done in the repeated reconstruction algorithm.

Along with this, this patch adds a generic in memory record abstraction that is intended to only be used for testing and demonstration of the reconstruction algorithm. It is expected that eventually the in memory record format will be Thrift/Avro/Proto/etc. Those record assemblers would be based off of the generic abstraction.


example/compute_stats.cc
```diff
@@ -114,8 +118,8 @@ int main(int argc, char** argv) {
             break;
           }
           case Type::INT32: {
-            int32_t val = reader.GetInt32(&def_level, &rep_level);;
```
julienledem: ;; ?

example/compute_stats.cc
```diff
@@ -126,7 +130,8 @@ int main(int argc, char** argv) {
             break;
           }
           case Type::INT64: {
-            int64_t val = reader.GetInt64(&def_level, &rep_level);;
```
julienledem: ;; ?

example/generic-record-test.cc
```diff
@@ -0,0 +1,45 @@
...
+void TestStruct() {
+  shared_ptr<GenericStruct> r = GenericStruct::Create();
+  const GenericDatum* v;
+
+  v = (GenericDatum*)r->Get(0);
+  assert(v == NULL);
+
+  r->Put(0, Int32Datum::Create(3));
+  v = (GenericDatum*)r->Get(0);
+  printf("%d\n", v->GetInt32());
```
julienledem: should an assert check the return value?

example/parquet-record-reader.cc
```diff
@@ -0,0 +1,120 @@
...
+    vector<InputStream*> streams;
+    vector<const Schema::Element*> projected_cols;
+    vector<const ColumnMetaData*> col_metadata;
+    vector<uint8_t*> col_buffers;
+
+    // Read all the columns we are interested in.
+    for (int c = 0; c < columns.size(); ++c) {
+      int col_idx = columns[c];
+      if (col_idx >= row_group.columns.size()) {
+        cerr << "Invalid col idx." << endl;
```
julienledem: add the column idx in the message?

example/parquet_reader.cc
```diff
@@ -153,43 +156,46 @@ void* read_parquet(char* filename) {
...
+            ByteArray val = reader.GetByteArray(&is_null, &def_level, &rep_level);
+            ByteArray* dst = &((ByteArray*)(column_ptr[c]))[num_values];
+            if (is_null) {
+              dst->ptr = NULL;
+              dst->len = 0;
+            } else {
+              dst->len = val.len;
+              dst->ptr = new uint8_t[val.len];
+              memcpy((char*)dst->ptr, val.ptr, val.len);
+            }
```
julienledem: why the if (is_null) only for bytearray? should this logic (dst->ptr = Null ...) be in reader.GetByteArray ?

example/parquet_reader.cc
```diff
@@ -212,7 +218,7 @@ void* read_parquet(char* filename) {
 
     // prints column name
     for (j = 0; j < row_group.columns.size(); ++j) {
-      char *str = (char*)malloc(50);
+      char* str = (char*)malloc(50);
       assert(str);
```
julienledem: what is this assert for?

src/generic-record.cc
```diff
@@ -0,0 +1,241 @@
...
+      if (!is_null) datum = DoubleDatum::Create(v);
+      break;
+    }
+    case Type::BYTE_ARRAY: {
+      ByteArray v = reader->GetByteArray(&is_null, &def_level, &rep_level);
+      if (!is_null) datum = ByteArrayDatum::Create(v);
+      break;
+    }
+    default:
+      PARQUET_NOT_YET_IMPLEMENTED("Unsupported type");
```
julienledem: add reader->type() to the message?

src/parquet.cc
```diff
@@ -142,6 +143,8 @@ bool ColumnReader::ReadNewPage() {
     const uint8_t* buffer = stream_->Peek(DATA_PAGE_SIZE, &bytes_read);
     if (bytes_read == 0) return false;
     uint32_t header_size = bytes_read;
+
+    // TODO: error handling. This breaks if the page is too big.
```
julienledem: how does it break?

src/parquet/generic-record.h
```diff
@@ -0,0 +1,255 @@
+// Copyright 2012 Cloudera Inc.
```
julienledem: I think we drop this line now that we are Apache.

julienledem:
This looks good to me.
Possibly you could add a few tests.

wesm:
@nongli I've been reading through this. 

- How would you feel about me cannibalizing a few parts of it? For PARQUET-442, I wanted to have a "logical" schema construct in this library, also with a builder API for creating new Parquet files, perhaps not unlike the equivalent code in parquet-mr. I was planning to skim the relevant parquet-mr code and then start building a schema test suite here, and I can use this code as the starting point. 

- Any feelings on preserving the scalar record API you started here? As compared w/ the columnar efforts in (or under way in) Drill and Spark?

nongli:
Feel free to take whatever you'd like. I'm not sure how valuable it is to have a record api here. My latest thinking is decode to columnar and then assemble records from the columnar representation if necessary.

wesm:
Sounds good, I agree -- as soon as you compute array sizes of repeated nodes and nullness for the logical value slots in a columnar layout, reconstructing records is not hard.

julienledem:
should we close this one?

majetideepak:
Yes, we can close this one

PR #7, opened 2015-04-02
PARQUET-232: minor compilation issue

ffabbri4:
I find out some very minor issue when I tried to compile the reader on my environment due to some namespace clashing. 
As example shared_ptr and unordered_map are also in C++11 std namespace. Some compile don't like it.
Also I find that with my test files that I'm reading there was a dereference to a null pointer, if the field is required definition_level_decoder_ is null.


nongli:
+1, merging shortly. Thanks!

PR #9, opened 2015-04-22
PARQUET-259: Support Travis CI in parquet-cpp

kalaxy:
It doesn't appear like there are any tests to run yet, so the build just ensures that compilation succeeds.


kalaxy:
Link to apache jira request: https://issues.apache.org/jira/browse/PARQUET-259

PR #10, opened 2015-04-29
PARQUET-267 - Relax Third Party Dependency Restrictions

kalaxy:
Apache Jira Link: https://issues.apache.org/jira/browse/PARQUET-267

This pull request builds on top of https://github.com/apache/parquet-cpp/pull/9.


nongli:
Can you update the README with some steps on how to get these dependencies you removed? Perhaps just pointing them to the travis config file.

It would be nice to have as few steps as possible between cloning the repo and running the code.

kalaxy:
Sure thing.  I added a few lines indicating how to set up a build env and pointed to the .travis.yml as well.  I also rebased the pull request off of the current master.  Let me know what you think.

wesm:
@kalaxy this is potentially subsumed by #13; I don't have Travis working yet, perhaps we can collaborate.

kalaxy:
@wesm, I checked out your branch and tried compiling.  I didn't get far.  It seems that native toolchain, I'm not sure what that is, is sort of a requirement with your cmake setup.  Setting DISABLE_NATIVE_TOOLCHAIN still doesn't allow various libraries to be found properly.  However, we might be able to come up with a merge between our branches that supports both of our use cases.

kalaxy:
@wesm, A couple more thoughts.  It looks like cmake_modules/toolchain.cmake is setting up a build environment.  This is a bit contrary to what I was trying to accomplish here.  My personal philosophy is that the build environment should not be in committed code.  I think build config should be sufficiently parameterized to allow any build maintainer or developer flexibility to override library locations and versions as desired while defaulting to system locations and versions.  That was my goal with this pull request, to permit build environment configurability.

wesm:
@kalaxy the idea is adding an *optional* portable (across Linux distributions) build environment. I didn't intend to make it the only option, so I will need to fix the build presuming that the user has installed the appropriate system libraries. I think merging our efforts should not be too difficult. 

The end goal here is to make adding more 3rd party dependencies easier as well as building and linking libparquet with other projects sharing a common library toolchain.

kalaxy:
Superseded by https://github.com/apache/parquet-cpp/pull/16.

PR #12, opened 2015-11-06
Fix decoding of boolean columns

edani:
The parquet spec is misleading for "Plain, BOOLEAN" columns: it says: "BOOLEAN: Bit Packed, LSB first" but links to the description of the RLE encoding scheme which doesn't make sense in a PLAIN context.

The Java implementation uses plain bit-packed encoding, that is, one bit per row.  I tested this code change against parquet files produced by it.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -58,7 +58,7 @@ include_directories(SYSTEM ${LZ4_INCLUDE_DIR})
 add_library(lz4static STATIC IMPORTED)
 set_target_properties(lz4static PROPERTIES IMPORTED_LOCATION ${LZ4_STATIC_LIB})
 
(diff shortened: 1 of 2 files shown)


wesm:
See https://issues.apache.org/jira/browse/PARQUET-454 tracking this issue. With PARQUET-437 merged, it would be great to add unit tests that verify this patch. 

I opened a separate JIRA about compiler flags (I see you added `-O3`), let's take care of that in a separate patch: https://issues.apache.org/jira/browse/PARQUET-447

majetideepak:
BitReader supports BIT_PACKED decoding. Can we use close this PR in favour of PARQUET-454 ?

wesm:
When there is a patch for PARQUET-454 you can put "Closes #12" in it

wesm:
@edani or @nongli can you close this, I messed up the regex detecting "closes #12" and the patch #34 didn't close this.

edani:
@wesm  Closing.. Thanks for the fix!

PR #15, opened 2016-01-08
Feature/osx

Mistobaan:
few patches to make it work on macosx


wesm:
See https://issues.apache.org/jira/browse/PARQUET-455

Could you describe what you mean by "make it work"? Were they just compiler warnings or was the build failing ofr you?

wesm:
This is fixed in #40, this PR will be automatically closed when that's merged.

PR #18, opened 2016-01-19
PARQUET-418: Refactored parquet_reader utility for printing file contents.

asandryh:
This pull request contains the following changes:
* Modified parquet_reader utility: refactored, fixed memory leaks, merged compute_stats utility to get rid of code duplication.
* Added a flag --only-stats to parquet_reader to print only file statistics.
* Modified InMemoryInputStream to own its buffer.

All the code repetition still remaining in parquet_reader clearly highlights the need for specialized ColumnReader classes. I will create a new JIRA for this improvement.


src/parquet.cc
```diff
@@ -33,8 +33,20 @@ using parquet::PageType;
...
 
-InMemoryInputStream::InMemoryInputStream(const uint8_t* buffer, int64_t len) :
-  buffer_(buffer), len_(len), offset_(0) {
+InMemoryInputStream::InMemoryInputStream(int64_t len) : len_(len), offset_(0) {
+  buffer_ = (uint8_t*)malloc(sizeof(uint8_t) * len_);
+}
+
+InMemoryInputStream::~InMemoryInputStream() {
+  free(buffer_);
+}
```
wesm: I think it would be preferable to leave the management of this memory to parquet_reader.cc; the intent of this class was to implement InputStream on any contiguous data in virtual address space, so by removing the `buffer` parameter it is no longer able to work with memory regions that are not allocated and freed by the class instance.

example/parquet_reader.cc
```diff
@@ -41,254 +39,227 @@ static string ByteArrayToString(const ByteArray& a) {
...
+        if (col.meta_data.__isset.dictionary_page_offset &&
+            col_start > col.meta_data.dictionary_page_offset) {
+          col_start = col.meta_data.dictionary_page_offset;
         }
-        case Type::BYTE_ARRAY: {
-          ((ByteArray**)column_ptr)[c] =
-              (ByteArray*)malloc(sizeof(ByteArray) * INIT_SIZE);
-          type_array[c] = Type::BYTE_ARRAY;
-          break;
+        std::auto_ptr<InMemoryInputStream> input(
```
wesm: Note `std::auto_ptr` is deprecated in C++11 in favor of `std::unique_ptr`

src/parquet.cc
asandryh: IMHO, it's dangerous to manage the life cycle of a memory buffer outside of the object that's using it. Anyway, instead of modifying InMemoryInputStream, I can add another InputStream that manages its own memory.

example/parquet_reader.cc
asandryh: Right. When we switch to C++11, we'll replace it with `unique_ptr`.

example/parquet_reader.cc
```diff
@@ -41,254 +39,227 @@ static string ByteArrayToString(const ByteArray& a) {
...
+    } else {
+      filename = argv[i];
     }
-
-    return column_ptr;
   }
 
-  fclose(file);
-  return NULL;
+  readParquet(filename, printContents);
```
wesm: It would be great if we could turn this into a `ParquetFileReader::DebugPrint` on top of #20, then the parquet_reader.cc executable would only be a few lines of code.

wesm:
@nongli this patch is non-overlapping with #19 and #20. I suggest we merge all 3 patches once you review and I'll make some follow up patches to continue library-ifying things.

example/parquet_reader.cc
asandryh: It is reasonable to convert parquet_reader.cc into a ParquetFileReader method. For simplicity, let's do this in a separate PR, after this #18 and #20 are completed.

wesm:
lgtm, thank you

asandryh:
@nongli Can you pls take a look at this patch and provide feedback (or pull this into parquet-cpp if you're ok with everything)?

wesm:
@asandryh This patch needs to be rebased on upstream master before it can be merged. There are multiple merge commits here so unfortunately rebasing is a bit tricky

wesm:
In the future you will want to avoid merging master into a branch; in the case of conflict you can rebase.

wesm:
I prepared a rebased commit here

https://github.com/wesm/parquet-cpp/commit/f65424bbc9685c2682f7adb69e0da691247546ad

If this happens in the future, you can use `git merge --squash` to stage the changes in the git index and then create a new commit with only the diff from the original branch. 

```
git merge --squash asandryh/PARQUET-418 
git commit --author="Aliaksei Sandryhaila <aliaksei.sandryhaila@hp.com>" -m "[PARQUET-418]: Added/modified a utility for printing a file contents."
```

wesm:
Please see #23. @nongli I refactored parquet_reader.cc into an instance method of `ParquetFileReader`. Since #23 includes this commit, I suggest for expediency that you review first #22 (PARQUET-433) and then #23 (PARQUET-451 and PARQUET-453)

src/parquet/parquet.h
```diff
@@ -117,6 +117,22 @@ class InMemoryInputStream : public InputStream {
...
+  ScopedInMemoryInputStream(const int64_t len);
+  ~ScopedInMemoryInputStream();
+  uint8_t* data();
+  int64_t size();
+  virtual const uint8_t* Peek(int num_to_peek, int* num_bytes);
+  virtual const uint8_t* Read(int num_to_read, int* num_bytes);
+
+ private:
+  int64_t len_;
+  uint8_t* buffer_;
```
nongli: use std::vector<uint8_t>
and
scoped_ptr<InMemoryInputStream>.

src/parquet/parquet.h
```diff
@@ -117,6 +117,22 @@ class InMemoryInputStream : public InputStream {
   int64_t offset_;
 };
 
+// A wrapper for InMemoryInputStream to manage the memory.
+class ScopedInMemoryInputStream : public InputStream {
+ public:
+  ScopedInMemoryInputStream(const int64_t len);
```
nongli: Comment 'len'

example/parquet_reader.cc
```diff
@@ -41,254 +39,229 @@ static string ByteArrayToString(const ByteArray& a) {
...
 }
 
+string type2String(Type::type t) {
+  switch(t) {
+    case Type::BOOLEAN:
+      return "BOOLEAN";
+      break;
+    case Type::INT32:
+      return "INT32";
+      break;
```
nongli: you have some redundant breaks

nongli:
Can you address the comments in parquet.h/.cc and we can deal with anything else as a follow up?

Also, can you update the title of the PR to be "PARQUET-XXX: ..." (remove "Fixed")

example/parquet_reader.cc
asandryh: You are right, semantically these breaks are redundant. However, some compilers may complain about missing breaks in a switch statement, so it's safer to keep them.

src/parquet/parquet.h
asandryh: This is the length of the buffer allocated by ScopedInMemoryInputStream.
@wesm Since you have incorporate this PR into #23, can you please add a comment to #23?

src/parquet/parquet.h
asandryh: It is redundant. buffer_ is deallocated in the destructor.

asandryh:
@nongli Thank you for the comments!

src/parquet/parquet.h
wesm: I think he is saying to use `std::vector` for memory allocation/deallocation rather than `malloc / free`

asandryh: malloc is perfectly fine in this case. std::vector will call it, internally, too; there is no need to obscure this memory allocation. And we don't need vector or a smart pointer to manage memory deallocation - this is done in ~ScopedInMemoryInputStream() dctor.

wesm: Well, the point is, why re-implement RAII in this class when we already have an STL class available that does this for us?

nongli: I think as a matter of coding style, we should eliminate as many uses of malloc/news as possible. 

We can just remove the dtor, the owernship is self documenting. I believe the code right now is correct but if the construction of the stream in the ctor throws, buffer_ will be leaked. Using RAII, this problem would not exist.

wesm: +1 -- I'm still acclimatizing to modern C++ (and thus have made a couple `new` mistakes, e.g. https://github.com/wesm/parquet-cpp/commit/0714c473f5ea7eaa0f6c5c4055028948951f93b9) so will endeavor to follow these guidelines.

asandryh: That's fine, let's use std::vector<char>.

src/parquet/parquet.h
wesm: Sure, I will do that

src/parquet/parquet.h
wesm: If you can make this change, I will rebase my patches on and get a passing build so we can move along the patch queue

wesm: @nongli I can fix this in my downstream patches since I already have a rebased version of this patch in #23

asandryh: Will do shortly.

wesm:
@asandryh this branch has gotten pretty messed up. here are some commands to help fix it

```
git remote add wesm git@github.com:wesm/parquet-cpp.git
git fetch upstream  # presumes this is apache/parquet-cpp.git
git fetch wesm
git reset --hard upstream/master
git cherry-pick 7279b20   # this is the rebased commit from #23
git push -f
```

wesm:
If you ever screw up, find your last commit in `git reflog` and run `git reset --hard HASH`

asandryh:
@wesm Thanks! I tried to avoid hard reset by squashing my commits, but it didn't work.

wesm:
@asandryh Travis CI builds on the apache org are super slow this week. If you can enable Travis CI for asandryh/parquet-cpp and post verified builds from your branch that will help @nongli verify the commit to merge faster

asandryh:
I've enabled Travis and confirm that this PR builds successfully.

wesm:
OK -- can you make the std::vector change then post a link to the verified build so @nongli can merge? Then I will rebase my patches

CMakeLists.txt
```diff
@@ -133,7 +133,7 @@ message(STATUS "Boost libraries: " ${Boost_LIBRARIES})
 
 # find thrift headers and libs
 find_package(Thrift REQUIRED)
-include_directories(SYSTEM ${THRIFT_INCLUDE_DIR})
+include_directories(SYSTEM ${THRIFT_INCLUDE_DIR} ${THRIFT_INCLUDE_DIR}/thrift)
```
nongli: Why did you change this? Is this to change #include <thrift/xyz> to #include<xyz>

We should namespace include paths as much as possible.

asandryh: This ensures that parquet-cpp compiles with older versions of Thrift.

asandryh:
Successful Travis build: https://travis-ci.org/asandryh/parquet-cpp/builds/105048491

example/example_util.h
```diff
@@ -17,7 +17,28 @@
...
+  }
+  ~InputFile() {
+    if (file != NULL) {
+      fclose(file);
+    }
+  }
+
+  FILE* getFileHandle() { return file; }
+  bool isOpen() { return file != NULL; }
+  std::string getFilename()  { return filename; }
```
nongli: this should be const std::string&

src/parquet.cc
```diff
@@ -48,7 +47,32 @@ const uint8_t* InMemoryInputStream::Read(int num_to_read, int* num_bytes) {
...
+
+const uint8_t* ScopedInMemoryInputStream::Read(int num_to_read,
+                                               int* num_bytes) {
+  return stream_->Read(num_to_read, num_bytes);
+}
+
+
 ColumnReader::~ColumnReader() {
+  delete stream_;
```
nongli: This has unclear ownership semantics. We should make this a unique ptr too. Take a look at this example
http://en.cppreference.com/w/cpp/memory/unique_ptr

nongli:
I'm going to merge this. Please address in follow ups. Thanks!

example/example_util.h
wesm: this is removed in subsequent commits

src/parquet.cc
wesm: thanks, I will fix

PR #19, opened 2016-01-19
PARQUET-437: Add googletest setup and ADD_PARQUET_TEST helper

wesm:
I adapted this functionality from Apache Kudu (incubating). There are no real unit tests, yet, but you can now run `ctest` after building to run all tests that have been created with `ADD_PARQUET_TEST`.


wesm:
I think I got an older version of gtest, let me try updating to 1.7.0.

wesm:
OK, this is now googletest 1.7.0 with the source files manually inlined into gtest-all.cc to simplify compilation.

nongli:
Why are we inlining the source?

wesm:
Here was the original https://github.com/google/googletest/blob/release-1.7.0/src/gtest-all.cc. I can omit gtest-all.cc and instead list out each of the individual sources

wesm:
Either way is fine with me

nongli:
Why are we adding the gtest source into this repo? This seems different than how we manage other third party dependencies.

wesm:
Let me set things up to make googletest the same as the other thirdparty libraries

nongli:
Thanks. I think kudu was particularly fancy and has patched it but I don't think we need that yet. If we do, we can think harder about how to deal with cases like that.

wesm:
Annoyingly googletest is blacklisted in homebrew. Almost have this working

wesm:
@nongli OK, I got the build working. The Linux compiler settings were messing up OS X but I fixed that finally

.travis.yml
```diff
@@ -35,19 +35,29 @@ before_install:
     - >
       if [ $TRAVIS_OS_NAME == osx ]; then
         brew update &&
-        brew install thrift lz4 snappy;
+        brew install thrift lz4 snappy &&
+        cp -r $TRAVIS_BUILD_DIR/thirdparty . &&
+        ./thirdparty/download_thirdparty.sh &&
+        ./thirdparty/build_thirdparty.sh gtest &&
+        export GTEST_HOME=$HOME/build_dir/thirdparty/googletest-release-1.7.0;
```
nongli: can this pull this from thirdparty/versions.sh?

.travis.yml
```diff
@@ -35,19 +35,29 @@ before_install:
...
     - >
       if [ $TRAVIS_OS_NAME == linux ]; then
         cp -r $TRAVIS_BUILD_DIR/thirdparty . &&
         ./thirdparty/download_thirdparty.sh &&
         ./thirdparty/build_thirdparty.sh &&
-        export THRIFT_HOME=$HOME/build_dir/thirdparty/installed SNAPPY_HOME=$HOME/build_dir/thirdparty/installed LZ4_HOME=$HOME/build_dir/thirdparty/installed;
+        export THRIFT_HOME=$HOME/build_dir/thirdparty/installed &&
+        export SNAPPY_HOME=$HOME/build_dir/thirdparty/installed &&
+        export LZ4_HOME=$HOME/build_dir/thirdparty/installed &&
+        export GTEST_HOME=$HOME/build_dir/thirdparty/googletest-release-1.7.0;
```
nongli: same here

setup_build_env.sh
```diff
@@ -18,10 +19,11 @@ if [ "$(uname)" != "Darwin" ]; then
   export THRIFT_HOME=$BUILD_DIR/thirdparty/installed
 fi
 
+export GTEST_HOME=$BUILD_DIR/thirdparty/$GTEST_BASE_DIR
```
nongli: Why isn't this in installed? (just curious)

nongli:
Thanks for doing this. Can you add a quick update to the README and the end of the build step on how to run tests?

.travis.yml
wesm: Yes let me put all this environment setup outside the travis.yml

setup_build_env.sh
wesm: The CMake setup doesn't provide a "make install" target, similarly Homebrew doesn't allow users to install it. I don't have 100% context why that is

wesm:
Sure thing.

wesm:
Something's up with Travis. I pushed commits to address the comments and the builds errored with no logs. Trying a touched version of the last passing build

wesm:
https://www.traviscistatus.com/incidents/fcllblkclgmb

wesm:
Okay, after a few bash mishaps we are back to green. @nongli have a look and lmk if any more comments.

setup_build_env.sh
```diff
@@ -10,6 +10,7 @@ cp -r $SOURCE_DIR/thirdparty $BUILD_DIR
 cd $BUILD_DIR
 ./thirdparty/download_thirdparty.sh
 ./thirdparty/build_thirdparty.sh
+./thirdparty/versions.sh
```
nongli: This doesn't work for me. I think build_thirdparty leaves the working directory in build/.

wesm: looking now

wesm:
@nongli fixed; this was already fixed this in #23 actually

nongli:
How do you run setup_build_env.sh? If i just run it
> ./setup_build_env.sh, it doesn't set *_HOME so cmake . can't find them.

If I source it, it seems to mess up the shell environment (I think -e is set).

wesm:
It should work now

nongli:
Running into this now. Any ideas? The version mismatch is weird.

```
ld: warning: object file (../../build/thirdparty/googletest-release-1.7.0/libgtest.a(gtest-all.cc.o)) was built for newer OSX version (10.10) than being linked (10.9)
Undefined symbols for architecture x86_64:
  "testing::internal::MakeAndRegisterTestInfo(char const*, char const*, char const*, char const*, testing::internal::CodeLocation, void const*, void (*)(), void (*)(), testing::internal::TestFactoryBase*)", referenced from:
      ___cxx_global_var_init in reader-test.cc.o
ld: symbol(s) not found for architecture x86_64
clang: error: linker command failed with exit code 1 (use -v to see invocation)
```

wesm:
I think that's an `-fPIC` issue, let me try to build from scratch on my Mac

wesm:
I was able to compile without making any changes but I made sure that `gtest` is being compiled with `-fPIC` (it was not before, I think that's your linker error). I also removed the `CMAKE_OSX_DEPLOYMENT_TARGET` which was causing the `ld` version warning.

nongli:
@wesm can you remove the brackets in the title. I'll merge this.

wesm:
done

nongli:
@wesm merged. Thanks!

wesm:
thanks!

PR #21, opened 2016-01-21
PARQUET-449: updated to latest parquet.thrift

kashif:


kashif:
@wesm can you kindly have a look?

wesm:
Looks good, thank you! Will leave to @nongli to merge

wesm:
@kashif can you try to fix the build or else roll back to Thrift 0.9.1 (and regenerate the bindings using the 0.9.1 compiler)?

kashif:
yes sure @wesm looking into why it doesn't build on linux...

wesm:
It seems that removing the `make clean` in `build_thirdparty.sh` might fix the issue, if you want to give that a shot. 

https://issues.apache.org/jira/browse/THRIFT-3274

wesm:
Cool that worked!

julienledem:
LGTM. +1
We should figure out a good way to depend on the central thrift version.

nongli:
Can you rename the PR title to "Parquet-449: ..." (No brackets)? This is consistent with parquet-mr and the tool we use to merge enforces this. I think you can just do this in github.

kashif:
@nongli done

nongli:
@kashif Thanks! merged.

wesm:
@julienledem I created a JIRA about this https://issues.apache.org/jira/browse/PARQUET-466

wesm:
I was bummed to find that we now require Thrift 0.9.3 to compile the thrift headers. Perhaps we can add an option to generate them with the active thrift compiler to be less burdensome. However, Thrift 0.9.3 has better cross-platform portability so this will probably be needed to make this library work on MSVC

wesm:
https://issues.apache.org/jira/browse/PARQUET-468

PR #20, opened 2016-01-20
PARQUET-434: Add a ParquetFileReader class

wesm:
Starting point for build out.

PR #22, opened 2016-01-23
PARQUET-433: Add TypedColumnReader template and templatize value decoders

wesm:
Having the physical Parquet types available at compile time opens up many new avenues to simplify the code and make it easier to extend and refactor. Fewer switch statements, and the `type_traits` template pattern (see `parquet/types.h`) lets us associate value types and other stuff with the `parquet::Type` values at compile time. This also makes the decoder classes (especially dictionary encoding) much simpler. We'll reap the same benefits on the encoding / writing paths in the future.

This includes #18, #19, and #20, and I will rebase once those patches are completed and merged.


src/parquet/column_reader.h
```diff
@@ -0,0 +1,175 @@
...
+  // Compression codec to use.
+  std::unique_ptr<Codec> decompressor_;
+  std::vector<uint8_t> decompression_buffer_;
+
+  parquet::PageHeader current_page_header_;
+
+  // Not set if field is required.
+  std::unique_ptr<RleDecoder> definition_level_decoder_;
+  // Not set for flat schemas.
+  std::unique_ptr<RleDecoder> repetition_level_decoder_;
```
majetideepak: Should we have a separate class for the Levels ? We can have a single ReadLevel() for both Repetition and Definition.

wesm: That's out of scope for this patch, but something we could consider

src/parquet/column_reader.h
```diff
@@ -0,0 +1,175 @@
...
+
+ protected:
+  // Reads the next definition and repetition level. Returns true if the value is NULL.
+  bool ReadDefinitionRepetitionLevels(int* def_level, int* rep_level);
+
+  Config config_;
+
+  const parquet::ColumnMetaData* metadata_;
+  const parquet::SchemaElement* schema_;
+  InputStream* stream_;
```
majetideepak: Some helper methods to return the fields inside metadata_(type, num_values) and schema_(type_length, scale, precision) will be  helpful.

src/parquet/column_reader.h
```diff
@@ -0,0 +1,175 @@
...
+    }
+    return true;
+  }
+
+ protected:
+  // Reads the next definition and repetition level. Returns true if the value is NULL.
+  bool ReadDefinitionRepetitionLevels(int* def_level, int* rep_level);
+
+  Config config_;
+
```
majetideepak: add some more fields ? column_index (index of the schema in the schema list), max_definition_level, max_repetition_level ?

src/parquet/column_reader.h
wesm: Let's create some JIRAs about `ColumnReader` API improvements? The goal of this patch is not to make any real functional changes but rather to have 1 ColumnReader type for each Parquet type, with maximum code reuse (i.e. getting rid of as many GetTYPE methods and switch statements as possible).

src/parquet/column_reader.h
wesm: Please see my above comment; this is not in scope for this patch.

src/parquet/column_reader.h
majetideepak: Makes sense. Lets create some more JIRAs for the API improvements.

example/example_util.h
```diff
@@ -17,7 +17,28 @@
 
 #include <string>
 #include <parquet/parquet.h>
+#include <stdio.h>
 
 bool GetFileMetadata(const std::string& path, parquet::FileMetaData* metadata);
 
-#endif
+class InputFile {
```
asandryh: This class is redundant now (it came from pull request #18). It is only used in parquet_reader, though, so it's easy to replace it with LocalFile.

setup_build_env.sh
```diff
@@ -10,6 +10,7 @@ cp -r $SOURCE_DIR/thirdparty $BUILD_DIR
 cd $BUILD_DIR
 ./thirdparty/download_thirdparty.sh
 ./thirdparty/build_thirdparty.sh
+./thirdparty/versions.sh
```
asandryh: Do you mean 'source ./thirdparty/versions.sh'?

setup_build_env.sh
```diff
@@ -18,10 +19,11 @@ if [ "$(uname)" != "Darwin" ]; then
   export THRIFT_HOME=$BUILD_DIR/thirdparty/installed
 fi
 
+export GTEST_HOME=$BUILD_DIR/thirdparty/$GTEST_BASE_DIR
```
asandryh: Is this supposed to be GTEST_BASEDIR?

src/parquet/column_reader.cc
```diff
@@ -243,4 +166,27 @@ bool ColumnReader::ReadNewPage() {
   return true;
 }
 
+ColumnReader* make_column_reader(const parquet::ColumnMetaData* metadata,
```
asandryh: It will be much safer to return std::unique_ptr<ColumnReader> instead of a naked pointer.

example/example_util.h
wesm: Yep, I removed it in #23, both this file and `example_util.cc` are removed there

src/parquet/column_reader.cc
wesm: Good point. How about `std::shared_ptr` (which spares us from having to transfer the pointer from `unique_ptr` to `shared_ptr` in `RowGroupReader::Column` (see #23)

setup_build_env.sh
wesm: Yep, thanks. It is sourced in the other scripts but you'll want `source setup_build_env.sh` to set the right env variables

setup_build_env.sh
wesm: Yep, will fix.

src/parquet/column_reader.cc
asandryh: Sure, shared_ptr is fine, too.

wesm:
For the sake of my sanity (since I already refactored `parquet_reader.cc`) I'm going to address these comments in #23

wesm:
Closing this PR in favor of #23 which has 1 commit per JIRA

PR #23, opened 2016-01-23
PARQUET-451: Add RowGroupReader helper class and refactor parquet_reader.cc into DebugPrint

wesm:
This also addresses PARQUET-433 and PARQUET-453.


src/parquet/reader.cc
```diff
@@ -0,0 +1,392 @@
...
+    std::cout << "Bytes needed: " << col.meta_data.total_compressed_size << std::endl;
+    std::cout << "Bytes read: " << bytes_read << std::endl;
+    throw ParquetException("Unable to read column chunk data");
+  }
+
+  // TODO(wesm): This presumes a flat schema
+  ColumnReader* reader = make_column_reader(&col.meta_data,
+      &this->parent_->metadata_.schema[i + 1], input.release());
+
+  column_readers_[i] = std::shared_ptr<ColumnReader>(reader);
```
asandryh: There is a non-zero chance of memory leak if the program crashes before or during line 121. This is marginally better:
column_readers_[i].reset(make_column_reader(...));

wesm: I'll make `make_column_reader` return a `shared_ptr` then combine these two lines.

wesm:
@asandryh I addressed your reviews in #22 here. @nongli I packaged up the stack of CLs so you can `git merge --ff-only` after we are done reviewing

wesm:
@nongli this has been rebased (the conflicts weren't too bad) and here's a green build

https://travis-ci.org/wesm/parquet-cpp/builds/105058183

I've been hassling Travis CI about the build problems and they are having infrastructure issues that has caused the build queue to stall. I think it's ridiculous, and the ASF is supposed to have 30 concurrent build slaves, so this really should not be happening. 

pls let me know what other code comments you have and we can get this merged. Then @majetideepak can refactor #24 to be compatible with this revamped code structure

asandryh:
+1, LGTM. @nongli

src/parquet/column_reader.cc
```diff
@@ -269,4 +167,36 @@ bool ColumnReader::ReadNewPage() {
   return true;
 }
 
+std::shared_ptr<ColumnReader> make_column_reader(const parquet::ColumnMetaData* metadata,
```
nongli: MakeColumnReader

src/parquet/column_reader.cc
```diff
@@ -269,4 +167,36 @@ bool ColumnReader::ReadNewPage() {
...
+    case Type::BYTE_ARRAY:
+      result = new ByteArrayReader(metadata, element, stream);
+      break;
+    case Type::INT96:
+      result = new Int96Reader(metadata, element, stream);
+      break;
+    default:
+      ParquetException::NYI("type reader not implemented");
+  }
+  return std::shared_ptr<ColumnReader>(result);
```
nongli: make_shared is generally preferred but probably not a big deal here

http://en.cppreference.com/w/cpp/memory/shared_ptr/make_shared

src/parquet/encodings/dictionary-encoding.h
```diff
@@ -143,6 +75,39 @@ class DictionaryDecoder : public Decoder {
...
+inline void DictionaryDecoder<TYPE>::Init(Decoder<TYPE>* dictionary) {
+  int num_dictionary_values = dictionary->values_left();
+  dictionary_.resize(num_dictionary_values);
+  dictionary->Decode(&dictionary_[0], num_dictionary_values);
+}
+
+template <>
+inline void DictionaryDecoder<parquet::Type::BOOLEAN>::Init(
+    Decoder<parquet::Type::BOOLEAN>* dictionary) {
+  // Not implemented
```
nongli: have this throw NYI

src/parquet/encodings/encodings.h
```diff
@@ -24,8 +26,11 @@
 
 namespace parquet_cpp {
 
+template <int TYPE>
```
nongli: Comment here that TYPE is the parquet data type enum.

nongli:
This looks good. 

Can we update reader-test to read all the data files in the repo and verify some results? (as another PR).

wesm:
Opened https://issues.apache.org/jira/browse/PARQUET-475. Will plan to write some more smoke tests to verify the data in those files.

Within short order (next few weeks), we really need to be able to round-trip data to files so that unit tests can generate test data and verify it can be read and written successfully.

src/parquet/column_reader.cc
wesm: Done. What do you think about making this a static method like `ColumnReader::Make` ?

wesm:
@nongli thank you, good to go

majetideepak:
I closed my pull request to avoid complicated rebasing with the current re-factored code.
Once #23 get pushed, I will submit a new pull request for PARQUET-428

src/parquet/column_reader.cc
nongli: Yea, that would be better.

wesm: ok done.

nongli:
Merged. Thanks!

wesm:
thank you sir!

PR #24, opened 2016-01-23
PARQUET-428: Support INT96 and FIXED_LEN_BYTE_ARRAY types

majetideepak:
Added support for INT96 and FIXED_LEN_BYTE_ARRAY.
Will add examples to test this in another JIRA.


wesm:
Cool!

Depending on the feedback / code review on #22 / PARQUET-433, how would you like to handle refactoring this to sit on top of templated decoders and column readers? I am happy to do it if that is easiest.

Note: please try to avoid merge commits in feature branches in the future (rebase instead).

src/parquet/parquet.h
```diff
@@ -226,6 +245,16 @@ inline ByteArray ColumnReader::GetByteArray(int* def_level, int* rep_level) {
...
 }
 
+inline FixedLenByteArray ColumnReader::GetFixedLenByteArray(int* def_level, int* rep_level) {
+  if (ReadDefinitionRepetitionLevels(def_level, rep_level)) return FixedLenByteArray();
+  if (buffered_values_offset_ == num_decoded_values_) BatchDecode();
+  return reinterpret_cast<FixedLenByteArray*>(&values_buffer_[0])[buffered_values_offset_++];
+}
+
+inline int ColumnReader::GetFixedLengthSize(){
+	return fixed_length_size;
```
wesm: Indentation uses 2 spaces

src/parquet/parquet.h
```diff
@@ -61,6 +61,14 @@ struct ByteArray {
   const uint8_t* ptr;
 };
 
+struct FixedLenByteArray {
+  const uint8_t* ptr;
+};
+
+struct Int96 {
+  uint32_t val[3];
+};
```
wesm: How do you want to deal with struct padding / portability?

wesm:
You have several code style issues that would be caught by cpplint, run `make lint` to see them. Once #20 is merged, cpplint errors will cause the build to fail.

majetideepak:
I just fixed the formatting issues. Thanks! for the review.

src/parquet/encodings/plain-encoding.h
```diff
@@ -52,6 +52,19 @@ class PlainDecoder : public Decoder {
     return GetValues(buffer, max_values, sizeof(int64_t));
   }
 
+  virtual int GetInt96(Int96* buffer, int max_values) {
+    max_values = std::min(max_values, num_values_);
+    int size = 3*sizeof(uint32_t);//Struct Int96 could be padded
```
majetideepak: I am not depending upon the sizeof(Int96) inside the decoder.
I am using the explicit size.

majetideepak:
I can merge this code after the review of #22. I am waiting for it to merge to add more test cases as well.

src/parquet/encodings/plain-encoding.h
wesm: Yes, but it would be convenient if `sizeof(Int96)` were always 12. I don't know the performance implications for unaligned memory access. This would be especially useful a template context, but it's easy to specialize the templates for `INT96`, too.

CMakeLists.txt
```diff
@@ -79,6 +79,7 @@ message(STATUS "Boost libraries: " ${Boost_LIBRARIES})
 # find thrift headers and libs
 find_package(Thrift REQUIRED)
 include_directories(SYSTEM ${THRIFT_INCLUDE_DIR})
+include_directories(SYSTEM ${THRIFT_INCLUDE_DIR}/thrift)
```
majetideepak: required for thrift versions < 0.9.0

src/parquet/encodings/plain-encoding.h
wesm: If we decide to portably pack `Int96` we can use this code from flatbuffers

https://github.com/google/flatbuffers/blob/master/include/flatbuffers/flatbuffers.h#L1355

CMakeLists.txt
wesm: This got added in #18 too =|

wesm:
If you could give some feedback on #22 (esp. in the context of adding these last types) that would be a big help.

src/parquet/encodings/encodings.h
```diff
@@ -62,12 +68,13 @@ class Decoder {
...
  protected:
-  Decoder(const parquet::Type::type& type, const parquet::Encoding::type& encoding)
-    : type_(type), encoding_(encoding), num_values_(0) {}
+  Decoder(const parquet::Type::type& type, const parquet::Encoding::type& encoding,int fixed_length)
+    : type_(type), encoding_(encoding), num_values_(0), fixed_length_size(fixed_length) {}
 
   const parquet::Type::type type_;
   const parquet::Encoding::type encoding_;
   int num_values_;
+  int fixed_length_size;
```
wesm: Let's revisit after #22 is reviewed and merged, but having `fixed_length_size` "leak" through to affect all of the decoders is evidence that the Decoder API needs to be changed. I think that the Decoder constructors should accept `parquet::SchemaElement*` so that type-specific metadata can be accessed in the appropriate place and nowhere else.

wesm: In fact let me go ahead and do this in #22 right now

wesm: https://github.com/wesm/parquet-cpp/commit/766b6a52e6622be9dae0a8d93d237bda9cf064d6

majetideepak:
Sure, I will do it now.

wesm:
The struct packing? Or something else

wesm:
Oh I see, the feedback on #22. Thanks

majetideepak:
Yeah, review #22. I am still looking at the struct packing. I will get back to you once I go through it.

src/parquet/parquet.h
```diff
@@ -179,6 +190,8 @@ class ColumnReader {
   std::vector<uint8_t> values_buffer_;
   int num_decoded_values_;
   int buffered_values_offset_;
+  //support for fixed length byte arrays
+  int fixed_length_size;
```
wesm: You can grab this from the `parquet::SchemaElement` below and avoid data type-specific members

majetideepak:
@wesm I looked at #22 and it looks good to me.

src/parquet/parquet.h
majetideepak: Yes!

src/parquet/encodings/plain-encoding.h
majetideepak: The code from flatbuffers looks reasonable to me. 
However, I am inclined towards having the padding to allow aligned memory access. Can you explain where the padding is an issue in a template context ? If required, we can also specialize the template for "Int96" type at places where there is a problem.

wesm: I'll let @nongli weigh in; it might be worth doing some performance experiments. The main benefit in a template context is being able to allocate only 12 bytes per buffered value and do pointer arithmetic on `Int96*`. If you allocate `n * sizeof(Int96)` you can still do pointer arithmetic but you are wasting space. 

I've been hearing nowadays that on modern processors, the performance penalty for unaligned memory access is not what it used to be. 

Probably we're doing to want to add a defines that turns on/off struct packing and use `sizeof(Int96)` for memory allocation and assume that pointer arithmetic and dereferencing on works as expected.

majetideepak: Your statement on performance got me intrigued. After a bit of googling, people seem to have mixed reviews. Performance will depend upon other factors as well like caching and SSE/AVX. I will disable padding for Int96 to simplify and avoid any specialization of templates. We can look into this later when we deep dive into performance.

majetideepak:
Checked in macros to disable padding of Int96

src/parquet/parquet.h
```diff
@@ -40,6 +40,26 @@
 #include <thrift/protocol/TBinaryProtocol.h>
 #include <thrift/transport/TBufferTransports.h>
 
+//macros to disable padding. Required for Int96
+//These macros ensure portablity across compilers and platforms [https://github.com/google/flatbuffers/blob/master/include/flatbuffers/flatbuffers.h#L1355]
+#if defined(_MSC_VER)
```
wesm: When you rebase this, can you please put this in `parquet/util/compiler-util.h`?

src/parquet/parquet.h
```diff
@@ -40,6 +40,26 @@
 #include <thrift/protocol/TBinaryProtocol.h>
 #include <thrift/transport/TBufferTransports.h>
 
+//macros to disable padding. Required for Int96
```
wesm: We should add appropriate defines so that struct packing can be disabled by a cmake option. I'll create a JIRA for this

src/parquet/parquet.h
majetideepak: Will do.

src/parquet/parquet.h
majetideepak: I don't think we should give an option of disabling padding until we specialize the templates to handle padding as well.

wesm: Hence creating a separate JIRA about it :) We will need several unit tests before we go that route as well.

majetideepak: Sure! We should create a JIRA for specialization of templates for Int96 with padding as well.
The unit tests will capture both cases.

wesm: I think we should do that here and use `sizeof(Int96)` for memory allocations. As soon as #22 and/or #23 are merged these changes can be rebased on the templated column reader and decoders.

majetideepak: I already changed the code to sizeof(Int96) since padding is disabled in the current updated version. Once #22 and #23 are in, the same template code will work for Int96 as well since sizeof(Int96) will always be 12.
 If we plan to support the padding case as well, we have to use the pragmas you suggested to enable a specialization for Int96 where we explicitly set size to 12.

wesm:
@majetideepak whenever it seems like #23 is close to merge ready (today hopefully?) can you refactor these changes on top of the new reader / decoder template structure?

Let's defer on unit testing this (let's create a JIRA) until #26 is reviewed and merged, since it changes API for accessing values

majetideepak:
Yes, I can quickly refactor these changes. I will test the code on my side and wait until #26 is merged to use the updated API.

PR #26, opened 2016-01-27
PARQUET-435: Change column reader methods to be array-oriented rather than scalar

wesm:
Column scanning and record reconstruction is independent of the Parquet file format and depends, among other things, on the data structures where the reconstructed data will end up. This is a work-in progress, but the basic idea is:

- APIs for reading a batch of repetition `ReadRepetitionLevels` or definition levels `ReadDefinitionLevels` into a preallocated `int16_t*`
- APIs for reading arrays of decoded values into preallocated memory (`ReadValues`)

These methods are only able to read data within a particular data page. Once you exhaust the data available in the data page (`ReadValues` returns 0), you must call `ReadNewPage`, which returns `true` is there is more data available.

Separately, I added a simple `Scanner` class that emulates the scalar value iteration functionality that existed previously. I used this to reimplement the `DebugPrint` method in `parquet_scanner.cc`. This obviously only works currently for flat data. 

I would like to keep the `ColumnReader` low level and primitive, concerned only with providing access to the raw data in a Parquet file as fast as possible. We can devise separate algorithms for inferring nested record structure by examining the arrays of decoded values and repetition/definition levels. The major benefit of separating raw data access from structure inference is that this can be pipelined with threads: one thread decompresses and decodes values and levels, and another thread can turn batches into a nested record- or column-oriented structure.


src/parquet/column_scanner.h
```diff
@@ -0,0 +1,149 @@
...
+    }
+    *def_level = def_levels_[level_offset_++];
+    *rep_level = 1;
+    return true;
+  }
+
+ protected:
+  ColumnReader* reader_;
+
+  static constexpr size_t BATCHSIZE = 128;
```
asandryh: Batch size should be configurable.

src/parquet/column_scanner.h
```diff
@@ -0,0 +1,149 @@
...
+        return false;
+      }
+    }
+    *def_level = def_levels_[level_offset_++];
+    *rep_level = 1;
+    return true;
+  }
+
+ protected:
+  ColumnReader* reader_;
```
asandryh: Is it worth making reader_ a shared_ptr<ColumnReader>?

src/parquet/column_scanner.h
wesm: Absolutely. I'll do that in this patch

src/parquet/column_scanner.h
wesm: Possibly. 

Right now the `ColumnReader` object lifetime is tied to the `RowGroupReader`, and this `Scanner`'s lifetime is tied to ColumnReader's lifetime, so unclear that shared ownership of `ColumnReader` makes sense semantically. 

If you think that we should decouple `ColumnReader` object lifetime from `RowGroupReader` then we can do that.

asandryh: If I read the code correctly, ColumnReader::GetScanner() returns a shared_ptr<Scanner>. So, it is possible that after a reader is destroyed, the scanner is still alive. To be usable (and avoid crashes), it needs reader_ .

wesm: Makes sense, will do.

src/parquet/column_reader.cc
```diff
@@ -212,6 +121,9 @@ bool ColumnReader::ReadNewPage() {
       // Read a data page.
       num_buffered_values_ = current_page_header_.data_page_header.num_values;
```
majetideepak: I am not sure if you can still use num_buffered_values_ = current_page_header_.data_page_header.num_values; for the vectorized reader.
current_page_header_.data_page_header.num_values counts the NULLs as well, which are actually not present physically in the DATA_PAGE. Inside the decoder (plain-decoder), you loop on this number to populate the buffers. If you have nulls, you shouldnt be looping on this value.

majetideepak: The DataPageHeaderV2 counts the nulls separately. We could then use `num_buffered_values_ = num_values - num_nulls`

wesm: I missed that part of the spec. How annoying! 

Perhaps the API should be then:

`size_t ReadBatch(int batchsize, int16_t* def_levels, int16_t* rep_levels, T* values, size_t* values_read)`

Since you will have to inspect the definition levels (using the maximum definition level -- needs to be passed in) to determine how many encoded values are available. thoughts?

wesm:
@asandryh @majetideepak I'll rebase this patch first thing tomorrow and change the batch read API as described (since it's not possible — with DataPageV1 to know how many values to decode without first examining the repetition / definition levels). I'll address the other comments as well. Pls let me know any other feedback

src/parquet/column_reader.cc
majetideepak: Can you elaborate more on the decoder side? Are you planning to process the definition levels in the batch to find the number of non-nulls ?

wesm: Yes -- after decoding the definition levels you can count the number of materialized values and then request that many to be decoded in batch. 

For applications that are looking to save as many CPU cycles as possible we can figure out an API that avoids this; for now I'd like to get to a correct-and-tested state and then focus on optimizations.

wesm:
I'll work on getting this rebased and with the read API fixed by EOD

src/parquet/column_reader.cc
majetideepak: Can we not pass the maximum def/rep level as an argument in the API?
We do not need this for DataPageV2. Can we have a field inside the ColumnReader instead ?

wesm: Sure, that's fine. Let me clean up the patch with that in mind and I'll ping you to have a look

wesm:
@majetideepak @asandryh Rebased and responded to your comments. For the moment it's easiest to read the values and levels together (see `ReadBatch`) -- I suggest we pursue optimizations after we have more rigorous unit testing. I'm going to work on schemas tomorrow and plotting a course toward writing flat primitive data so we can verify roundtripping data to a file. 

We should start a side discussion about building up testing tooling, because continuing to do major development without thorough testing isn't likely to be a good use of anyone's time. I'm open to ideas, including scripting `parquet-mr` to generate sample files for testing.

asandryh:
I've created PARQUET-479 to work on the testing setup. Regression tests should be a part of each Travis CI build, too.

src/parquet/column/reader.cc
```diff
@@ -119,15 +118,25 @@ bool TypedColumnReader<TYPE>::ReadNewPage() {
...
             new RleDecoder(buffer, num_definition_bytes, 1));
         buffer += num_definition_bytes;
         uncompressed_len -= sizeof(uint32_t);
         uncompressed_len -= num_definition_bytes;
+      } else {
+        // REQUIRED field
+        max_definition_level_ = 0;
       }
 
       // TODO: repetition levels
```
majetideepak: I am going to implement this part via PARQUET-169 in the next couple of hours.

src/parquet/column/reader.h
```diff
@@ -152,6 +184,32 @@ class TypedColumnReader : public ColumnReader {
...
+template <int TYPE>
+inline size_t TypedColumnReader<TYPE>::ReadValues(size_t batch_size, T* out) {
+  size_t num_decoded = current_decoder_->Decode(out, batch_size);
+  num_decoded_values_ += num_decoded;
+  return num_decoded;
+}
+
+template <int TYPE>
+inline size_t TypedColumnReader<TYPE>::ReadBatch(int batch_size, int16_t* def_levels,
+    int16_t* rep_levels, T* values, size_t* values_read) {
```
majetideepak: `batch_size = min(batch_size, num_buffered_values_)` should be here ?

src/parquet/column/reader.h
```diff
@@ -152,6 +184,32 @@ class TypedColumnReader : public ColumnReader {
...
+  size_t num_levels = ReadDefinitionLevels(batch_size, def_levels);
+
```
majetideepak: add `size_t num_rep_levels = ReadRepetitionLevels(batch_size, rep_levels);` here. We should add a check to make sure number of rep/dep levels match.

src/parquet/column/reader.h
```diff
@@ -152,6 +184,32 @@ class TypedColumnReader : public ColumnReader {
...
+  size_t values_to_read = 0;
+  for (size_t i = 0; i < num_levels; ++i) {
+    if (def_levels[i] == max_definition_level_) {
+      ++values_to_read;
```
majetideepak: TODO later: Will be efficient if we fuse this loop inside DecodeMany.

src/parquet/column/scanner.h
```diff
@@ -0,0 +1,208 @@
...
+
+#include "parquet/column/reader.h"
+#include "parquet/thrift/parquet_types.h"
+
+namespace parquet_cpp {
+
+class ColumnReader;
+
+template <int TYPE>
+class TypedColumnReader;
```
asandryh: Nitpicking: You don't need lines 31-34.

src/parquet/column/scanner.h
```diff
@@ -0,0 +1,208 @@
...
+
+#define DEFAULT_SCANNER_BATCH_SIZE 128
+
+class Scanner {
+ public:
+  explicit Scanner(std::shared_ptr<ColumnReader> reader) :
+      reader_(reader),
+      batch_size_(DEFAULT_SCANNER_BATCH_SIZE),
```
asandryh: Instead of pre-defining the default batch size, it is better to make it a ctor argument with a default value 128.

src/parquet/column/reader.h
wesm: Thanks -- I did some digging to clarify that `num_buffered_values_` is equal to the number of encoded levels (when the data is optional and/or repeated) and added a comment to that effect.

src/parquet/column/reader.h
wesm: done

src/parquet/column/reader.h
wesm: added a TODO comment

src/parquet/column/scanner.h
wesm: fixed

src/parquet/column/scanner.h
wesm: done

wesm:
thanks for the comments! I just rebased on `upstream/master` (since I moved `column_*` to `parquet/column/`).

src/parquet/reader.h
```diff
@@ -81,7 +81,7 @@ class RowGroupReader {
 
   // Construct a ColumnReader for the indicated row group-relative column. The
   // returned object is owned by the RowGroupReader
```
nongli: this comment can't be right anymore

src/parquet/column/scanner.h
```diff
@@ -0,0 +1,206 @@
...
+#include <ostream>
+#include <string>
+#include <vector>
+
+#include "parquet/column/reader.h"
+#include "parquet/thrift/parquet_types.h"
+
+namespace parquet_cpp {
+
+#define DEFAULT_SCANNER_BATCH_SIZE 128
```
nongli: Let's use static const member vars instead of #defines

majetideepak:
Where does `ReadNewPage()` happen in the batch reader ? Is the downstream application expected to call ReadNewPage() explicitly ?
Another TODO is to read multiple data pages until `batch_size < num_buffered_values_`;

majetideepak:
Sorry! You specified this in the main comments.

wesm:
@majetideepak oops, thanks for pointing that out. I will make sure that ReadBatch automatically called ReadNewPage. I'm also making `ReadNewPage` private

wesm:
OK, I think this should be merge-ready unless there are more comments

majetideepak:
+1 LGTM

src/parquet/column/reader.cc
```diff
@@ -165,6 +173,39 @@ bool TypedColumnReader<TYPE>::ReadNewPage() {
...
 
+// ----------------------------------------------------------------------
+// Batch read APIs
+
+static size_t DecodeMany(RleDecoder* decoder, int16_t* levels, size_t batch_size) {
+  size_t num_decoded = 0;
+
+  // TODO(wesm): Push this decoding down into RleDecoder itself
+  for (size_t i = 0; i < batch_size; ++i) {
+    if (!decoder->Get(levels + i)) {
```
nongli: When does this return false? (We shouldn't fix this in this patch)

This is only for EOS errors right? Should we make the contract so the caller of the low level APIs should know and not over run it? (The lower levels can throw if the data is bad, but this is not the normal control flow path).

src/parquet/column/reader.h
```diff
@@ -68,13 +69,14 @@ class ColumnReader {
...
-  virtual bool ReadNewPage() = 0;
-
   // Returns true if there are still values in this column.
   bool HasNext() {
-    if (num_buffered_values_ == 0) {
-      ReadNewPage();
-      if (num_buffered_values_ == 0) return false;
+    // Either there is no data page available yet, or the data page has been
+    // exhausted
+    if (!num_buffered_values_ || num_decoded_values_ == num_buffered_values_) {
```
nongli: I prefer consistently using num_buffered_values_ == 0 over !num_buffered_values_

src/parquet/column/reader.h
```diff
@@ -103,17 +118,29 @@ class ColumnReader {
 
   parquet::PageHeader current_page_header_;
 
-  // Not set if field is required.
+  // Not set if field is required (flat schemas), or if the whole schema tree
```
nongli: This is not set if the entire schema is required? The flat or not shouldn't matter right?

src/parquet/column/scanner.h
```diff
@@ -0,0 +1,206 @@
...
+      values_buffered_(0) {
+    // TODO: don't allocate for required fields
+    def_levels_.resize(batch_size_);
+    rep_levels_.resize(batch_size_);
+  }
+
+  static std::shared_ptr<Scanner> Make(std::shared_ptr<ColumnReader> col_reader,
+      size_t batch_size = DEFAULT_SCANNER_BATCH_SIZE);
+
+  virtual void PrintNext(std::ostream& out, int width) = 0;
```
nongli: classes with virtual functions need a virtual dtor. Our compile flags should warn on this.

src/parquet/column/scanner.h
```diff
@@ -0,0 +1,206 @@
...
+    return value_offset_ < values_buffered_ || reader_->HasNext();
+  }
+
+  size_t batch_size() const { return batch_size_;}
+
+  void SetBatchSize(size_t batch_size) {
+    batch_size_ = batch_size;
+  }
+
+ protected:
```
nongli: Do all these need to be protected? I'd like to expose as little as possible to the subclass.

src/parquet/column/reader.cc
wesm: Right, only for EOS errors. I'm planning to write some tests that will explore these issues in some detail so we can define precisely what will / will not raise an exception.

src/parquet/column/reader.h
wesm: True. I'll update the doc

nongli:
LGTM

src/parquet/column/reader.h
wesm: noted, fixing now

src/parquet/column/scanner.h
wesm: Hmm. No compiler warning but adding a virtual dtor anyway

src/parquet/column/scanner.h
wesm: No, I'm making `reader_` private

wesm:
@nongli we are good to go. Finally added a pre-commit hook so I don't get bitten with a cpplint issue again

wesm:
Thank you!

PR #27, opened 2016-01-28
PARQUET-428: Support INT96 and FIXED_LEN_BYTE_ARRAY types

majetideepak:
This PR adds support for INT96 and FIXED_LEN_BYTE_ARRAY types.
It modifies the examples and DebugPrint to handle these types.

### DIFF
--- a/src/parquet/encodings/plain-encoding.h
+++ b/src/parquet/encodings/plain-encoding.h
@@ -74,6 +74,22 @@ inline int PlainDecoder<parquet::Type::BYTE_ARRAY>::Decode(ByteArray* buffer,
   return max_values;
 }
 
+// Template specialization for FIXED_LEN_BYTE_ARRAY
+template <>
+inline int PlainDecoder<parquet::Type::FIXED_LEN_BYTE_ARRAY>::Decode(FixedLenByteArray* buffer,
+    int max_values) {
+  max_values = std::min(max_values, num_values_);
+  int len = schema_->type_length;
+  for (int i = 0; i < max_values; ++i) {
+    if (len_ < len) ParquetException::EofException();
+    buffer[i].ptr = data_;
+    data_ += len;
+    len_ -= len;
+  }
+  num_values_ -= max_values;
+  return max_values;
+}
+
 template <>
 class PlainDecoder<parquet::Type::BOOLEAN> : public Decoder<parquet::Type::BOOLEAN> {
  public:
--- a/src/parquet/reader.cc
+++ b/src/parquet/reader.cc
@@ -217,6 +217,9 @@ static string parquet_type_to_string(Type::type t) {
     case Type::INT64:
       return "INT64";
       break;
+    case Type::INT96:
+      return "INT96";
+      break;
     case Type::FLOAT:
       return "FLOAT";
       break;
@@ -225,9 +228,6 @@ static string parquet_type_to_string(Type::type t) {
       break;
     case Type::BYTE_ARRAY:
       return "BYTE_ARRAY";
-      break;
-    case Type::INT96:
-      return "INT96";
       break;
     case Type::FIXED_LEN_BYTE_ARRAY:
       return "FIXED_LEN_BYTE_ARRAY";
@@ -239,7 +239,7 @@ static string parquet_type_to_string(Type::type t) {
 }
 
 // the fixed initial size is just for an example
-#define COL_WIDTH "17"
+#define COL_WIDTH "20"
 
 void ParquetFileReader::DebugPrint(std::ostream& stream, bool print_values) {
   if (!parsed_metadata_) {
@@ -251,10 +251,6 @@ void ParquetFileReader::DebugPrint(std::ostream& stream, bool print_values) {
   for (int c = 1; c < metadata_.schema.size(); ++c) {
     stream << "Column " << c-1 << ": " << metadata_.schema[c].name << " ("
            << parquet_type_to_string(metadata_.schema[c].type);
-    if (metadata_.schema[c].type == Type::INT96 ||
-        metadata_.schema[c].type == Type::FIXED_LEN_BYTE_ARRAY) {
-      stream << " - not supported";
-    }
     stream << ")\n";
   }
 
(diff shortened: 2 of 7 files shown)


majetideepak:
The first check passed on Travis CI https://travis-ci.org/apache/parquet-cpp/builds/105457807

src/parquet/encodings/plain-encoding.h
```diff
@@ -71,6 +71,22 @@ inline int PlainDecoder<parquet::Type::BYTE_ARRAY>::Decode(ByteArray* buffer,
   return max_values;
 }
 
+// Template specialization for FIXED_LEN_BYTE_ARRAY
+template <>
+inline int PlainDecoder<parquet::Type::FIXED_LEN_BYTE_ARRAY>::Decode(FixedLenByteArray* buffer,
```
wesm: How about dictionary encoding (see the `BYTE_ARRAY` specialization there)? It's hard to verify that it works without tests, though, we should try to address that soon.

src/parquet/reader.cc
```diff
@@ -217,6 +217,9 @@ static string parquet_type_to_string(Type::type t) {
     case Type::INT64:
       return "INT64";
       break;
+    case Type::INT96:
+      return "INT96";
```
wesm: Aside: we should move these names to `static constexpr const char*` members of `type_traits`

src/parquet/types.h
```diff
@@ -32,11 +33,38 @@ struct ByteArray {
   const uint8_t* ptr;
 };
 
+struct FixedLenByteArray {
+  const uint8_t* ptr;
+};
+
+MANUALLY_ALIGNED_STRUCT(1) Int96{
```
wesm: Space after `Int96`

src/parquet/types.h
```diff
@@ -32,11 +33,38 @@ struct ByteArray {
...
+  uint32_t value[3];
+};
+STRUCT_END(Int96,12);
```
wesm: space after comma

src/parquet/types.h
```diff
@@ -32,11 +33,38 @@ struct ByteArray {
...
 
 static inline std::string ByteArrayToString(const ByteArray& a) {
   return std::string(reinterpret_cast<const char*>(a.ptr), a.len);
 }
 
+static inline std::string Int96ToString(const Int96& a) {
+  std::string result = "";
+  for (int i=0; i<3; i++) {
```
wesm: whitespace around `=` and `<`

src/parquet/types.h
```diff
@@ -32,11 +33,38 @@ struct ByteArray {
...
+    result.append(std::to_string(a.value[i]));
+    result.append(" ");
+  }
+  return result;
+}
+
+static inline std::string FixedLenByteArrayToString(const FixedLenByteArray& a, int len) {
+  const uint8_t *bytes = reinterpret_cast<const uint8_t*>(a.ptr);
+  std::string result = "";
+  for (int i=0; i<len; i++) {
```
wesm: whitespace

src/parquet/types.h
```diff
@@ -32,11 +33,38 @@ struct ByteArray {
...
+  uint32_t value[3];
+};
+STRUCT_END(Int96,12);
 
 static inline std::string ByteArrayToString(const ByteArray& a) {
   return std::string(reinterpret_cast<const char*>(a.ptr), a.len);
 }
 
+static inline std::string Int96ToString(const Int96& a) {
+  std::string result = "";
```
wesm: `std::stringstream`?

src/parquet/types.h
```diff
@@ -32,11 +33,38 @@ struct ByteArray {
...
+  for (int i=0; i<3; i++) {
+    result.append(std::to_string(a.value[i]));
+    result.append(" ");
+  }
+  return result;
+}
+
+static inline std::string FixedLenByteArrayToString(const FixedLenByteArray& a, int len) {
+  const uint8_t *bytes = reinterpret_cast<const uint8_t*>(a.ptr);
+  std::string result = "";
```
wesm: Same question re: `std::stringstream`

src/parquet/types.h
majetideepak: strangely `make lint` did not catch these. Will  Fix.

src/parquet/reader.cc
```diff
@@ -373,7 +375,17 @@ void ParquetFileReader::DebugPrint(std::ostream& stream, bool print_values) {
...
+              FixedLenByteArray val = reinterpret_cast<FixedLenByteArrayReader*>(
+                  readers[c])->NextValue(&def_level[c], &rep_level[c]);
+              if (def_level[c] >= rep_level[c]) {
+                string result = FixedLenByteArrayToString(val, metadata_.schema[c+1].type_length);
+                snprintf(buffer, bufsize, "%-" COL_WIDTH"s", result.c_str());
+                stream << buffer;
+              }
+              break;
+            }
+           default:
```
wesm: This is a bummer since I removed this whole loop / switch statement in #26, but I can take care of refactoring / rebase

src/parquet/encodings/plain-encoding.h
majetideepak: I will add that now. Actually its rare for these types to be dictionary encoded.
I tested this on my side and it seems to work okay. We should definitely add tests after PARQUET-435

wesm:
Can you create `src/parquet/types-test.cc` and add a unit test asserting `ASSERT_EQ(12, sizeof(Int96))`? We can add more unit tests relating to types as we go.

majetideepak:
The STRUCT_END(Int96, 12) includes a static assert. I checked that this works locally.
Do you still want to add a test ?

wesm:
Ah good point, thanks. I think it's fine to omit the test, then. We should eventually move `parquet_type_to_string` to `types.h` and add tests that verify the type names (and other properties and utilities we add).

julienledem:
Is this one good to go?
@wesm @majetideepak

majetideepak:
Rebased and good to go from my side.

wesm:
lgtm +1. @asandryh will need to fix up minor conflicts #29

wesm:
You're in the wrong place -- this is the original repo for the Parquet C++ project

PR #29, opened 2016-01-28
PARQUET-472: Changed the ownership of InputStream in ColumnReader.

asandryh:


src/parquet/column_reader.cc
```diff
@@ -33,16 +33,13 @@ using parquet::FieldRepetitionType;
 using parquet::PageType;
 using parquet::Type;
 
-
-ColumnReader::~ColumnReader() {
-  delete stream_;
-}
+ColumnReader::~ColumnReader() {}
```
nongli: Can we delete this entirely?

asandryh: Yes. Missed it.

src/parquet/column_reader.cc
```diff
@@ -167,8 +164,10 @@ bool TypedColumnReader<TYPE>::ReadNewPage() {
   return true;
 }
 
-std::shared_ptr<ColumnReader> ColumnReader::Make(const parquet::ColumnMetaData* metadata,
-    const parquet::SchemaElement* element, InputStream* stream) {
+std::shared_ptr<ColumnReader> ColumnReader::Make(
+                                const parquet::ColumnMetaData* metadata,
+                                const parquet::SchemaElement* element,
+                                std::unique_ptr<InputStream>& stream) {
```
wesm: This indentation (similarly below) is inconsistent with the rest of the codebase. Since we are sharing code with Impala (and will continue to) and Kudu, we should indent subsequent lines of function signatures 4 spaces.

src/parquet/reader.cc
```diff
@@ -106,9 +106,9 @@ ColumnReader* RowGroupReader::Column(size_t i) {
   source->Seek(col_start);
 
   // TODO(wesm): Law of demeter violation
-  size_t bytes_read = source->Read(input->size(), input->data());
-
-  if (bytes_read != input->size()) {
+  ScopedInMemoryInputStream* pInput = static_cast<ScopedInMemoryInputStream*>(input.get());
+  size_t bytes_read = source->Read(pInput->size(), pInput->data());
```
wesm: Google C++ forbids camel case and Hungarian notation

- https://google.github.io/styleguide/cppguide.html#Variable_Names
- https://google.github.io/styleguide/cppguide.html#Windows_Code

src/parquet/column_reader.cc
asandryh: Sure, will fix it. Wasn't aware of this guideline.

src/parquet/column_reader.h
```diff
@@ -123,7 +124,7 @@ class TypedColumnReader : public ColumnReader {
   typedef typename type_traits<TYPE>::value_type T;
 
   TypedColumnReader(const parquet::ColumnMetaData* metadata,
-      const parquet::SchemaElement* schema, InputStream* stream) :
+      const parquet::SchemaElement* schema, std::unique_ptr<InputStream>& stream) :
```
wesm: I think it would be better to pass by value here, otherwise you can only pass an l-value as the `stream` argument:

http://stackoverflow.com/questions/8114276/how-do-i-pass-a-unique-ptr-argument-to-a-constructor-or-a-function

src/parquet/reader.cc
asandryh: @wesm I'll fix the variable name.
@nongli Using clang-format is helpful. Is it possible to check during a Travis build whether code confirm to style?

nongli: @asandryh I've never set it up but maybe this is what we need?
http://clang.llvm.org/extra/clang-tidy/

src/parquet/column_reader.h
asandryh: You're right. I wanted to avoid extra std::move() calls. This works now, but in the future we may want to pass a temporary to the ctor.

src/parquet/reader.cc
asandryh: A quick online scan suggests it is possible to integrate with Travis CI. I've created PARQUET-477 to track this.

julienledem:
@wesm good to go?

wesm:
Needs to be rebased

asandryh:
The PR is rebased on the latest master. @nongli or @julienledem Please take a look and commit, if there are no other comments.

wesm:
lgtm +1

PR #30, opened 2016-01-29
PARQUET-169: Implement support for bulk reading and writing rep/def levels

majetideepak:
Added a LevelDecoder and LevelEncoder class to read and write batches of def/rep levels.
Added tests to verify the functionality.

### DIFF
--- a/src/parquet/column/column-reader-test.cc
+++ b/src/parquet/column/column-reader-test.cc
@@ -124,9 +124,6 @@ TEST_F(TestPrimitiveReader, TestInt32FlatOptional) {
 
   Int32Reader* reader = static_cast<Int32Reader*>(reader_.get());
 
-  std::vector<int32_t> vexpected;
-  std::vector<int16_t> dexpected;
-
   size_t values_read = 0;
   size_t batch_actual = 0;
 
@@ -157,6 +154,62 @@ TEST_F(TestPrimitiveReader, TestInt32FlatOptional) {
   ASSERT_EQ(0, values_read);
 }
 
+TEST_F(TestPrimitiveReader, TestInt32FlatRepeated) {
+  vector<int32_t> values = {1, 2, 3, 4, 5};
+  vector<int16_t> def_levels = {2, 1, 1, 2, 2, 1, 1, 2, 2, 1};
+  vector<int16_t> rep_levels = {0, 1, 1, 0, 0, 1, 1, 0, 0, 1};
+
+  size_t num_values = values.size();
+  parquet::Encoding::type value_encoding = parquet::Encoding::PLAIN;
+
+  vector<uint8_t> page1;
+  test::DataPageBuilder<Type::INT32> page_builder(&page1);
+
+  // Definition levels precede the values
+  page_builder.AppendRepLevels(rep_levels, 1, parquet::Encoding::RLE);
+  page_builder.AppendDefLevels(def_levels, 2, parquet::Encoding::RLE);
+  page_builder.AppendValues(values, parquet::Encoding::PLAIN);
+
+  pages_.push_back(page_builder.Finish());
+
+  NodePtr type = schema::Int32("a", Repetition::REPEATED);
+  ColumnDescriptor descr(type, 2, 1);
+  InitReader(&descr);
+
+  Int32Reader* reader = static_cast<Int32Reader*>(reader_.get());
+
+  size_t values_read = 0;
+  size_t batch_actual = 0;
+
+  vector<int32_t> vresult(3, -1);
+  vector<int16_t> dresult(5, -1);
+  vector<int16_t> rresult(5, -1);
+
+  batch_actual = reader->ReadBatch(5, &dresult[0], &rresult[0],
+      &vresult[0], &values_read);
+  ASSERT_EQ(5, batch_actual);
+  ASSERT_EQ(3, values_read);
+
+  ASSERT_TRUE(vector_equal(vresult, slice(values, 0, 3)));
+  ASSERT_TRUE(vector_equal(dresult, slice(def_levels, 0, 5)));
+  ASSERT_TRUE(vector_equal(rresult, slice(rep_levels, 0, 5)));
+
+  batch_actual = reader->ReadBatch(5, &dresult[0], &rresult[0],
+      &vresult[0], &values_read);
+  ASSERT_EQ(5, batch_actual);
+  ASSERT_EQ(2, values_read);
+
+  ASSERT_TRUE(vector_equal(slice(vresult, 0, 2), slice(values, 3, 5)));
+  ASSERT_TRUE(vector_equal(dresult, slice(def_levels, 5, 10)));
+  ASSERT_TRUE(vector_equal(rresult, slice(rep_levels, 5, 10)));
+
+  // EOS, pass all nullptrs to check for improper writes. Do not segfault /
+  // core dump
+  batch_actual = reader->ReadBatch(5, nullptr, nullptr,
+      nullptr, &values_read);
+  ASSERT_EQ(0, batch_actual);
+  ASSERT_EQ(0, values_read);
+}
 } // namespace test
 
 } // namespace parquet_cpp
--- a/src/parquet/column/levels-test.cc
+++ b/src/parquet/column/levels-test.cc
@@ -1,0 +1,129 @@
+// Licensed to the Apache Software Foundation (ASF) under one
+// or more contributor license agreements.  See the NOTICE file
+// distributed with this work for additional information
+// regarding copyright ownership.  The ASF licenses this file
+// to you under the Apache License, Version 2.0 (the
+// "License"); you may not use this file except in compliance
+// with the License.  You may obtain a copy of the License at
+//
+//   http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing,
+// software distributed under the License is distributed on an
+// "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+// KIND, either express or implied.  See the License for the
+// specific language governing permissions and limitations
+// under the License.
+
+#include <cstdlib>
+#include <iostream>
+#include <sstream>
+#include <string>
+
+#include <gtest/gtest.h>
+
+#include "parquet/thrift/parquet_types.h"
+#include "parquet/column/levels.h"
+
+using std::string;
+
+namespace parquet_cpp {
+
+class TestLevels : public ::testing::Test {
+ public:
+  int GenerateLevels(int min_repeat_factor, int max_repeat_factor,
+      int max_level, std::vector<int16_t>& input_levels) {
+    int total_count = 0;
+    // for each repetition count upto max_repeat_factor
+    for (int repeat = min_repeat_factor; repeat <= max_repeat_factor; repeat++) {
+      // repeat count increase by a factor of 2 for every iteration
+      int repeat_count = (1 << repeat);
+      // generate levels for repetition count upto the maximum level
+      int value = 0;
+      int bwidth = 0;
+      while (value <= max_level) {
+        for (int i = 0; i < repeat_count; i++) {
+          input_levels[total_count++] = value;
+        }
+        value = (2 << bwidth) - 1;
+        bwidth++;
+      }
+    }
+    return total_count;
+  }
+
+  void VerifyLevelsEncoding(parquet::Encoding::type encoding, int max_level,
+      std::vector<int16_t>& input_levels) {
+    LevelEncoder encoder;
+    LevelDecoder decoder;
+    int levels_count = 0;
+    std::vector<int16_t> output_levels;
+    std::vector<uint8_t> bytes;
+    int num_levels = input_levels.size();
+    output_levels.resize(num_levels);
+    bytes.resize(2 * num_levels);
+    ASSERT_EQ(num_levels, output_levels.size());
+    ASSERT_EQ(2 * num_levels, bytes.size());
+    // start encoding and decoding
+    if (encoding == parquet::Encoding::RLE) {
+      // leave space to write the rle length value
+      encoder.Init(encoding, max_level, num_levels,
+          bytes.data() + sizeof(uint32_t), bytes.size());
+
+      levels_count = encoder.Encode(num_levels, input_levels.data());
+      (reinterpret_cast<uint32_t*>(bytes.data()))[0] = encoder.len();
+
+    } else {
+      encoder.Init(encoding, max_level, num_levels,
+          bytes.data(), bytes.size());
+      levels_count = encoder.Encode(num_levels, input_levels.data());
+    }
+
+    ASSERT_EQ(levels_count, num_levels);
+
+    decoder.Init(encoding, max_level, num_levels, bytes.data());
+    levels_count = decoder.Decode(num_levels, output_levels.data());
+
+    ASSERT_EQ(levels_count, num_levels);
+
+    for (int i = 0; i < num_levels; i++) {
+      EXPECT_EQ(input_levels[i], output_levels[i]);
+    }
+  }
+};
+
+// test levels with maximum bit-width from 1 to 8
+// increase the repetition count for each iteration by a factor of 2
+TEST_F(TestLevels, TestEncodeDecodeLevels) {
+  int min_repeat_factor = 0;
+  int max_repeat_factor = 7; // 128
+  int max_bit_width = 8;
+  std::vector<int16_t> input_levels;
+  parquet::Encoding::type encodings[2] = {parquet::Encoding::RLE,
+      parquet::Encoding::BIT_PACKED};
+
+  // for each encoding
+  for (int encode = 0; encode < 2; encode++) {
+    parquet::Encoding::type encoding = encodings[encode];
+    // BIT_PACKED requires a sequence of atleast 8
+    if (encoding == parquet::Encoding::BIT_PACKED) min_repeat_factor = 3;
+
+    // for each maximum bit-width
+    for (int bit_width = 1; bit_width <= max_bit_width; bit_width++) {
+      int num_levels_per_width = ((2 << max_repeat_factor) - (1 << min_repeat_factor));
+      int num_levels = (bit_width + 1) * num_levels_per_width;
+      input_levels.resize(num_levels);
+      ASSERT_EQ(num_levels, input_levels.size());
+
+      // find the maximum level for the current bit_width
+      int max_level = (1 << bit_width) - 1;
+      // Generate levels
+      int total_count = GenerateLevels(min_repeat_factor, max_repeat_factor,
+          max_level, input_levels);
+      ASSERT_EQ(num_levels, total_count);
+      VerifyLevelsEncoding(encoding, max_level, input_levels);
+    }
+  }
+}
+
+} // namespace parquet_cpp
--- a/src/parquet/column/levels-test.cc
+++ b/src/parquet/column/levels-test.cc
@@ -79,12 +79,12 @@ class TestLevels : public ::testing::Test {
       levels_count = encoder.Encode(num_levels, input_levels.data());
     }
 
-    ASSERT_EQ(levels_count, num_levels);
+    ASSERT_EQ(num_levels, levels_count);
 
     decoder.Init(encoding, max_level, num_levels, bytes.data());
     levels_count = decoder.Decode(num_levels, output_levels.data());
 
-    ASSERT_EQ(levels_count, num_levels);
+    ASSERT_EQ(num_levels, levels_count);
 
     for (int i = 0; i < num_levels; i++) {
       EXPECT_EQ(input_levels[i], output_levels[i]);
--- a/src/parquet/column/levels.h
+++ b/src/parquet/column/levels.h
@@ -1,0 +1,156 @@
+// Licensed to the Apache Software Foundation (ASF) under one
+// or more contributor license agreements.  See the NOTICE file
+// distributed with this work for additional information
+// regarding copyright ownership.  The ASF licenses this file
+// to you under the Apache License, Version 2.0 (the
+// "License"); you may not use this file except in compliance
+// with the License.  You may obtain a copy of the License at
+//
+//   http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing,
+// software distributed under the License is distributed on an
+// "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+// KIND, either express or implied.  See the License for the
+// specific language governing permissions and limitations
+// under the License.
+
+#ifndef PARQUET_COLUMN_LEVELS_H
+#define PARQUET_COLUMN_LEVELS_H
+
+#include "parquet/exception.h"
+#include "parquet/thrift/parquet_types.h"
+#include "parquet/encodings/encodings.h"
+#include "parquet/util/rle-encoding.h"
+
+namespace parquet_cpp {
+
+class LevelEncoder {
+ public:
+  LevelEncoder() {}
+
+  // Initialize the LevelEncoder.
+  void Init(parquet::Encoding::type encoding, int16_t max_level,
+      int num_buffered_values, uint8_t* data, int data_size) {
+    bit_width_ = BitUtil::Log2(max_level + 1);
+    encoding_ = encoding;
+    switch (encoding) {
+      case parquet::Encoding::RLE: {
+        rle_encoder_.reset(new RleEncoder(data, data_size, bit_width_));
+        break;
+      }
+      case parquet::Encoding::BIT_PACKED: {
+        int num_bytes = BitUtil::Ceil(num_buffered_values * bit_width_, 8);
+        bit_packed_encoder_.reset(new BitWriter(data, num_bytes));
+        break;
+      }
+      default:
+        throw ParquetException("Unknown encoding type for levels.");
+    }
+  }
+
+  // Encodes a batch of levels from an array and returns the number of levels encoded
+  size_t Encode(size_t batch_size, const int16_t* levels) {
+    size_t num_encoded = 0;
+    if (!rle_encoder_ && !bit_packed_encoder_) {
+      throw ParquetException("Level encoders are not initialized.");
+    }
+
+    if (encoding_ == parquet::Encoding::RLE) {
+      for (size_t i = 0; i < batch_size; ++i) {
+        if (!rle_encoder_->Put(*(levels + i))) {
+          break;
+        }
+        ++num_encoded;
+      }
+      rle_encoder_->Flush();
+      rle_length_ = rle_encoder_->len();
+    } else {
+      for (size_t i = 0; i < batch_size; ++i) {
+        if (!bit_packed_encoder_->PutValue(*(levels + i), bit_width_)) {
+          break;
+        }
+        ++num_encoded;
+      }
+      bit_packed_encoder_->Flush();
+    }
+    return num_encoded;
+  }
+
+  int32_t len() {
+    assert(encoding_ == parquet::Encoding::RLE);
+    return rle_length_;
+  }
+
+ private:
+  int bit_width_;
+  int rle_length_;
+  parquet::Encoding::type encoding_;
+  std::unique_ptr<RleEncoder> rle_encoder_;
+  std::unique_ptr<BitWriter> bit_packed_encoder_;
+};
+
+
+class LevelDecoder {
+ public:
+  LevelDecoder() {}
+
+  // Initialize the LevelDecoder and return the number of bytes consumed
+  size_t Init(parquet::Encoding::type encoding, int16_t max_level,
+      int num_buffered_values, const uint8_t* data) {
+    uint32_t num_bytes = 0;
+    uint32_t total_bytes = 0;
+    bit_width_ = BitUtil::Log2(max_level + 1);
+    encoding_ = encoding;
+    switch (encoding) {
+      case parquet::Encoding::RLE: {
+        num_bytes = *reinterpret_cast<const uint32_t*>(data);
+        const uint8_t* decoder_data = data + sizeof(uint32_t);
+        rle_decoder_.reset(new RleDecoder(decoder_data, num_bytes, bit_width_));
+        return sizeof(uint32_t) + num_bytes;
+      }
+      case parquet::Encoding::BIT_PACKED: {
+        num_bytes = BitUtil::Ceil(num_buffered_values * bit_width_, 8);
+        bit_packed_decoder_.reset(new BitReader(data, num_bytes));
+        return num_bytes;
+      }
+      default:
+        throw ParquetException("Unknown encoding type for levels.");
+    }
+    return -1;
+  }
+
+  // Decodes a batch of levels into an array and returns the number of levels decoded
+  size_t Decode(size_t batch_size, int16_t* levels) {
+    size_t num_decoded = 0;
+    if (!rle_decoder_ && !bit_packed_decoder_) {
+      throw ParquetException("Level decoders are not initialized.");
+    }
+
+    if (encoding_ == parquet::Encoding::RLE) {
+      for (size_t i = 0; i < batch_size; ++i) {
+        if (!rle_decoder_->Get(levels + i)) {
+          break;
+        }
+        ++num_decoded;
+      }
+    } else {
+      for (size_t i = 0; i < batch_size; ++i) {
+        if (!bit_packed_decoder_->GetValue(bit_width_, levels + i)) {
+          break;
+        }
+        ++num_decoded;
+      }
+    }
+    return num_decoded;
+  }
+
+ private:
+  int bit_width_;
+  parquet::Encoding::type encoding_;
+  std::unique_ptr<RleDecoder> rle_decoder_;
+  std::unique_ptr<BitReader> bit_packed_decoder_;
+};
+
+} // namespace parquet_cpp
+#endif // PARQUET_COLUMN_LEVELS_H
--- a/src/parquet/column/reader.cc
+++ b/src/parquet/column/reader.cc
@@ -59,18 +59,6 @@ void TypedColumnReader<TYPE>::ConfigureDictionary(const DictionaryPage* page) {
   current_decoder_ = decoders_[encoding].get();
 }
 
-
-static size_t InitializeLevelDecoder(const uint8_t* buffer,
-    int16_t max_level, std::unique_ptr<RleDecoder>& decoder) {
-  int num_definition_bytes = *reinterpret_cast<const uint32_t*>(buffer);
-
-  decoder.reset(new RleDecoder(buffer + sizeof(uint32_t),
-          num_definition_bytes,
-          BitUtil::NumRequiredBits(max_level)));
-
-  return sizeof(uint32_t) + num_definition_bytes;
-}
-
 // PLAIN_DICTIONARY is deprecated but used to be used as a dictionary index
 // encoding.
 static bool IsDictionaryIndexEncoding(const parquet::Encoding::type& e) {
@@ -109,23 +97,29 @@ bool TypedColumnReader<TYPE>::ReadNewPage() {
       // the page size to determine the number of bytes in the encoded data.
       size_t data_size = page->size();
 
-      max_definition_level_ = descr_->max_definition_level();
+      int16_t max_definition_level = descr_->max_definition_level();
+      int16_t max_repetition_level = descr_->max_repetition_level();
+      //Data page Layout: Repetition Levels - Definition Levels - encoded values.
+      //Levels are encoded as rle or bit-packed.
+      //Init repetition levels
+      if (max_repetition_level > 0) {
+        size_t rep_levels_bytes = repetition_level_decoder_.Init(
+            page->repetition_level_encoding(),
+            max_repetition_level, num_buffered_values_, buffer);
+        buffer += rep_levels_bytes;
+        data_size -= rep_levels_bytes;
+      }
+      //TODO figure a way to set max_definition_level_ to 0
+      //if the initial value is invalid
 
-      // Read definition levels.
-      if (max_definition_level_ > 0) {
-        // Temporary hack until schema resolution implemented
-
-        size_t def_levels_bytes = InitializeLevelDecoder(buffer,
-            max_definition_level_, definition_level_decoder_);
-
+      //Init definition levels
+      if (max_definition_level > 0) {
+        size_t def_levels_bytes = definition_level_decoder_.Init(
+            page->definition_level_encoding(),
+            max_definition_level, num_buffered_values_, buffer);
         buffer += def_levels_bytes;
         data_size -= def_levels_bytes;
-      } else {
-        // REQUIRED field
-        max_definition_level_ = 0;
       }
-
-      // TODO: repetition levels
 
       // Get a decoder object for this page or create a new decoder if this is the
       // first page with this encoding.
@@ -172,31 +166,18 @@ bool TypedColumnReader<TYPE>::ReadNewPage() {
 // ----------------------------------------------------------------------
 // Batch read APIs
 
-static size_t DecodeMany(RleDecoder* decoder, int16_t* levels, size_t batch_size) {
-  size_t num_decoded = 0;
-
-  // TODO(wesm): Push this decoding down into RleDecoder itself
-  for (size_t i = 0; i < batch_size; ++i) {
-    if (!decoder->Get(levels + i)) {
-      break;
-    }
-    ++num_decoded;
-  }
-  return num_decoded;
-}
-
 size_t ColumnReader::ReadDefinitionLevels(size_t batch_size, int16_t* levels) {
-  if (!definition_level_decoder_) {
+  if (descr_->max_definition_level() == 0) {
     return 0;
   }
-  return DecodeMany(definition_level_decoder_.get(), levels, batch_size);
+  return definition_level_decoder_.Decode(batch_size, levels);
 }
 
 size_t ColumnReader::ReadRepetitionLevels(size_t batch_size, int16_t* levels) {
-  if (!repetition_level_decoder_) {
+  if (descr_->max_repetition_level() == 0) {
     return 0;
   }
-  return DecodeMany(repetition_level_decoder_.get(), levels, batch_size);
+  return repetition_level_decoder_.Decode(batch_size, levels);
 }
 
 // ----------------------------------------------------------------------
--- a/src/parquet/column/reader.h
+++ b/src/parquet/column/reader.h
@@ -33,8 +33,10 @@
 #include "parquet/encodings/encodings.h"
 #include "parquet/schema/descriptor.h"
 #include "parquet/util/rle-encoding.h"
+#include "parquet/column/levels.h"
 
 namespace parquet_cpp {
+
 
 class Codec;
 class Scanner;
@@ -85,13 +87,10 @@ class ColumnReader {
   std::shared_ptr<Page> current_page_;
 
   // Not set if full schema for this field has no optional or repeated elements
-  std::unique_ptr<RleDecoder> definition_level_decoder_;
+  LevelDecoder definition_level_decoder_;
 
   // Not set for flat schemas.
-  std::unique_ptr<RleDecoder> repetition_level_decoder_;
-
-  // Temporarily storing this to assist with batch reading
-  int16_t max_definition_level_;
+  LevelDecoder repetition_level_decoder_;
 
   // The total number of values stored in the data page. This is the maximum of
   // the number of encoded definition levels or encoded values. For
@@ -182,13 +181,12 @@ inline size_t TypedColumnReader<TYPE>::ReadBatch(int batch_size, int16_t* def_le
   size_t values_to_read = 0;
 
   // If the field is required and non-repeated, there are no definition levels
-  if (definition_level_decoder_) {
+  if (descr_->max_definition_level() > 0) {
     num_def_levels = ReadDefinitionLevels(batch_size, def_levels);
-
     // TODO(wesm): this tallying of values-to-decode can be performed with better
     // cache-efficiency if fused with the level decoding.
     for (size_t i = 0; i < num_def_levels; ++i) {
-      if (def_levels[i] == max_definition_level_) {
+      if (def_levels[i] == descr_->max_definition_level()) {
         ++values_to_read;
       }
     }
@@ -198,9 +196,8 @@ inline size_t TypedColumnReader<TYPE>::ReadBatch(int batch_size, int16_t* def_le
   }
 
   // Not present for non-repeated fields
-  if (repetition_level_decoder_) {
+  if (descr_->max_repetition_level() > 0) {
     num_rep_levels = ReadRepetitionLevels(batch_size, rep_levels);
-
     if (num_def_levels != num_rep_levels) {
       throw ParquetException("Number of decoded rep / def levels did not match");
     }
--- a/src/parquet/reader.cc
+++ b/src/parquet/reader.cc
@@ -242,10 +242,10 @@ void ParquetFileReader::DebugPrint(std::ostream& stream, bool print_values) {
            << ")" << std::endl;
   }
 
-  for (int i = 0; i < num_row_groups(); ++i) {
-    stream << "--- Row Group " << i << " ---\n";
+  for (int r = 0; r < num_row_groups(); ++r) {
+    stream << "--- Row Group " << r << " ---\n";
 
-    RowGroupReader* group_reader = RowGroup(i);
+    RowGroupReader* group_reader = RowGroup(r);
 
     // Print column metadata
     size_t num_columns = group_reader->num_columns();
--- a/src/parquet/column/CMakeLists.txt
+++ b/src/parquet/column/CMakeLists.txt
@@ -18,9 +18,11 @@
 # Headers: top level
 install(FILES
   page.h
+  levels.h
   reader.h
   serialized-page.h
   scanner.h
   DESTINATION include/parquet/column)
 
 ADD_PARQUET_TEST(column-reader-test)
+ADD_PARQUET_TEST(levels-test)
--- a/src/parquet/column/page.h
+++ b/src/parquet/column/page.h
@@ -84,6 +84,14 @@ class DataPage : public Page {
     return header_.encoding;
   }
 
+  parquet::Encoding::type repetition_level_encoding() const {
+    return header_.repetition_level_encoding;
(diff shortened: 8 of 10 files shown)


wesm:
Can you rebase on #26 ? Also, `BIT_PACKED` decoding is currently unsupported. Let's tackle in a separate patch per PARQUET-462

majetideepak:
Sure!

src/parquet/column_reader.cc
```diff
@@ -114,18 +114,51 @@ bool TypedColumnReader<TYPE>::ReadNewPage() {
...
+          if (encoding == parquet::Encoding::RLE || encoding == parquet::Encoding::BIT_PACKED) {
+            num_repetition_bytes = *reinterpret_cast<const uint32_t*>(buffer);
+            buffer += sizeof(uint32_t);
+            uncompressed_len -= sizeof(uint32_t);
+            repetition_level_decoder_.reset(
+	        new RleDecoder(buffer, num_repetition_bytes, rep_bit_width));
+            buffer += num_repetition_bytes;
+            uncompressed_len -= num_repetition_bytes;
+          } else {
+            throw ParquetException("Unknown encoding type for levels.");
```
wesm: Can you refactor this to avoid code duplication? Perhaps combine this effort to create a patch for PARQUET-462 to kill two birds with one stone

majetideepak: Thats the plan.

majetideepak:
It is much easier now if I rebase on #31

majetideepak:
@wesm can you give your comments on the level decoder class ? Thanks!

src/parquet/column/reader.cc
```diff
@@ -115,25 +115,31 @@ bool TypedColumnReader<TYPE>::ReadNewPage() {
...
+        if (max_repetition_level_ > 0) {
+           repetition_level_decoder_.Init(
+                  current_page_header_.data_page_header.repetition_level_encoding,
+                  max_repetition_level_, num_buffered_values_, &buffer, &uncompressed_len);
+        }
+        //Init definition levels
+        if (max_definition_level_ > 0) {
+           definition_level_decoder_.Init(
+                  current_page_header_.data_page_header.definition_level_encoding,
+                  max_definition_level_, num_buffered_values_, &buffer, &uncompressed_len);
```
wesm: indentation problems here

src/parquet/column/reader.cc
```diff
@@ -176,12 +182,12 @@ bool TypedColumnReader<TYPE>::ReadNewPage() {
 // ----------------------------------------------------------------------
 // Batch read APIs
 
-static size_t DecodeMany(RleDecoder* decoder, int16_t* levels, size_t batch_size) {
+static size_t DecodeMany(LevelDecoder &decoder, int16_t* levels, size_t batch_size) {
```
wesm: Can you push this code down into into `LevelDecoder`? The `Decode` API for LevelDecoder should be array oriented to be consistent with the other decoders in `parquet/encodings`

src/parquet/column/reader.h
```diff
@@ -47,6 +47,66 @@ struct hash<parquet::Encoding::type> {
 
 namespace parquet_cpp {
 
+class LevelDecoder {
+
+ private:
```
wesm: Private variables go at the end of the class

src/parquet/column/reader.h
```diff
@@ -47,6 +47,66 @@ struct hash<parquet::Encoding::type> {
...
+  parquet::Encoding::type encoding_;
+  std::unique_ptr<RleDecoder> rle_decoder;
+  std::unique_ptr<BitReader> bit_packed_decoder;
```
wesm: Member variables have a `_` postfix

src/parquet/column/reader.h
```diff
@@ -47,6 +47,66 @@ struct hash<parquet::Encoding::type> {
...
+
+ public:
+  LevelDecoder() {}
+
+  //return true if one of the decoder is set
+  bool set() {
+     return ((NULL != rle_decoder.get()) || (NULL != bit_packed_decoder.get()));
```
wesm: Indentation. Also, `NULL` is not supposed to be used in C++11

src/parquet/column/reader.cc
```diff
@@ -190,17 +196,17 @@ static size_t DecodeMany(RleDecoder* decoder, int16_t* levels, size_t batch_size
 }
 
 size_t ColumnReader::ReadDefinitionLevels(size_t batch_size, int16_t* levels) {
-  if (!definition_level_decoder_) {
+  if (!definition_level_decoder_.set()) {
```
wesm: I think it would be better to check whether the `max_definition_level_` is greater than 0 here. The contract about the existence of the definition level decoder should be enforced elsewhere

src/parquet/column/reader.h
wesm: Recommend removing this method per my comment above

src/parquet/column/reader.h
```diff
@@ -47,6 +47,66 @@ struct hash<parquet::Encoding::type> {
...
+  }
+
+  // Initialize the LevelDecoder.
+  void Init(parquet::Encoding::type encoding, int16_t max_level, int num_buffered_values,
+             const uint8_t** data, int* data_size) {
+
+     uint32_t num_bytes = 0;
```
wesm: Indentation uses two spaces

src/parquet/column/reader.h
```diff
@@ -47,6 +47,66 @@ struct hash<parquet::Encoding::type> {
...
+  LevelDecoder() {}
+
+  //return true if one of the decoder is set
+  bool set() {
+     return ((NULL != rle_decoder.get()) || (NULL != bit_packed_decoder.get()));
+  }
+
+  // Initialize the LevelDecoder.
+  void Init(parquet::Encoding::type encoding, int16_t max_level, int num_buffered_values,
+             const uint8_t** data, int* data_size) {
```
wesm: Inconsistent indentation

wesm: Modifying `data` and `data_size` is not a good pattern to use. Instead, `Init` should accept `const uint8_t* data` and `size_t data_size` and return `size_t` (instead of `void`), indicating the number of bytes consumed by initializing the level decoder.

src/parquet/column/reader.h
```diff
@@ -47,6 +47,66 @@ struct hash<parquet::Encoding::type> {
...
+
+     uint32_t num_bytes = 0;
+     encoding_ = encoding;
+     switch (encoding) {
+       case parquet::Encoding::RLE: {
+         num_bytes = *reinterpret_cast<const uint32_t*>(*data);
+         *data += sizeof(uint32_t);
+         *data_size -= sizeof(uint32_t);
```
wesm: remove this per above

src/parquet/column/reader.h
```diff
@@ -47,6 +47,66 @@ struct hash<parquet::Encoding::type> {
...
+       case parquet::Encoding::BIT_PACKED: {
+         num_bytes = BitUtil::Ceil(num_buffered_values, 8);
+         bit_packed_decoder.reset(new BitReader(*data, num_bytes));
+         break;
+       }
+       default:
+         throw ParquetException("Unknown encoding type for levels.");
+     }
+     *data += num_bytes;
+     *data_size += num_bytes;
```
wesm: remove this. return `num_bytes` instead

src/parquet/column/reader.h
```diff
@@ -47,6 +47,66 @@ struct hash<parquet::Encoding::type> {
...
+
+  /// Returns the next level.
+  inline int16_t ReadLevel(int16_t *level){
+
+    bool valid;
+    if (encoding_ == parquet::Encoding::RLE) {
+          valid = rle_decoder.get()->Get(level);
+     } else {
+          assert(encoding_ == parquet::Encoding::BIT_PACKED);
+          valid = bit_packed_decoder.get()->GetValue(1, level);
```
wesm: indentation

src/parquet/column/reader.h
```diff
@@ -47,6 +47,66 @@ struct hash<parquet::Encoding::type> {
...
+       }
+       default:
+         throw ParquetException("Unknown encoding type for levels.");
+     }
+     *data += num_bytes;
+     *data_size += num_bytes;
+  }
+
+  /// Returns the next level.
+  inline int16_t ReadLevel(int16_t *level){
```
wesm: Per discussion above, the public API for this class should be array oriented (with `batch_size` as a parameter) and return the actual number of levels read (in case the decoder reaches EOS before reaching the batch size -- will need some tests around this)

src/parquet/column/reader.h
```diff
@@ -47,6 +47,66 @@ struct hash<parquet::Encoding::type> {
...
+
+    bool valid;
+    if (encoding_ == parquet::Encoding::RLE) {
+          valid = rle_decoder.get()->Get(level);
```
wesm: you don't need `.get()` here or below, you can use `->` on a `unique_ptr`

src/parquet/column/reader.h
```diff
@@ -214,12 +276,12 @@ inline size_t TypedColumnReader<TYPE>::ReadBatch(int batch_size, int16_t* def_le
   size_t num_rep_levels = 0;
 
   // If the field is required and non-repeated, there are no definition levels
-  if (definition_level_decoder_) {
+  if (definition_level_decoder_.set()) {
```
wesm: per above, check `max_definition_level_` instead

src/parquet/column/reader.h
```diff
@@ -214,12 +276,12 @@ inline size_t TypedColumnReader<TYPE>::ReadBatch(int batch_size, int16_t* def_le
...
     num_def_levels = ReadDefinitionLevels(batch_size, def_levels);
   }
 
   // Not present for non-repeated fields
-  if (repetition_level_decoder_) {
+  if (repetition_level_decoder_.set()) {
```
wesm: ditto

src/parquet/column/reader.h
```diff
@@ -47,6 +47,66 @@ struct hash<parquet::Encoding::type> {
 
 namespace parquet_cpp {
 
+class LevelDecoder {
```
wesm: Can you place this class in its own header file `src/parquet/column/levels.h`. We will also start a unit test suite for the level encoders and decoders in `src/parquet/column/levels-test.cc`

wesm:
On code style, can you figure out a solution to prevent these indentation errors? We obviously need to update cpplint and incorporate clang-tidy to automatically catch these things.

Here is what my emacs setup looks like

```
(require 'google-c-style)

(defconst my-cc-style
  '("Google" (c-offsets-alist .
                           ((innamespace . [0])
                            (arglist-cont-nonempty . 4)
                            (arglist-intro . 4)
                            (access-label . -1)
                            (member-init-intro . 4))
                           )))
(c-add-style "my-cc-style" my-cc-style)
(defun set-my-cc-style () (interactive) (google-set-c-style) (c-set-style "my-cc-style"))
(add-hook 'c-mode-common-hook 'set-my-cc-style)
```

wesm:
Let's use this patch to start a test suite that verifies level decoding as isolation. Please put these tests in `src/parquet/column/levels-test.cc` per above and add it to the test suite in the `CMakeLists.txt` in that directory.

To facilitate the testing, you'll want to create a `LevelEncoder` class that mirrors the `LevelDecoder` to generate bytes containing encoded levels. It should be able to generate both `Encoding::RLE` and `Encoding::BIT_PACKED` data. The `LevelEncoder` constructor should accept the `max_level` to determine the minimum bit width required to encode the data.

You may wish to rebase on #32, but the current conflicts are not too onerous, so it's up to you.

Once this is complete we can also resolve PARQUET-462.

Let me know if you need help. thank you!

src/parquet/column/reader.h
wesm: to be clear, the API should be

`size_t Decode(size_t batch_size, int16_t* out)`

wesm: Also, note that `inline` is redundant within a class in a header

majetideepak:
@wesm thanks for the comments. You have some good points. I am getting familiar with the coding style. I will make the changes and update with the test cases.

wesm:
Refer to https://google.github.io/styleguide/cppguide.html for more on style and conventions

wesm:
@majetideepak also, it is better to exclusively make pull requests out of git branches (this is incoming from your master branch)

wesm:
@majetideepak you will want to rebase to pick up the new `cpplint` configuration

majetideepak:
@wesm I rebased the new cpplint configuration. Thanks!
I will base my future pull requests from git branches.

majetideepak:
@wesm I incorporated your changes and ran the new cpplint configuration. Let me know your comments. Thanks!

src/parquet/column/levels.h
```diff
@@ -0,0 +1,183 @@
...
+      }
+      default:
+        throw ParquetException("Unknown encoding type for levels.");
+    }
+    total_bytes += num_bytes;
+    return total_bytes;
+  }
+
+  //Returns the next level.
+  int16_t ReadLevel(int16_t *level) {
```
wesm: Remove this API

src/parquet/column/reader.cc
```diff
@@ -176,31 +185,18 @@ bool TypedColumnReader<TYPE>::ReadNewPage() {
...
-      break;
-    }
-    ++num_decoded;
-  }
-  return num_decoded;
-}
-
 size_t ColumnReader::ReadDefinitionLevels(size_t batch_size, int16_t* levels) {
-  if (!definition_level_decoder_) {
+  if (max_definition_level_ < 1) {
```
wesm: `== 0`, also below

src/parquet/column/levels.h
```diff
@@ -0,0 +1,183 @@
...
+    } else {
+      assert(encoding_ == parquet::Encoding::BIT_PACKED);
+      valid = bit_packed_decoder_->GetValue(bit_width_, level);
+    }
+    //valid is 1 for correct values and -1 otherwise
+    return valid;
+  }
+
+  //Decodes a batch of levels into an array and returns the number of levels decoded
+  size_t DecodeMany(size_t batch_size, int16_t* levels) {
```
wesm: call this just `Decode` since all our APIs are array-oriented

src/parquet/column/levels.h
```diff
@@ -0,0 +1,183 @@
...
+
+    if (encoding_ == parquet::Encoding::RLE) {
+      for (size_t i = 0; i < batch_size; ++i) {
+        if (!rle_decoder_->Get(levels + i)) {
+          break;
+        }
+        ++num_decoded;
+      }
+    } else {
+      assert(encoding_ == parquet::Encoding::BIT_PACKED);
```
wesm: Safe to omit this

src/parquet/column/levels-test.cc
```diff
@@ -0,0 +1,112 @@
...
+
+    bit_packed_encoder_.Init(parquet::Encoding::BIT_PACKED, max_level_, NUM_LEVELS,
+         bit_packed_bytes_.data(), sizeof(uint8_t)*NUM_LEVELS);
+
+    num_rle_levels_ = rle_encoder_.EncodeMany(NUM_LEVELS, level_values_.data());
+    num_bit_packed_levels_ = bit_packed_encoder_.EncodeMany(NUM_LEVELS,
+        level_values_.data());
+
+    int32_t rle_length = rle_encoder_.getRleLength();
+    (reinterpret_cast<int32_t*>(rle_bytes_.data()))[0] = rle_length;
```
wesm: This needs to be a `uint32_t`

src/parquet/column/levels-test.cc
```diff
@@ -0,0 +1,112 @@
...
+
+  rle_decoder_.Init(parquet::Encoding::RLE, max_level_, NUM_LEVELS,
+       rle_bytes_.data());
+
+  int num_values = rle_decoder_.DecodeMany(NUM_LEVELS, level_values_.data());
+
+  ASSERT_EQ(NUM_LEVELS, num_values);
+
+  for (int i = 0; i < NUM_LEVELS; i++) {
+    ASSERT_EQ(level_values_[i], (i < NUM_LEVELS/2) ? 1 : 2);
```
wesm: You are decoding into the same buffer where you generated the data in the first place. How do you know that `level_values_` was actually written to? =)

please revise to:

* random data (C++11 `<random>` helps a lot here!)
* decode into new memory in the test case

src/parquet/column/levels-test.cc
```diff
@@ -0,0 +1,112 @@
...
+
+  bit_packed_decoder_.Init(parquet::Encoding::BIT_PACKED, max_level_, NUM_LEVELS,
+       bit_packed_bytes_.data());
+
+  int num_values = bit_packed_decoder_.DecodeMany(NUM_LEVELS, level_values_.data());
+
+  ASSERT_EQ(NUM_LEVELS, num_values);
+
+  for (int i = 0; i < NUM_LEVELS; i++) {
+    ASSERT_EQ(level_values_[i], (i < NUM_LEVELS/2) ? 1 : 2);
```
wesm: same problem as above. decode into a new vector and compare with `level_values_`

src/parquet/column/levels-test.cc
wesm: Actually scratch that re: random data. see my upcoming note

wesm:
Cool this is looking a lot better, thank you. I'm not done reviewing `levels.h`, but the gap in the unit tests is that you are only testing one rep/def levels in the span of 1 to 2. 

Can you test the following bit widths: 1 (0 or 1), 2, 4, 8 (0 through 127). To test that RLE encoding is working properly, can you devise a number generation algorithm that yields runs with at least 8 elements (maybe at least 32 or 64-length runs to be safe)? Sticking with ~1000 levels seems fine.

wesm:
I'll finish reviewing the rest of the code after these tests are completed.

src/parquet/column/levels-test.cc
wesm: oops I missed the `level_values_` clear, sorry about that. I still think it would be better to decode into new memory to make the tests more flexible

wesm:
@nongli can you comment on the test data generation advice I gave here?

src/parquet/column/reader.cc
majetideepak: I used < 1 to handle negative values as well.

wesm: This should never happen, though, right? Slippery slope as a bug elsewhere could slip through here by accident.

wesm: Better to add a dcheck elsewhere asserting it is >= 0

wesm:
@majetideepak Cool, you'll want to rebase this on master now; since I factored out the decoder initialization hopefully reconciling the changes will not be too onerous

majetideepak:
@wesm Rebased and addressed your comments. Let me know if you have more feedback. Thanks!

src/parquet/column/levels-test.cc
```diff
@@ -0,0 +1,193 @@
...
+  std::vector<int16_t> level_values_;
+  std::vector<int16_t> output_levels_;
+  LevelEncoder rle_encoder_;
+  LevelDecoder rle_decoder_;
+  LevelEncoder bit_packed_encoder_;
+  LevelDecoder bit_packed_decoder_;
+  std::vector<uint8_t> rle_bytes_;
+  std::vector<uint8_t> bit_packed_bytes_;
+  int num_rle_levels_;
+  int num_bit_packed_levels_;
```
wesm: These last 8 variables are superfluous (only need 4) because the class is instantiated for every test case.

src/parquet/column/levels-test.cc
```diff
@@ -0,0 +1,193 @@
...
+
+const int NUM_LEVELS = 1024;
+
+class TestLevels : public ::testing::Test {
+ public:
+  void SetUp() {
+    level_values_.resize(NUM_LEVELS);
+    output_levels_.resize(NUM_LEVELS);
+    rle_bytes_.resize(2*NUM_LEVELS);
+    bit_packed_bytes_.resize(2*NUM_LEVELS);
```
wesm: Can you make `NUM_LEVELS` a parameter? e.g. moving these to a method `void Init(size_t num_levels)`

src/parquet/column/levels-test.cc
```diff
@@ -0,0 +1,193 @@
...
+#include <gtest/gtest.h>
+
+#include "parquet/thrift/parquet_types.h"
+#include "parquet/column/levels.h"
+
+using std::string;
+
+namespace parquet_cpp {
+
+const int NUM_LEVELS = 1024;
```
wesm: Use this locally in the test cases (this will make it easier to create tests with a different parameter)

src/parquet/column/levels-test.cc
```diff
@@ -0,0 +1,193 @@
...
+TEST_F(TestLevels, TestLevelsBitPacked2to8) {
+  //for each maximum bit-width
+  for (int bit_width = 2; bit_width <= 8; bit_width++) {
+    int value = 2;
+    //generate levels uniformly upto the maximum bit-width
+    for (int i = 0; i < NUM_LEVELS; i++) {
+      level_values_[i] = ((value << (i/(NUM_LEVELS/bit_width))) - 1);
+    }
+    int max_level = level_values_[NUM_LEVELS - 1];
+    bit_packed_encoder_.Init(parquet::Encoding::BIT_PACKED, max_level, NUM_LEVELS,
```
wesm: Can you make `Encoding::BIT_PACKED` (and `Encoding::RLE`) a parameter instead of hard-coding so you can avoid code duplication?

src/parquet/column/levels-test.cc
```diff
@@ -0,0 +1,193 @@
...
+  }
+
+  rle_bytes_.clear();
+  level_values_.clear();
+  output_levels_.clear();
+}
+
+//test levels with bit_width = 1
+//[0,0,0,...1,1,1,..]
+TEST_F(TestLevels, TestLevelsBitPacked1) {
```
wesm: why is `bit_width == 1` a special case?

src/parquet/column/levels-test.cc
```diff
@@ -0,0 +1,193 @@
...
+
+  ASSERT_EQ(NUM_LEVELS, num_values);
+
+  for (int i = 0; i < NUM_LEVELS; i++) {
+    ASSERT_EQ(level_values_[i], output_levels_[i]);
+  }
+
+  rle_bytes_.clear();
+  level_values_.clear();
+  output_levels_.clear();
```
wesm: this isn't necessary

src/parquet/column/levels-test.cc
```diff
@@ -0,0 +1,193 @@
...
+
+//test levels with maximum bit-width from 2 to 8
+//[1,1,1,...3,3,3,..,7,7,7...,255,255,255]
+TEST_F(TestLevels, TestLevelsRle2to8) {
+  //for each maximum bit-width
+  for (int bit_width = 2; bit_width <= 8; bit_width++) {
+    int value = 2;
+    //generate levels uniformly upto the maximum bit-width
+    for (int i = 0; i < NUM_LEVELS; i++) {
+      level_values_[i] = ((value << (i/(NUM_LEVELS/bit_width))) - 1);
```
wesm: I find this pretty inscrutable. It would be clearer for readers of the code if you declared run-sizes of values that fit in the indicated bit width then generated consecutive sequences of those run sizes (e.g. 1, 2, 4, 8, 16, 32, 64) until you reach `NUM_LEVELS`.

src/parquet/column/levels-test.cc
```diff
@@ -0,0 +1,193 @@
...
+  LevelDecoder rle_decoder_;
+  LevelEncoder bit_packed_encoder_;
+  LevelDecoder bit_packed_decoder_;
+  std::vector<uint8_t> rle_bytes_;
+  std::vector<uint8_t> bit_packed_bytes_;
+  int num_rle_levels_;
+  int num_bit_packed_levels_;
+};
+//test levels with bit_width = 1
+//[0,0,0,...1,1,1,..]
```
wesm: I guess cpplint doesn't complain about this, and I'm not sure if the Google style guide provides a guideline, but I think comments are easier to read with a space between the `//` and the test. This is consistent with the rest of the codebase and more readable IMHO, would you mind terribly? Thanks

src/parquet/column/levels.h
```diff
@@ -0,0 +1,155 @@
...
+
+namespace parquet_cpp {
+
+class LevelEncoder {
+ public:
+  LevelEncoder() {}
+
+  // Initialize the LevelEncoder.
+  void Init(parquet::Encoding::type encoding, int16_t max_level,
+              int num_buffered_values, uint8_t* data, int data_size) {
```
wesm: indentation

src/parquet/column/levels.h
```diff
@@ -0,0 +1,155 @@
...
+    bit_width_ = BitUtil::Log2(max_level + 1);
+    encoding_ = encoding;
+    switch (encoding) {
+      case parquet::Encoding::RLE: {
+        rle_encoder_.reset(new RleEncoder(data, data_size, bit_width_));
+        break;
+      }
+      case parquet::Encoding::BIT_PACKED: {
+        int num_bytes = BitUtil::Ceil(num_buffered_values*bit_width_, 8);
```
wesm: whitespace

src/parquet/column/levels.h
```diff
@@ -0,0 +1,155 @@
...
+        bit_packed_encoder_.reset(new BitWriter(data, num_bytes));
+        break;
+      }
+      default:
+        throw ParquetException("Unknown encoding type for levels.");
+    }
+  }
+
+  //Encodes a batch of levels from an array and returns the number of levels encoded
```
wesm: whitespace

src/parquet/column/levels.h
```diff
@@ -0,0 +1,155 @@
...
+          break;
+        }
+        ++num_encoded;
+      }
+      bit_packed_encoder_->Flush();
+    }
+    return num_encoded;
+  }
+
+  int32_t getRleLength() {
```
wesm: Please do not use camel case (https://google.github.io/styleguide/cppguide.html#Function_Names). In `rle-encoding.h` the equivalent method is called `len()` so maybe let's use that here for now.

src/parquet/column/levels.h
```diff
@@ -0,0 +1,155 @@
...
+};
+
+
+class LevelDecoder {
+ public:
+  LevelDecoder() {}
+
+  // Initialize the LevelDecoder.
+  size_t Init(parquet::Encoding::type encoding, int16_t max_level,
+              int num_buffered_values, const uint8_t* data) {
```
wesm: indentation (this line should be indented 4 spaces right of the indentation level of the function signature)

src/parquet/column/levels.h
```diff
@@ -0,0 +1,155 @@
...
+    uint32_t num_bytes = 0;
+    uint32_t total_bytes = 0;
+    bit_width_ = BitUtil::Log2(max_level + 1);
+    encoding_ = encoding;
+    switch (encoding) {
+      case parquet::Encoding::RLE: {
+        num_bytes = *reinterpret_cast<const uint32_t*>(data);
+        const uint8_t* decoder_data = data + sizeof(uint32_t);
+        rle_decoder_.reset(new RleDecoder(decoder_data, num_bytes, bit_width_));
+        total_bytes += sizeof(uint32_t);
```
wesm: `total_bytes = sizeof(uint32_t) + num_bytes`

src/parquet/column/levels.h
```diff
@@ -0,0 +1,155 @@
...
+        break;
+      }
+      case parquet::Encoding::BIT_PACKED: {
+        num_bytes = BitUtil::Ceil(num_buffered_values*bit_width_, 8);
```
wesm: Set `total_bytes` here instead (to match above)

src/parquet/column/levels.h
```diff
@@ -0,0 +1,155 @@
...
+        bit_packed_decoder_.reset(new BitReader(data, num_bytes));
+        break;
+      }
+      default:
+        throw ParquetException("Unknown encoding type for levels.");
+    }
+    total_bytes += num_bytes;
```
wesm: You can remove this with the above changes

src/parquet/column/levels.h
```diff
@@ -0,0 +1,155 @@
...
+    return total_bytes;
+  }
+
+  //Decodes a batch of levels into an array and returns the number of levels decoded
```
wesm: whitespace

src/parquet/column/levels.h
```diff
@@ -0,0 +1,155 @@
...
+  size_t Decode(size_t batch_size, int16_t* levels) {
+    size_t num_decoded = 0;
+    if (!rle_decoder_ && !bit_packed_decoder_)
+      throw ParquetException("Level decoders are not initialized.");
```
wesm: missing braces

wesm:
In addition to the code comments, can you construct a test case in `parquet/column/column-reader-test.cc` that verifies that repetition levels can be correctly read from a data page? Start with `TestInt32FlatOptional` and add distinct rep levels so they can be verified alongside the definition levels and values. I know there's a little code duplication there; we can return to refactor it for better code reuse and expand to a broader set of test cases. 

thank you sir!

src/parquet/column/levels-test.cc
wesm: Please let me know if this isn't clear. I'm looking to see data that is clearly like

`[0, 1, 1, 0, 0, 0, 0, 1, 1, 1, 1, 1, 1, 1, 1, ...]`

and so forth, where the actual levels are chosen so that they fit in the desired bit width.

src/parquet/column/levels.h
majetideepak: https://google.github.io/styleguide/cppguide.html#Function_Declarations_and_Definitions
also aligns with the first parameter (style 2).

src/parquet/column/levels-test.cc
wesm: (and this code may already be doing this but it isn't obvious without carefully scrutinizing the math)

majetideepak:
@wesm  thanks for the comments. Sorry for the formatting issues. Neither I nor cpplint could catch them. I will modify the tests.

src/parquet/column/levels.h
wesm: This is true. Impala conforms on style 3, but I'm fine with style 2 for short function names, as long we don't do things like

```
XXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXX(YYYYYYYYYYYYYYYYYYYYYYYYYY,
                                                       YYYYYYYYYYYYYYYYYYYYYYYYYY,
                                                       YYYYYYYYYYYYYYYYYYYYYYYYYY)
```

from an editor perspective, it's hard to get Emacs or VIM to do *both* styles 2 and 3, so it's easier to standardize on 3 (because return value + function name is often very long) and avoid unnecessary diff noise. but this one is fine for now.

src/parquet/column/levels-test.cc
majetideepak: @wesm the code above generates the following sequences
bit_width = 2
`[1, 1, 1, ....(512 times), 3, 3, 3,...(512 times)]`
bit_width = 3
......
bit_width = 4
`[1, 1, 1, ....(256 times), 3, 3, 3,...(256 times), 7, 7, 7 .... (256 times), 15, 15, 15 ... (256 times)]`
so on.

majetideepak: It does capture your pattern but it does not intersperse them.

src/parquet/column/levels.h
majetideepak: We can stick to style 3 for compatibility. I will change it.

src/parquet/column/levels-test.cc
wesm: It would be a stronger test case if you use a mix of different run lengths (make sure to include some literal runs without repeated elements to hit the literal run code path for RLE) with the levels for each bit width rather than use the same run length for all elements in each bit width. 

My intent here is to make sure we can feel confident about refactoring the innards of `LevelDecoder` or `LevelEncoder` to improve performance. Defensive unit testing!

src/parquet/column/levels-test.cc
```diff
@@ -0,0 +1,122 @@
...
+      ASSERT_EQ(levels_count, num_levels);
+
+      decoder.Init(encoding, max_level, num_levels, bytes_.data());
+      levels_count = decoder.Decode(num_levels, output_levels_.data());
+
+      ASSERT_EQ(levels_count, num_levels);
+
+      for (int i = 0; i < num_levels; i++) {
+        EXPECT_EQ(input_levels_[i], output_levels_[i]);
+      }
```
wesm: can you turn this into a reusable testing function that accepts const std::vector<int16_t>& levels and the encoding, then verifies that the values survive the round trip? This way we can more easily expand the test cases and different types of data generation in the future.

src/parquet/column/levels-test.cc
```diff
@@ -0,0 +1,122 @@
...
+        int value = 0;
+        int bwidth = 0;
+        while (value <= max_level) {
+          for (int i = 0; i < repeat_count; i++) {
+            input_levels_[total_count++] = value;
+          }
+          value = (2 << bwidth) - 1;
+          bwidth++;
+        }
+      }
```
wesm: Can you factor this data generation into a standalone function?

wesm:
I hope you'll pardon my clean code obsession, it's worth it, I promise! We are getting close — I will rebase #38 on this after it's merged.

majetideepak:
@wesm not a problem at all :+1: . I am still modifying the `TestInt32FlatOptional` in parquet/column/column-reader-test.cc and the test fails even for `max_repetition_level = 0`. I am looking into that.

src/parquet/column/levels-test.cc
```diff
@@ -0,0 +1,133 @@
...
+
+  void TearDown() {}
+
+ protected:
+  std::vector<int16_t> input_levels_;
+  std::vector<int16_t> output_levels_;
+  std::vector<uint8_t> bytes_;
+  int max_level_;
+  int num_levels_;
+};
```
majetideepak: I added `max_level_` and `num_levels_` as class variables since multiple methods require them.

src/parquet/column/levels-test.cc
```diff
@@ -0,0 +1,133 @@
...
+  int max_repeat_factor = 7; // 128
+  int max_bit_width = 8;
+  parquet::Encoding::type encodings[2] = {parquet::Encoding::RLE,
+      parquet::Encoding::BIT_PACKED};
+
+  // for each encoding
+  for (int encode = 0; encode < 2; encode++) {
+    parquet::Encoding::type encoding = encodings[encode];
+    //BIT_PACKED requires a sequence of atleast 8
+    if ((encoding == parquet::Encoding::BIT_PACKED)) min_repeat_factor = 3;
```
majetideepak: Will fix this space.

src/parquet/column/levels-test.cc
```diff
@@ -0,0 +1,133 @@
...
+  int min_repeat_factor = 0;
+  int max_repeat_factor = 7; // 128
+  int max_bit_width = 8;
+  parquet::Encoding::type encodings[2] = {parquet::Encoding::RLE,
+      parquet::Encoding::BIT_PACKED};
+
+  // for each encoding
+  for (int encode = 0; encode < 2; encode++) {
+    parquet::Encoding::type encoding = encodings[encode];
+    //BIT_PACKED requires a sequence of atleast 8
```
majetideepak: will fix whitespace

src/parquet/column/levels-test.cc
```diff
@@ -0,0 +1,133 @@
...
+
+class TestLevels : public ::testing::Test {
+ public:
+  void InitBuffers() {
+    input_levels_.resize(num_levels_);
+    output_levels_.resize(num_levels_);
+    bytes_.resize(2 * num_levels_);
+  }
+
+  int GenerateLevels(int min_repeat_factor, int max_repeat_factor) {
```
wesm: having this as a static method outside of the class would be better

src/parquet/column/levels-test.cc
```diff
@@ -0,0 +1,133 @@
...
+          input_levels_[total_count++] = value;
+        }
+        value = (2 << bwidth) - 1;
+        bwidth++;
+      }
+    }
+    return total_count;
+  }
+
+  void VerifyLevelsEncoding(parquet::Encoding::type encoding) {
```
wesm: same for this method

src/parquet/column/levels-test.cc
```diff
@@ -0,0 +1,133 @@
...
+  void SetUp() {}
+
+  void TearDown() {}
+
+ protected:
+  std::vector<int16_t> input_levels_;
+  std::vector<int16_t> output_levels_;
+  std::vector<uint8_t> bytes_;
+  int max_level_;
+  int num_levels_;
```
wesm: I'm not sure that this class is required at all if you make the data generation and verification plain old functions (without relying on any state)

majetideepak:
@wesm does it look okay now? Should we also open another JIRA to refactor/extend the existing code (`src/parquet/column/test-util.h`) and other tests (`src/parquet/column/column-reader-test.cc`) based on this feature ? Merging this PR will help you move faster.

wesm:
Yes — can you add a unit test that checks the repetition level read path in `column-reader-test`? I don't see that here yet. As soon as that's here and passing this will be ready to merge.

wesm:
Otherwise we have no way of knowing that repetition levels can actually be read.

wesm:
To give you some more insight on the testing philosophy I'm expressing:

Unit testing is about two big things

* Verifying correctness, freedom from bugs
* Enabling freedom from refactoring fear (humorous, but insightful, on this topic: https://www.youtube.com/watch?v=YX3iRjKj7C0)

Since we are collaborating and working on different parts of the codebases, if a component is not well tested, I could change its functionality or break it completely without even knowing! In this particular case, we already have a merge conflict that has to be resolved. What if I accidentally broke the code that you worked on? That makes me extremely uncomfortable. 

So: safeguarding code that you write with tests is about making the software correct and enabling your collaborators to refactor your code without being worried about it. Unless a piece of code is write-once-never-touch-it-again, refactoring will always happen. I have occasionally seen engineers express the notion that refactoring means you "didn't get it right the first time". This is total nonsense. Refactoring is good, and should be encouraged if it makes the code cleaner, more reusable, or faster.

wesm:
@majetideepak can you rebase first thing next week and let me know when you have the additional unit test? Thank you very much

src/parquet/column/column-reader-test.cc
```diff
@@ -157,6 +156,65 @@ TEST_F(TestPrimitiveReader, TestInt32FlatOptional) {
...
+
+  size_t num_values = values.size();
+  Encoding::type value_encoding = Encoding::PLAIN;
+
+  vector<uint8_t> page1;
+  test::DataPageBuilder<Type::INT32> page_builder(&page1);
+
+  // Definition levels precede the values
+  page_builder.AppendRepLevels(rep_levels, 1, Encoding::RLE);
+  page_builder.AppendDefLevels(def_levels, 2, Encoding::RLE);
```
wesm: Is this order correct?

src/parquet/column/reader.h
```diff
@@ -66,6 +68,11 @@ class ColumnReader {
     return descr_;
   }
 
+  void SetMaxDefRepLevels(int16_t max_def_level, int16_t max_rep_level) {
+    max_definition_level_ = max_def_level;
+    max_repetition_level_ = max_rep_level;
+  }
```
wesm: Pass this in the ColumnDescriptor

src/parquet/column/scanner.h
```diff
@@ -68,6 +68,10 @@ class Scanner {
     batch_size_ = batch_size;
   }
 
+  void SetReaderMaxDefRepLevels(int16_t max_def_level, int16_t max_rep_level) {
+    reader_->SetMaxDefRepLevels(max_def_level, max_rep_level);
+  }
```
wesm: Pass this in the ColumnDescriptor

src/parquet/column/reader.h
```diff
@@ -104,6 +108,10 @@ class ColumnReader {
   // The number of values from the current data page that have been decoded
   // into memory
   int num_decoded_values_;
+
+  int16_t max_definition_level_;
+
+  int16_t max_repetition_level_;
```
wesm: These are stored in the ColumnDescriptor now

src/parquet/column/column-reader-test.cc
majetideepak: Yes, the repetition levels appear before the definition levels.

src/parquet/column/reader.h
majetideepak: I just saw this after rebase. I am making these changes

src/parquet/column/column-reader-test.cc
wesm: good to know, thanks, will have to update some incorrect code comments =)

src/parquet/reader.cc
```diff
@@ -242,10 +242,10 @@ void ParquetFileReader::DebugPrint(std::ostream& stream, bool print_values) {
            << ")" << std::endl;
   }
 
-  for (int i = 0; i < num_row_groups(); ++i) {
-    stream << "--- Row Group " << i << " ---\n";
```
majetideepak: The loop iterator `i` was used again in a nested loop.

src/parquet/column/reader.h
```diff
@@ -85,13 +87,10 @@ class ColumnReader {
...
 
   // Not set if full schema for this field has no optional or repeated elements
-  std::unique_ptr<RleDecoder> definition_level_decoder_;
+  LevelDecoder definition_level_decoder_;
 
   // Not set for flat schemas.
-  std::unique_ptr<RleDecoder> repetition_level_decoder_;
-
-  // Temporarily storing this to assist with batch reading
-  int16_t max_definition_level_;
```
majetideepak: Do we need a copy here ?

src/parquet/column/reader.cc
```diff
@@ -109,24 +97,30 @@ bool TypedColumnReader<TYPE>::ReadNewPage() {
...
+      //Init repetition levels
+      if (max_repetition_level > 0) {
+        size_t rep_levels_bytes = repetition_level_decoder_.Init(
+            page->repetition_level_encoding(),
+            max_repetition_level, num_buffered_values_, buffer);
+        buffer += rep_levels_bytes;
+        data_size -= rep_levels_bytes;
+      }
+      //TODO figure a way to set max_definition_level_ to 0
+      //if the initial value is invalid
```
wesm: I don't think this should ever happen, except with invalid files?

src/parquet/column/reader.h
wesm: I don't think so since it's stored (and immutable) in the `descr_`

src/parquet/reader.cc
wesm: :+1:

src/parquet/column/reader.cc
majetideepak: Yes, it shouldn't occur unless we have invalid files or we have a bug in the tree schema resolution. I just left a note because you had code to handle that bit and I removed it.

src/parquet/column/column-reader-test.cc
```diff
@@ -157,6 +154,62 @@ TEST_F(TestPrimitiveReader, TestInt32FlatOptional) {
...
+  test::DataPageBuilder<Type::INT32> page_builder(&page1);
+
+  // Definition levels precede the values
+  page_builder.AppendRepLevels(rep_levels, 1, parquet::Encoding::RLE);
+  page_builder.AppendDefLevels(def_levels, 2, parquet::Encoding::RLE);
+  page_builder.AppendValues(values, parquet::Encoding::PLAIN);
+
+  pages_.push_back(page_builder.Finish());
+
+  NodePtr type = schema::Int32("a", Repetition::OPTIONAL);
```
majetideepak: changing this to REPEATED.

wesm:
Cool this LGTM +1. Thanks for sticking with the code reviews / tests! ping @nongli or @julienledem when you're done with the above tweaks and have a green build

majetideepak:
@wesm Will do. Thanks! for the tips about testing and in general. :100:

majetideepak:
@nongli  @julienledem this is good to go. Thanks!

majetideepak:
This PR will also implement PARQUET-462 and PARQUET-467

src/parquet/column/levels.h
```diff
@@ -0,0 +1,155 @@
...
+          break;
+        }
+        ++num_encoded;
+      }
+      bit_packed_encoder_->Flush();
+    }
+    return num_encoded;
+  }
+
+  int32_t len() {
```
nongli: How are you suppose to use this function? Assert this is not called with encoding_ != RLE?

src/parquet/column/levels.h
```diff
@@ -0,0 +1,155 @@
...
+  std::unique_ptr<RleEncoder> rle_encoder_;
+  std::unique_ptr<BitWriter> bit_packed_encoder_;
+};
+
+
+class LevelDecoder {
+ public:
+  LevelDecoder() {}
+
+  // Initialize the LevelDecoder.
```
nongli: Comment on return value.

src/parquet/column/levels.h
```diff
@@ -0,0 +1,155 @@
...
+#define PARQUET_COLUMN_LEVELS_H
+
+#include "parquet/exception.h"
+#include "parquet/thrift/parquet_types.h"
+#include "parquet/encodings/encodings.h"
+#include "parquet/util/rle-encoding.h"
+
+namespace parquet_cpp {
+
+class LevelEncoder {
```
nongli: Just so I understand this: this will eventually need a reset() or something like that right? The calling pattern is:

Init
Loop
  Encode()
  Encode()
  // do something
  Reset()

?

src/parquet/column/levels.h
majetideepak: This function is used when packing the rle data into a buffer (len() followed by rle-encoded-values). I will add the assertion. Thanks!

src/parquet/column/levels.h
majetideepak: Will do.

src/parquet/column/levels.h
majetideepak: We changed the implementation from Impala. The LevelDecoder does **not** inherit from the rle-decoder class here. The `Init`  above does the Impala `Reset`. We now just have to call `Init` for every new data page.

majetideepak: I spoke too soon. I will add a `Reset`. It is more efficient than creating a new object every time.

majetideepak: adding a Reset() for the encoder.

majetideepak: I am planning to postpone the encoder/decoder `Reset()`  for now. The advantage of using the decoder Reset is that we can use the same decoder instance for multiple data pages on a given column rather than `Init`ing every time. However, the performance gains might not be much. I will wait on the encoder Reset until the Parquet writer design.
@nongli any suggestions?

majetideepak:
Added PARQUET-515 to track the "Reset" feature.

wesm:
@majetideepak there's no need to squash your commits with each iteration if there are no merge / rebase conflicts -- it makes it harder to follow changes due to code reviews

majetideepak:
@wesm I realized it later. I will not squash them next time.

wesm:
I'll leave it to @nongli to sign off on this patch -- I don't have a feel for the `Reset` issue. We should do it sooner rather than later. In order to be properly tested we will need to construct test cases with multiple data pages. So I'm inclined to leave this unit testing and refactoring for PARQUET-515.

src/parquet/column/levels-test.cc
```diff
@@ -0,0 +1,129 @@
...
+    ASSERT_EQ(levels_count, num_levels);
+
+    decoder.Init(encoding, max_level, num_levels, bytes.data());
+    levels_count = decoder.Decode(num_levels, output_levels.data());
+
+    ASSERT_EQ(levels_count, num_levels);
+
+    for (int i = 0; i < num_levels; i++) {
+      EXPECT_EQ(input_levels[i], output_levels[i]);
+    }
```
wesm: Can you flip the order of the assertions in this function? noticed this issue in #44 per https://github.com/google/googletest/blob/master/googletest/docs/Primer.md

majetideepak: Will do.

majetideepak: In this case, isn't the order correct ? `input_levels[i]` is the expected value and `output_levels[i]` is the actual value.

wesm: The `EXPECT_EQ` is OK but the two assertions above it need their arguments flipped (`num_levels` is the expected value, right?)

majetideepak: Yes! They have been patched.

On Tue, Feb 9, 2016 at 6:30 PM, Wes McKinney <notifications@github.com>
wrote:

> In src/parquet/column/levels-test.cc
> <https://github.com/apache/parquet-cpp/pull/30#discussion_r52392726>:
>
> > +    } else {
> > +      encoder.Init(encoding, max_level, num_levels,
> > +          bytes.data(), bytes.size());
> > +      levels_count = encoder.Encode(num_levels, input_levels.data());
> > +    }
> > +
> > +    ASSERT_EQ(levels_count, num_levels);
> > +
> > +    decoder.Init(encoding, max_level, num_levels, bytes.data());
> > +    levels_count = decoder.Decode(num_levels, output_levels.data());
> > +
> > +    ASSERT_EQ(levels_count, num_levels);
> > +
> > +    for (int i = 0; i < num_levels; i++) {
> > +      EXPECT_EQ(input_levels[i], output_levels[i]);
> > +    }
>
> The EXPECT_EQ is OK but the two assertions above it need their arguments
> flipped (num_levels is the expected value, right?)
>
> —
> <https://github.com/apache/parquet-cpp/pull/30/files#r52392726>.
>



-- 
regards,
Deepak Majeti

src/parquet/column/levels.h
nongli: This sounds fine. Let's see how things evolve.

majetideepak:
Thank you!

On Wed, Feb 10, 2016 at 11:43 PM, asfgit <notifications@github.com> wrote:

> Closed #30 <https://github.com/apache/parquet-cpp/pull/30> via 13843d3
> <https://github.com/apache/parquet-cpp/commit/13843d3d4d2fe8a2c70b6ab8a6f40f55ebe9680b>
> .
>
> —
> <https://github.com/apache/parquet-cpp/pull/30#event-546733437>.
>



-- 
regards,
Deepak Majeti

PR #31, opened 2016-01-29
PARQUET-438: Update RLE encoding tools and add unit tests from Impala

wesm:
This also changes the compiler options to use only SSE3. We can explore benchmarking to optionally enable SSE4 on supported systems for performance gains in the future. 

@nongli would it be OK if I remove the unused delta encodings (they depend on code that is not in Impala anymore, or never way)? Are these encodings used extensively, or primarily PLAIN / DICTIONARY encoding?

### DIFF
--- a/src/parquet/util/bit-stream-utils.inline.h
+++ b/src/parquet/util/bit-stream-utils.inline.h
@@ -152,6 +152,21 @@ inline bool BitReader::GetVlqInt(int32_t* v) {
   return true;
 }
 
+// TODO(nongli): review/test these implementations given divergence in Impala
+// functions
+
+// inline bool BitWriter::PutZigZagVlqInt(int32_t v) {
+//   uint32_t u = (v << 1) ^ (v >> 31);
+//   return PutVlqInt(u);
+// }
+
+// inline bool BitReader::GetZigZagVlqInt(int64_t* v) {
+//   uint64_t u;
+//   if (!GetVlqInt(&u)) return false;
+//   *reinterpret_cast<uint64_t*>(v) = (u >> 1) ^ -(u & 1);
+//   return true;
+// }
+
 } // namespace parquet_cpp
 
 #endif // PARQUET_UTIL_BIT_STREAM_UTILS_INLINE_H
--- a/src/parquet/util/bit-stream-utils.inline.h
+++ b/src/parquet/util/bit-stream-utils.inline.h
@@ -15,6 +15,8 @@
 // specific language governing permissions and limitations
 // under the License.
 
+// From Apache Impala as of 2016-01-29
+
 #ifndef PARQUET_UTIL_BIT_STREAM_UTILS_INLINE_H
 #define PARQUET_UTIL_BIT_STREAM_UTILS_INLINE_H
 
@@ -73,7 +75,7 @@ inline bool BitWriter::PutAligned(T val, int num_bytes) {
   return true;
 }
 
-inline bool BitWriter::PutVlqInt(uint32_t v) {
+inline bool BitWriter::PutVlqInt(int32_t v) {
   bool result = true;
   while ((v & 0xFFFFFF80) != 0L) {
     result &= PutAligned<uint8_t>((v & 0x7F) | 0x80, 1);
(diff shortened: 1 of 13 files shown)


majetideepak:
We need the RLE decoder to handle BIT_PACKED encoding as well. Can this code decode BIT_PACKED values ?

nongli:
@wesm Which encoding are you talking about? Some of these are the 2.0 encodings which should not removed. Let's just add the utilities we need from Impala in here.

wesm:
@majetideepak no, and it does not need to. You can just use a `BitReader` in that case, see https://github.com/cloudera/Impala/blob/cdh5-trunk/be/src/exec/hdfs-parquet-scanner.cc#L1247

To handle `BIT_PACKED` level encoding, like Impala we should insert a `LevelDecoder` abstraction which dispatches either to `RLE` or `BIT_PACKED` encoding as appropriate.

wesm:
@nongli all of the `delta*` headers

majetideepak:
`LevelDecoder` abstraction works as well.
I will combine PARQUET-169 and PARQUET-462 similar to the Impala's `LevelDecoder`

wesm:
@majetideepak can you base that change on PARQUET-435 / #26 ? thanks

majetideepak:
Yes! I will. Makes more sense.

nongli:
@wesm we should not remove those. Those are the 2.0 encoding implementations. Impala doesn't support 2.0 yet.

wesm:
Got it, thank you. I commented out the includes -- we can go back and reconcile the implementations when someone is ready to write unit tests. Perhaps by the time Impala will have implemented the encodings.

majetideepak:
+1 LGTM

majetideepak:
I plan to rebase PARQUET-169 and PARQUET-462 on #31

wesm:
I have an incoming patch that refactors `ReadNewPage` -- but I don't think the conflict will be too bad

wesm:
(per PARQUET-485)

src/parquet/util/rle-encoding.h
```diff
@@ -93,170 +94,171 @@ class RleDecoder {
...
+    repeat_count_ = 0;
+    literal_count_ = 0;
+  }
+
+  /// Gets the next value.  Returns false if there are no more.
   template<typename T>
   bool Get(T* val);
 
- private:
+ protected:
```
majetideepak: Should we keep this private ?

wesm: Since this was a copy paste job I would rather not make any unnecessary changes -- I'm also exploring the possibility of making parquet-cpp a dependency of Impala and this would cause code breakage.

wesm:
This has been rebased on PARQUET-496 and the build should be passing again

src/parquet/util/bit-stream-utils.inline.h
```diff
@@ -83,13 +85,9 @@ inline bool BitWriter::PutVlqInt(uint32_t v) {
   return result;
 }
 
-inline bool BitWriter::PutZigZagVlqInt(int32_t v) {
```
nongli: Can you add these (zigzag) back? I think this makes the new decoders work.

You can leave a TODO for me to add tests.

wesm: Sure.

wesm:
@nongli the signatures for `GetVlqInt` and `PutVlqInt` are different in Impala from what they were before this patch. I am not confident restoring the ZigZag functions and twiddling integer types without tests; let me know what you want to do

nongli:
Rather than just removing it, just comment it out and i'll take a look.

wesm:
Roger. Thanks

wesm:
All set.

src/parquet/util/rle-encoding.h
julienledem: +1 on impala using parquet-cpp!

src/parquet/util/bit-util-test.cc
```diff
@@ -0,0 +1,164 @@
...
+  EXPECT_EQ(BitUtil::RoundDownNumi32(33), 1);
+
+  EXPECT_EQ(BitUtil::RoundUpNumi64(63), 1);
+  EXPECT_EQ(BitUtil::RoundUpNumi64(64), 1);
+  EXPECT_EQ(BitUtil::RoundUpNumi64(65), 2);
+  EXPECT_EQ(BitUtil::RoundDownNumi64(63), 0);
+  EXPECT_EQ(BitUtil::RoundDownNumi64(64), 1);
+  EXPECT_EQ(BitUtil::RoundDownNumi64(65), 1);
+}
+
```
julienledem: nice

julienledem:
LGTM. any more comments?

majetideepak:
+1 LGTM

On Mon, Feb 1, 2016 at 7:18 PM, Julien Le Dem <notifications@github.com>
wrote:

> LGTM. any more comments?
>
> —
> <https://github.com/apache/parquet-cpp/pull/31#issuecomment-178270207>.
>



-- 
regards,
Deepak Majeti

julienledem:
I'll merge then :)

PR #32, opened 2016-01-31
PARQUET-485: Decouple page deserialization from column reader to facilitate unit testing

wesm:
Several things in this patch

* Adds PageReader abstraction, and a SerializedPageReader implementation
  according to the Parquet file format
* Adds a MockPageReader and a couple unit tests demonstrating end-to-end test
  without creating a Parquet file
* Adds a DataPageBuilder test fixture tool, may become part of the main write
  path later
* Adds PlainEncoder implementation for a few primitive types
* Fixes a few ColumnReader bugs exposed by the unit tests

### DIFF
--- a/src/parquet/column/page.h
+++ b/src/parquet/column/page.h
@@ -122,8 +122,8 @@ class PageReader {
  public:
   virtual ~PageReader() {}
 
-  // @returns: false on EOS, true if a new page has been returned
-  virtual bool NextPage(std::shared_ptr<Page>& out) = 0;
+  // @returns: pointer to null on EOS, std::shared_ptr<Page> otherwise
+  virtual std::shared_ptr<Page> NextPage() = 0;
 };
 
 } // namespace parquet_cpp
--- a/src/parquet/column/page.h
+++ b/src/parquet/column/page.h
@@ -1,0 +1,127 @@
+// Licensed to the Apache Software Foundation (ASF) under one
+// or more contributor license agreements.  See the NOTICE file
+// distributed with this work for additional information
+// regarding copyright ownership.  The ASF licenses this file
+// to you under the Apache License, Version 2.0 (the
+// "License"); you may not use this file except in compliance
+// with the License.  You may obtain a copy of the License at
+//
+//   http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing,
+// software distributed under the License is distributed on an
+// "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+// KIND, either express or implied.  See the License for the
+// specific language governing permissions and limitations
+// under the License.
+
+// This module defines an abstract interface for iterating through pages in a
+// Parquet column chunk within a row group. It could be extended in the future
+// to iterate through all data pages in all chunks in a file.
+
+#ifndef PARQUET_COLUMN_PAGE_H
+#define PARQUET_COLUMN_PAGE_H
+
+#include "parquet/thrift/parquet_types.h"
+
+namespace parquet_cpp {
+
+// Note: Copying the specific page header Thrift metadata to the Page
+// object (instead of using a pointer) presently so that data pages can be
+// decompressed and processed in parallel
+//
+// TODO: Parallel processing is not yet safe because of memory-ownership
+// semantics (the PageReader may or may not own the memory referenced by a
+// page)
+class Page {
+  // TODO(wesm): In the future Parquet implementations may store the crc code
+  // in parquet::PageHeader. parquet-mr currently does not, so we also skip it
+  // here, both on the read and write path
+ public:
+  Page(const uint8_t* buffer, size_t buffer_size, parquet::PageType::type type) :
+      buffer_(buffer),
+      buffer_size_(buffer_size),
+      type_(type) {}
+
+  parquet::PageType::type type() const {
+    return type_;
+  }
+
+  // @returns: a pointer to the page's data
+  const uint8_t* data() const {
+    return buffer_;
+  }
+
+  // @returns: the total size in bytes of the page's data buffer
+  size_t size() const {
+    return buffer_size_;
+  }
+
+ private:
+  const uint8_t* buffer_;
+  size_t buffer_size_;
+
+  parquet::PageType::type type_;
+};
+
+
+class DataPage : public Page {
+ public:
+  DataPage(const uint8_t* buffer, size_t buffer_size,
+      const parquet::DataPageHeader& header) :
+      Page(buffer, buffer_size, parquet::PageType::DATA_PAGE),
+      header_(header) {}
+
+  size_t num_values() const {
+    return header_.num_values;
+  }
+
+  parquet::Encoding::type encoding() const {
+    return header_.encoding;
+  }
+
+ private:
+  parquet::DataPageHeader header_;
+};
+
+
+class DataPageV2 : public Page {
+ public:
+  DataPageV2(const uint8_t* buffer, size_t buffer_size,
+      const parquet::DataPageHeaderV2& header) :
+      Page(buffer, buffer_size, parquet::PageType::DATA_PAGE_V2),
+      header_(header) {}
+
+ private:
+  parquet::DataPageHeaderV2 header_;
+};
+
+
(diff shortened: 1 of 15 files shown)


wesm:
@majetideepak I'm going to add a couple test examples using the `MockPageReader` in this patch. If you could base your changes for PARQUET-169 on this it will also enable you to write unit tests with data that you declare with the test case. 

Testing will be slightly tricky as we do not yet have schema support for repeated fields, so you may wish to add a temporary member (the `max_repetition_level`) that you can set in the test case. I'm going to implement schema resolution this week so that the data type as well as the maximum repetition and definition levels are passed in to the column reader ctors. You can also wait for me to complete this work as it will reduce code / unit test churn.

src/parquet/column/reader-test.cc
```diff
@@ -0,0 +1,84 @@
...
+};
+
+template <typename T>
+static void slice(const vector<T>& values, size_t start, size_t end,
+    vector<T>& out) {
+  if (end < start) {
+    return;
+  }
+
+  out.resize(end - start);
```
yixinglu: `out.reserve(end - start);` ?

src/parquet/column/page.h
```diff
@@ -0,0 +1,125 @@
...
+
+// Note: Copying the specific page header Thrift metadata to the Page
+// object (instead of using a pointer) presently so that data pages can be
+// decompressed and processed in parallel
+//
+// TODO: Parallel processing is not yet safe because of memory-ownership
+// semantics (the PageReader may or may not own the memory referenced by a
+// page)
+class Page {
+  // TODO(wesm): crc verification from parquet::PageHeader
```
julienledem: FYI: parquet-mr never implemented crc verification. We had it in the spec but it is still a todo there too.

src/parquet/column/reader-test.cc
wesm: Reserve doesn't change the array size

src/parquet/column/page.h
julienledem: or generation

wesm: Thanks, good to know!

julienledem:
Looks good to me.
Anybody else with comments?

wesm:
I'm going to do a little more on test cases. Will advise when done tomorrow

src/parquet/column/reader-test.cc
wesm: Ooh there's actually a bug on line 65, will fix

wesm:
OK, I've got a "hello world" test working on a REQUIRED field; I will add a test for an OPTIONAL field (with definition levels), then let's merge this. Hard to do a lot more without richer schema information. Code coverage will help steer further ColumnReader testing efforts

wesm:
OK, I added another unit test that verifies a flat optional field. It would be great if we can get this merged today as it simplifies unit testing `ColumnReader`. I fixed multiple correctness bugs (that would cause segfaults) as part of this patch.

We can continue to develop tooling to simplify setting up test fixtures and reducing the amount of boilerplate to a minimum.

wesm:
Just rebased this on master after #33 merged.

majetideepak:
I added test cases in #30 for the rle/bit-packed values encoding and decoding via the LevelEncoder and LevelDecoder API.

julienledem:
This looks good to me.
I'll merge it later today unless someone else has comments.

wesm:
I'd like @nongli to have a quick glance -- let me know.

src/parquet/encodings/encodings.h
```diff
@@ -67,6 +67,38 @@ class Decoder {
...
+class Encoder {
+ public:
+  typedef typename type_traits<TYPE>::value_type T;
+
+  virtual ~Encoder() {}
+
+  // Subclasses should override the ones they support
+  //
+  // @returns: the number of bytes written to dst
+  virtual size_t Encode(const T* src, int num_values, uint8_t* dst) {
```
nongli: Are you sure you want dst to be a raw array instead of something like output stream? We can revisit later when the encoding side is built out more but this might be hard to deal with. Haven't put too much thought but maybe something blocky:
i.e.
byte* buffer = output->Reserve(big block for a batch).

src/parquet/column/page.h
```diff
@@ -0,0 +1,127 @@
...
+  parquet::DictionaryPageHeader header_;
+};
+
+// Abstract page iterator interface. This way, we can feed column pages to the
+// ColumnReader through whatever mechanism we choose
+class PageReader {
+ public:
+  virtual ~PageReader() {}
+
+  // @returns: false on EOS, true if a new page has been returned
```
nongli: Should this just return shared_ptr<Page> and null on eos?

src/parquet/column/serialized-page.h
```diff
@@ -0,0 +1,61 @@
...
+
+#include "parquet/column/page.h"
+#include "parquet/compression/codec.h"
+#include "parquet/util/input_stream.h"
+#include "parquet/thrift/parquet_types.h"
+
+namespace parquet_cpp {
+
+// This subclass delimits pages appearing in a serialized stream, each
+// proceeded by a serialized Thrift parquet::PageHeader indicating the type of
```
nongli: preceded

src/parquet/column/page.h
```diff
@@ -0,0 +1,127 @@
...
+// to iterate through all data pages in all chunks in a file.
+
+#ifndef PARQUET_COLUMN_PAGE_H
+#define PARQUET_COLUMN_PAGE_H
+
+#include "parquet/thrift/parquet_types.h"
+
+namespace parquet_cpp {
+
+// Note: Copying the specific page header Thrift metadata to the Page
```
nongli: I'm not sure I understand this comment.

src/parquet/encodings/encodings.h
wesm: No, I'm not sure, I was also thinking about this because these classes have no ability to allocate memory. Let's create a JIRA to do that before we get in too deep with this assumption; I don't think this JIRA is the right place for it though...

src/parquet/column/page.h
wesm: Well, right now you can do

`if (!reader_->NextPage(&currentpage_))` which is nice. Are you suggesting returning `std::shared_ptr<PageReader>(nullptr)`?

src/parquet/column/page.h
wesm: Well, the issue here is that the deserialized page header has a limited lifetime in the current implementation. If you wanted to pipeline page processing, you need some way to preserve the page metadata for all of the "live" (unprocessed) pages that have been buffered in memory.

wesm: I.e., I'm referring to the `current_page_header_` member variable (which gets overwritten each time a page header is deserialized)

src/parquet/column/serialized-page.h
wesm: thanks will fix

src/parquet/column/page.h
wesm: Another option is to have each of these classes have a pointer to metadata, and if you want to have some pipelining abstraction then you can create a container that owns the deserialized metadata to eliminate unnecessary copies.

src/parquet/encodings/encodings.h
wesm: I created https://issues.apache.org/jira/browse/PARQUET-501 to track.

src/parquet/column/page.h
wesm: I thought some more about this and I'm fine with

```c++
current_page_ = reader->NextPage();
if (!current_page) {
  // eos
  ...
}
```

I wasn't sure whether returning a shared_ptr to null was standard practice but it seems fine to me.

wesm: Let me know if that works for you

wesm: Yeah, this is better, just pushed the change.

wesm:
Just rebased and restored a missing function that got lost in the RLE sync up. This should be good to go when the build is green.

wesm:
Travis CI is super unhappy today, seems like a trend. I sent them a support email about it

wesm:
:100: thank you! rebasing #34 now

PR #33, opened 2016-02-01
PARQUET-496: Fix cpplint configuration to catch more style errors

wesm:
We were using `--verbose=4` which catches only egregious style errors. `--verbose=2` with a few warning types suppressed is much more reasonable. This patch updates the cmake configuration and fixes the existing style errors to get the build passing.


nongli:
Thanks for doing this. Merging.

wesm:
thank you!

PR #34, opened 2016-02-01
PARQUET-454: Fix inconsistencies with boolean PLAIN encoding

wesm:
Requires PARQUET-485 (#32)

The boolean Encoding::PLAIN code path was using RleDecoder, inconsistent with
other implementations of Parquet. This patch adds an implementation of plain
encoding and uses BitReader instead of RleDecoder to decode plain-encoded
boolean data. Unit tests to verify.

Also closes PR #12. Thanks to @edani for reporting.

### DIFF
--- a/src/parquet/encodings/plain-encoding-test.cc
+++ b/src/parquet/encodings/plain-encoding-test.cc
@@ -1,0 +1,78 @@
+// Licensed to the Apache Software Foundation (ASF) under one
+// or more contributor license agreements.  See the NOTICE file
+// distributed with this work for additional information
+// regarding copyright ownership.  The ASF licenses this file
+// to you under the Apache License, Version 2.0 (the
+// "License"); you may not use this file except in compliance
+// with the License.  You may obtain a copy of the License at
+//
+//   http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing,
+// software distributed under the License is distributed on an
+// "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+// KIND, either express or implied.  See the License for the
+// specific language governing permissions and limitations
+// under the License.
+
+#include <cstdint>
+#include <random>
+#include <string>
+#include <vector>
+
+#include <gtest/gtest.h>
+#include "parquet/util/test-common.h"
+
+#include "parquet/encodings/encodings.h"
+
+using std::string;
+using std::vector;
+
+namespace parquet_cpp {
+
+namespace test {
+
+static vector<bool> flip_coins(size_t n, double p) {
+  std::random_device rd;
+  std::mt19937 gen(rd());
+
+  std::bernoulli_distribution d(p);
+
+  vector<bool> draws;
+  for (size_t i = 0; i < n; ++i) {
+    draws.push_back(d(gen));
+  }
+  return draws;
+}
+
+TEST(BooleanTest, TestEncodeDecode) {
+  // PARQUET-454
+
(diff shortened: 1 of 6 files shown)


src/parquet/column/serialized-page.cc
```diff
@@ -0,0 +1,104 @@
...
+    case parquet::CompressionCodec::SNAPPY:
+      decompressor_.reset(new SnappyCodec());
+      break;
+    default:
+      ParquetException::NYI("Reading compressed data");
+  }
+}
+
+// TODO(wesm): this may differ from file to file
+static constexpr int DATA_PAGE_SIZE = 64 * 1024;
```
julienledem: is it a default page size threshold?

src/parquet/column/serialized-page.cc
```diff
@@ -0,0 +1,104 @@
...
+      return true;
+    } else if (current_page_header_.type == PageType::DATA_PAGE) {
+      out.reset(new DataPage(buffer, uncompressed_len,
+              current_page_header_.data_page_header));
+      return true;
+    } else if (current_page_header_.type == PageType::DATA_PAGE_V2) {
+      ParquetException::NYI("data page v2");
+    } else {
+      // We don't know what this page type is. We're allowed to skip non-data
+      // pages.
```
julienledem: // this will enable adding index pages interleaved with data pages in the future

src/parquet/encodings/plain-encoding-test.cc
```diff
@@ -0,0 +1,78 @@
...
+  return draws;
+}
+
+TEST(BooleanTest, TestEncodeDecode) {
+  // PARQUET-454
+
+  size_t nvalues = 100;
+  size_t nbytes = BitUtil::RoundUp(nvalues, 8) / 8;
+
+  vector<bool> draws = flip_coins(nvalues, 0.5);
```
julienledem: randomized tests should print out a seed when they fail so that they can be reproduced.
(if that's already the case then ignore this comment)

src/parquet/util/bit-util.h
```diff
@@ -33,6 +33,14 @@ namespace parquet_cpp {
 // TODO: is this in boost or something else like that?
```
julienledem: is this still a valid comment?

julienledem:
This looks good to me. Needs rebase.

wesm:
Yep, will rebase after #32 is merged

src/parquet/encodings/plain-encoding-test.cc
wesm: I'm changing this to use a set seed for now.

src/parquet/util/bit-util.h
wesm: This comment is straight from the Impala codebase. We should review the bit utilities at some point and see if we can use 3rdparty code (if it's better-tested / more portable) to replace some of these

src/parquet/column/serialized-page.cc
wesm: this is the `DEFAULT_DATA_PAGE_SIZE` from Impala. This needs to be configurable and we should be flexible to automatically expand the page size when the headers get large. I'll open a JIRA now

wesm: https://issues.apache.org/jira/browse/PARQUET-505

wesm:
This is rebased -- have a look and let me know anything else we should do here. We'll need to significantly expand the plain encoder tests, so we should hit these code paths with more tests to exhaust the possibilities.

wesm:
thank you!

PR #35, opened 2016-02-02
PARQUET-503: Reenable parquet 2.0 encoding implementations.

nongli:


src/parquet/util/bit-stream-utils.inline.h
```diff
@@ -75,7 +75,7 @@ inline bool BitWriter::PutAligned(T val, int num_bytes) {
   return true;
 }
 
-inline bool BitWriter::PutVlqInt(int32_t v) {
+inline bool BitWriter::PutVlqInt(uint32_t v) {
```
wesm: does this have any correctness implications? I assume that this function only ever was getting used with non-negative integers

wesm:
lgtm

src/parquet/util/bit-stream-utils.inline.h
nongli: I think the impala implementation is wrong and doesn't handle negative numbers. We want to do a logical shift, not arithmetic shift. The version with int32_t will spin forever for negative numbers.

julienledem: FYI: the java side deals with signed integers

julienledem:
+1

src/parquet/util/bit-stream-utils.inline.h
nongli: I think this just does the right thing. The implicit cast from signed to unsigned is correct.

PR #36, opened 2016-02-02
PARQUET-481: Refactored and expanded reader-test and scanner-test.

asandryh:
Changes:
* Moved Scanner tests from reader-test.cc to scanner-test.cc
* Expanded both tests to run on multiple files.

This also resolves PARQUET-475 and PARQUET-502.

### DIFF
--- a/src/parquet/column/scanner-test.cc
+++ b/src/parquet/column/scanner-test.cc
@@ -22,7 +22,7 @@
 
 #include <gtest/gtest.h>
 
-#include "parquet/column/reader.h"
+#include "parquet/column/scanner.h"
 #include "parquet/util/test-common.h"
 
 using std::string;
@@ -31,50 +31,38 @@ namespace parquet_cpp {
 
 namespace test {
 
-  TEST_P(TestReader, Metadata) {
-    reader_.ParseMetaData();
-  }
-
-  TEST_P(TestReader, DebugPrintWorks) {
-    std::stringstream ss;
-    reader_.DebugPrint(ss);
-    std::string result = ss.str();
-    ASSERT_TRUE(result.length() > 0);
-  }
-
-  TEST_P(TestReader, BatchReadInt32) {
+  TEST_P(TestReader, FlatScannerInt32) {
     ssize_t c = findColumn(Type::INT32);
     if (c >= 0) {
       RowGroupReader* group = reader_.RowGroup(0);
-      const parquet::ColumnMetaData* metadata = group->column_metadata(c);
-      int64_t rows = metadata->num_values;
+      int64_t rows = group->column_metadata(c)->num_values;
 
-      std::shared_ptr<Int32Reader> col =
-          std::dynamic_pointer_cast<Int32Reader>(group->Column(c));
+      std::shared_ptr<Int32Scanner> scanner =
+          std::make_shared<Int32Scanner>(group->Column(c));
 
-      int16_t def_levels[rows];
-      int16_t rep_levels[rows];
-      int32_t values[rows];
-
-      // Read all but last rows
-      ASSERT_TRUE(col->HasNext());
-      size_t values_read;
-      size_t levels_read =
-          col->ReadBatch(rows-1, def_levels, rep_levels, values, &values_read);
-      ASSERT_EQ(rows-1, levels_read);
-      size_t total_values_read = values_read;
-
-      // Now read past the end of the file
-      ASSERT_TRUE(col->HasNext());
-      levels_read = col->ReadBatch(2, def_levels, rep_levels, values, &values_read);
-      ASSERT_EQ(1, levels_read);
-      total_values_read += values_read;
-
-      ASSERT_FALSE(col->HasNext());
-
-      ASSERT_EQ(total_values_read, rows - metadata->statistics.null_count);
+      int32_t val;
+      bool is_null;
+      for (; rows > 0; rows--) {
+        scanner->NextValue(&val, &is_null);
+      }
+      ASSERT_FALSE(scanner->HasNext());
+      ASSERT_FALSE(scanner->NextValue(&val, &is_null));
     }
   }
+
+  TEST_P(TestReader, ScannerBatchSize) {
+    ssize_t c = findColumn(Type::INT32);
+    if (c >= 0) {
+      RowGroupReader* group = reader_.RowGroup(0);
+      std::shared_ptr<Int32Scanner> scanner =
+          std::make_shared<Int32Scanner>(group->Column(c));
+
+      ASSERT_EQ(128, scanner->batch_size());
+      scanner->SetBatchSize(1024);
+      ASSERT_EQ(1024, scanner->batch_size());
+    }
+  }
+
 
   INSTANTIATE_TEST_CASE_P(ReaderTest, TestReader,
       testing::Values(
--- a/src/parquet/reader.cc
+++ b/src/parquet/reader.cc
@@ -317,11 +317,6 @@ void ParquetFileReader::DebugPrint(std::ostream& stream, bool print_values) {
     do {
       hasRow = false;
       for (int c = 0; c < nColumns; ++c) {
-        if (scanners[c] == NULL) {
-          snprintf(buffer, bufsize, "%-" COL_WIDTH"s", " ");
-          stream << buffer;
-          continue;
-        }
         if (scanners[c]->HasNext()) {
           hasRow = true;
           scanners[c]->PrintNext(stream, 17);
--- a/src/parquet/util/test-common.h
+++ b/src/parquet/util/test-common.h
@@ -20,7 +20,10 @@
 
 #include <iostream>
 #include <random>
+#include <string>
 #include <vector>
+
+#include "parquet/reader.h"
 
 using std::vector;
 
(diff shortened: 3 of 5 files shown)


wesm:
I haven't reviewed this yet carefully, but I'm -1 on this testing approach (test cases that rely on pre-generated data files and output) at this stage of the library's development. Can you explain why we should be doing this while the library is so immature and poorly unit tested? 

I would be supportive of setting up an external integration test repository where we can put things like this to assist with smoke-testing the library while we're building up our unit test capabilities.

asandryh:
Both reader and scanner need their corresponding tests. While it's true that the library does not yet have all features implemented, it is not a reason to not have tests for the components that are implemented. The library has many interacting components, and it is important to confirm that we maintain their full integration and can actually read parquet files correctly. As we add support for more features (encodings, compression, etc.), they can be reflected in these tests, too, in addition to the corresponding unit tests for those features.
Obviously, once we can write parquet, we'll get rid of the pre-generated files.

wesm:
I believe we should be doing automated integration testing with parquet-mr when the time comes, but I don't believe ad hoc integration tests have a place in parquet-cpp's git history -- by all means we should have an external smoke test repo. So I see a few options:

- Assist with implementing both read and write support and unit test tooling to verify each functional component in isolation sufficient to be satisfied with its functionality
- Set up an external smoke test repo to place a growing collection of files we want to be sure we can scan correctly
- Work on automated integration testing with parquet-mr

wesm:
@nongli @julienledem can you chime in with your thoughts here. thanks

asandryh:
Perhaps, I am missing something here, but (aside from a philosophical discussion on the value of end-to-end tests) what is wrong with adding tests that ensures that we can read/scan a file correctly? They are both unit tests for ParquetFileReader and Scanner, as well as integration tests for the underlying components. Without integration test, when someone modifies a component that changes its interaction with another component in a way that results in an incorrectly read file, this change will slip through because unit tests do not check for this. We can catch it later, when we integrate with parquet-mr or can write parquet, but why postpone catching and fixing an error if we can do now?

wesm:
Technically speaking they are not unit tests, they are UI tests on static data. I'm opposed to approaching new software in this manner while the unit tests remain inadequate. 

I'm supportive of having such tests, but not stored in this repo. In addition to bloating the git repo size (not as much of an issue with SVN / CVS), they tend to be brittle, especially with new software that is developing quickly. In my experience, in newer software they also create a false sense of security. 

Integration testing will help us find and fix flaws in the implementation, but it should be automated and not based on static data. When integration testing finds a bug, making the integration test pass is generally not acceptable by itself except in exceptional cases -- you need to write a unit test that isolates and reproduces the root cause of the integration test failure to resolve it. 

Like I said, happy to have an external repo for this and help keep its tests passing as long as maintaining them doesn't become onerous for me.

julienledem:
Hi,
I think we all agree that every component need their corresponding tests. 

In general I agree with the points @wesm makes about checking in pre-generated files:
I think we should not check in binary files that were generated somewhere else in the unit tests. This makes it hard to change anything to the test or understand what it is doing.
These are brittle because they are too coarse. Which also makes them harder to debug when they fail.
This, in the end impedes productivity. When more focused unit tests pinpoint exactly the part that's broken and tend to be more stable since they fail only when that part fails.

I also understand that @asandryh's priority is getting a Parquet reader working. So we need to enable that too. But i'd think we can unit test the various bits of a reader without depending on the writer.

It sounds like we need to decide where to put such integration tests. I can ask a new git repo for this but in the meantime that should not be a blocker either. Anyone can create a github repo and get started and others can submit PRs to it too.
Also, because of the way git works checking in and updating binary files grows the repo and becomes a burden over time. Which is another argument for a repo dedicated to this.
Does the separate repo sound like the way to go?

wesm:
+1. This has the side benefit of validating (on a continuous basis) that libparquet can be used as a 3rd-party library.

asandryh:
Let me look into setting up a separate repo with either pre-generated files or parquet-mr integration. I will also check if and how this can be added to Travis CI.

wesm:
Cool -- we should not hook this into parquet-cpp's CI process, though, as it will slow down builds -- when an integration test fails we can open a new JIRA to reproduce the bug and add a unit test to fix.

asandryh:
I meant CI for the testing repo. Basically, I'd like every commit to parquet-cpp to trigger the reading tests. I don't know if this can be done automatically, though.

wesm:
I see, that would be nice. I have nightly Jenkins builds for other projects I work on, hopefully Travis CI at least supports that use case.

asandryh:
Here's what I've found: At this time, Travis supports neither triggers for third-party repos nor nightly builds. We could add a custom script to parquet-cpp/.travis.yml -- but I don't think it's worth it in our case, since we expect to have write-read tests fairly soon. Let's revisit this discussion in a couple of weeks.

wesm:
In the meantime, you could run a nightly cron job that pushes a commit to the integration test repo to trigger a Travis CI build. Not the most robust solution but it should suffice for a few months at least. 

We can put a badge in the README so it's easy to check the build status similar to:

https://github.com/pydata/pandas/blob/master/README.md

asandryh:
Thanks, I'll look into this!
I will shortly update this PR to not include json file.

wesm:
Please remove the new .parquet files too. Can you add some unit tests for the JSON code (check out the new test fixture tools used in `parquet/column/column-reader-test`)? JSON is a can of worms because reconstructing nested JSON is somewhat onerous (and unclear how useful, versus yielding in-memory data structures).

wesm:
Another question: do you intend for `JsonPrint` to be used for anything other than debugging / integration test purposes? If not, it may be good to move both JsonPrint and DebugPrint to a separate `parquet::debug` namespace and compilation unit. I'd rather move them out of the `ParquetFileReader` public API

asandryh:
I am going to remove `JsonPrint`, since it's purpose was only to compare with json files. As for `DebugPrint`: it badly needs a re-write, and yes, it does not belong to `ParquetFileReader`. I will not do it in this PR, but rather file a separate issue to add `parquet_cpp::test::ParquetFilePrinter` class that prints a file contents out.

wesm:
Can you let me know before you embark on that `DebugPrint` rewrite? Things are going to be a bit chaotic while I'm working on PARQUET-442 (should be complete by the end of the week if not sooner).

asandryh:
Sure, it's not urgent. I'll cc you on the jira, and will work on it once I get a green light from you.

asandryh:
@wesm @julienledem Please review the updated PR and comment.

wesm:
I still have some mixed feelings about this patch. The first chance I get (within the next 4-7 days), I plan to change all these tests to use the `MockPageReader` test fixture and remove the test data files from the repository entirely. This is part of why I've been working tirelessly to make these components easier to unit test. I would rather do that directly rather than refactor only to refactor again. 

It is nice to have an integration test helper class, I will review now. 

I will let either @nongli and @julienledem pass final judgment with fair warning that either way everything related to the `data/` directory has a limited half life in this repository.

src/parquet/column/scanner-test.cc
```diff
@@ -0,0 +1,75 @@
...
+#include "parquet/util/test-common.h"
+
+using std::string;
+
+namespace parquet_cpp {
+
+namespace test {
+
+  TEST_P(TestReader, FlatScannerInt32) {
+    ssize_t c = findColumn(Type::INT32);
```
wesm: `ssize_t` is not portable

src/parquet/column/scanner-test.cc
```diff
@@ -0,0 +1,75 @@
...
+      for (; rows > 0; rows--) {
+        scanner->NextValue(&val, &is_null);
+      }
+      ASSERT_FALSE(scanner->HasNext());
+      ASSERT_FALSE(scanner->NextValue(&val, &is_null));
+    }
+  }
+
+  TEST_P(TestReader, ScannerBatchSize) {
+    ssize_t c = findColumn(Type::INT32);
```
wesm: here also

src/parquet/util/test-common.h
```diff
@@ -72,6 +75,49 @@ static inline vector<bool> flip_coins(size_t n, double p) {
...
+  std::string filename_;
+
+  explicit TestFileInfo(const std::string& filename): filename_(filename) {}
+};
+
+class TestReader: public testing::TestWithParam<TestFileInfo> {
+ public:
+  virtual ~TestReader();
+
+  std::string getFilename() {
```
wesm: `GetFilename`

src/parquet/util/test-common.h
```diff
@@ -72,6 +75,49 @@ static inline vector<bool> flip_coins(size_t n, double p) {
   return draws;
 }
 
+const char* data_dir = std::getenv("PARQUET_TEST_DATA");
+
+class TestFileInfo {
```
wesm: maybe struct here (for POD structs you can omit the `_` on members variables https://google.github.io/styleguide/cppguide.html#Variable_Names)?

src/parquet/util/test-common.h
```diff
@@ -72,6 +75,49 @@ static inline vector<bool> flip_coins(size_t n, double p) {
...
+    return filename.str();
+  }
+
+  TestReader() {
+    file_.Open(getFilename());
+    reader_.Open(&file_);
+  }
+
+  // Find a column of a specific type
+  ssize_t findColumn(parquet::Type::type t) {
```
wesm: Is this generally useful? I will comment above where it is used.

wesm: also, `FindColumn`

src/parquet/reader.cc
```diff
@@ -317,11 +317,6 @@ void ParquetFileReader::DebugPrint(std::ostream& stream, bool print_values) {
     do {
       hasRow = false;
       for (int c = 0; c < nColumns; ++c) {
-        if (scanners[c] == NULL) {
```
wesm: why was this removed?

src/parquet/reader-test.cc
```diff
@@ -22,106 +22,66 @@
...
-
-  int32_t val;
-  bool is_null;
-  for (size_t i = 0; i < 8; ++i) {
-    ASSERT_TRUE(scanner->HasNext());
-    ASSERT_TRUE(scanner->NextValue(&val, &is_null));
-    ASSERT_FALSE(is_null);
+  TEST_P(TestReader, BatchReadInt32) {
+    ssize_t c = findColumn(Type::INT32);
+    if (c >= 0) {
```
wesm: Some of the test files don't have an `Type::INT32` column I guess?

wesm:
On review I am -0 and will defer to the others. I would rather replace the existing code with unit tests and delete the test data files.

asandryh:
It looks like PARQUET-497 will introduce a lot of changes to the testing of ParquetFileReader, including changes made in this patch. To avoid modifying the code twice, I'm going to remove this PR for now and keep the changes on my branch.

@wesm Thank you for the comments, I'll address them on my branch.

PR #37, opened 2016-02-03
PARQUET-507: Reduce the runtime of rle-test

wesm:
I twiddled this a bit to cut the runtime in half. I'd like to reduce it further but looking for feedback -- my preference would be to use system entropy (`std::random_device`) to seed the PRNG and print the seed on failure. So we could run far fewer tests (e.g. only 50 or 100 or so) and occasionally run into flakiness or failure if we refactor and break something internally. Thoughts?

### DIFF
--- a/src/parquet/util/rle-test.cc
+++ b/src/parquet/util/rle-test.cc
@@ -318,14 +318,22 @@ TEST(BitRle, Flush) {
 // Test some random sequences.
 TEST(BitRle, Random) {
   int iters = 0;
-  while (iters < 1000) {
+
+  size_t niters = 500;
+  size_t ngroups = 1000;
+  size_t max_group_size = 16;
+  vector<int> values(ngroups + max_group_size);
+
+  while (iters < niters) {
     srand(iters++);
     if (iters % 10000 == 0) LOG(ERROR) << "Seed: " << iters;
-    vector<int> values;
     bool parity = 0;
-    for (int i = 0; i < 1000; ++i) {
+
+    values.resize(0);
+
+    for (int i = 0; i < ngroups; ++i) {
       int group_size = rand() % 20 + 1;  // NOLINT
-      if (group_size > 16) {
+      if (group_size > max_group_size) {
         group_size = 1;
       }
       for (int i = 0; i < group_size; ++i) {
--- a/src/parquet/util/rle-test.cc
+++ b/src/parquet/util/rle-test.cc
@@ -21,6 +21,7 @@
 #include <stdio.h>
 #include <cstdint>
 #include <iostream>
+#include <random>
 #include <vector>
 
 #include <boost/utility.hpp>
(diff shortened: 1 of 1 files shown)


julienledem:
if we randomly generate input data we can also print the corresponding input on failure.
That way it is easy to add it as a fixed test case for the input that made it fail.

wesm:
I'll see if I can make each iteration a function of a single randomly generated seed, and print that on failure, so in the event of a random failure it will be reproducible. Then we can trim the runtime down to a few hundred ms or less

src/parquet/util/rle-test.cc
```diff
@@ -318,14 +318,22 @@ TEST(BitRle, Flush) {
...
-  while (iters < 1000) {
+
+  size_t niters = 500;
+  size_t ngroups = 1000;
+  size_t max_group_size = 16;
+  vector<int> values(ngroups + max_group_size);
+
+  while (iters < niters) {
     srand(iters++);
     if (iters % 10000 == 0) LOG(ERROR) << "Seed: " << iters;
```
asandryh: Since niters = 500, this condition is always false.

src/parquet/util/rle-test.cc
```diff
@@ -318,14 +318,22 @@ TEST(BitRle, Flush) {
...
 TEST(BitRle, Random) {
   int iters = 0;
-  while (iters < 1000) {
+
+  size_t niters = 500;
+  size_t ngroups = 1000;
+  size_t max_group_size = 16;
+  vector<int> values(ngroups + max_group_size);
+
+  while (iters < niters) {
```
asandryh: A style comment: `iters` is only used inside this loop, so it's cleaner to remove line 320 and use
`for (int iters=0; iters < niters; iters++) { srand(iters);`

src/parquet/util/rle-test.cc
wesm: Note: I didn't write this code. Will clean it up more per the comments (and reducing runtime further)

asandryh: This is only a performance-improving suggestion. ;)

wesm:
@asandryh @julienledem I further shorted the runtime (whole test suite takes < 500ms for me locally) and using device entropy to generate PRNG seeds. On failure the seed is printed. Feel free to merge when build green

src/parquet/util/rle-test.cc
```diff
@@ -317,15 +344,27 @@ TEST(BitRle, Flush) {
...
-    if (iters % 10000 == 0) LOG(ERROR) << "Seed: " << iters;
-    vector<int> values;
+  size_t niters = 50;
+  size_t ngroups = 1000;
+  size_t max_group_size = 16;
+  vector<int> values(ngroups + max_group_size);
+
+  // prng setup
+  std::random_device rd;
+  std::uniform_int_distribution<int> dist(1, std::numeric_limits<int>::max());
```
asandryh: A small optimization: `std::uniform_int_distribution<int> dist(1,20);`
and change line 366 to
`int group_size = dist(gen);`

asandryh:
lgtm

wesm:
Done, thanks

julienledem:
+1

wesm:
thank you!

PR #38, opened 2016-02-04
PARQUET-442: Nested schema conversion, Thrift struct decoupling, dump-schema utility

wesm:
Several inter-related things here:

* Added SchemaDescriptor and ColumnDescriptor types to hold computed structure
  information (e.g. max ref/def levels) about the file schema. These are used
  now in the FileReader and ColumnReader
* I also added, very similar to parquet-mr (though leaned down), a logical
  schema node class structure which can be used for both the file reading and
  writing.
* Added FlatSchemaConverter to convert Parquet flat schema metadata into a
  nested logical schema
* Added a SchemaPrinter tool and parquet-dump-schema CLI tool to visit a nested
  schema and print it to the console.
* Another big thing here is that per PARQUET-446 and related work in
  parquet-mr, it's important for both the public API of this project and
  internal development to limit our coupling to the compiled Thrift headers. I
  added `Type`, `Repetition`, and `LogicalType` enums to the `parquet_cpp`
  namespace and inverted the dependency between the column readers, scanners,
  and encoders to use these enums.
* A bunch of unit tests.


wesm:
Here's what the schema dump utility looks like:

```
$ ./debug/parquet-dump-schema ~/data/tpch_nested_customer/000000_0 
required group hive_schema {
  optional int64 c_custkey
  optional byte_array c_name
  optional byte_array c_address
  optional int32 c_nationkey
  optional byte_array c_phone
  optional fixed_len_byte_array(6) c_acctbal
  optional byte_array c_mktsegment
  optional byte_array c_comment
  optional group c_orders {
    repeated group bag {
      optional group array_element {
        optional int64 o_orderkey
        optional byte_array o_orderstatus
        optional fixed_len_byte_array(6) o_totalprice
        optional byte_array o_orderdate
        optional byte_array o_orderpriority
        optional byte_array o_clerk
        optional int32 o_shippriority
        optional byte_array o_comment
        optional group o_lineitems {
          repeated group bag {
            optional group array_element {
              optional int64 l_partkey
              optional int64 l_suppkey
              optional int32 l_linenumber
              optional fixed_len_byte_array(6) l_quantity
              optional fixed_len_byte_array(6) l_extendedprice
              optional fixed_len_byte_array(6) l_discount
              optional fixed_len_byte_array(6) l_tax
              optional byte_array l_returnflag
              optional byte_array l_linestatus
              optional byte_array l_shipdate
              optional byte_array l_commitdate
              optional byte_array l_receiptdate
              optional byte_array l_shipinstruct
              optional byte_array l_shipmode
              optional byte_array l_comment
            }
          }
        }
      }
    }
  }
}
```

wesm:
On the subject of the points of contacts with the Thrift parts of the library:

```
$ grep -r thrift src/
src/parquet/reader.cc:#include "parquet/thrift/util.h"
src/parquet/schema/converter.h:#include "parquet/thrift/parquet_types.h"
src/parquet/schema/schema-descriptor-test.cc:#include "parquet/thrift/parquet_types.h"
src/parquet/schema/schema-converter-test.cc:#include "parquet/thrift/parquet_types.h"
src/parquet/column/page.h:#include "parquet/thrift/parquet_types.h"
src/parquet/column/serialized-page.cc:#include "parquet/thrift/util.h"
src/parquet/column/serialized-page.h:#include "parquet/thrift/parquet_types.h"
src/parquet/encodings/encodings.h:#include "parquet/thrift/parquet_types.h"
src/parquet/reader.h:#include "parquet/thrift/parquet_types.h"
```

Notes on some of this:

* As part of PARQUET-497, we'll be able to separate Thrift interactions from `parquet/reader.h` and `reader.cc`
* The encodings rely on the `parquet::Encoding` enum. It would likely be better to promote this to a `parquet_cpp::Encoding` enum that is defined within, and handle the mapping (if any) during deserialization.

wesm:
And...Travis CI is cranky again. I'll trigger a build in the morning, hopefully they'll have the infra issues fixed by then.

src/parquet/schema/types.h
```diff
@@ -0,0 +1,381 @@
...
+#include <cstdint>
+#include <memory>
+#include <string>
+#include <vector>
+
+#include "parquet/exception.h"
+
+namespace parquet_cpp {
+
+// Mirrors parquet::Type
```
asandryh: Is it possible to avoid this duplication?

wesm: I don't see it as duplication. We are decoupling from the Thrift compiled headers which has numerous benefits — parquet-mr does the same thing.

wesm: In particular, we need to try to make the public API never `#include <thrift/Thrift.h>` -- see PARQUET-446. It's going to take some time, but this is a pretty big deal for 3rd-party users of parquet-cpp if we can pull it off.

src/parquet/schema/converter.cc
```diff
@@ -0,0 +1,167 @@
...
+
+std::unique_ptr<Node> FlatSchemaConverter::Convert() {
+  const SchemaElement& root = elements_[0];
+
+  // Validate the root node
+  if (root.num_children == 0) {
+    throw ParquetException("Root node did not have children");
+  }
+
+  // Relaxing this restriction as some implementations don't set this
```
julienledem: but then parquet-mr does not read them?
we should make sure this does not impact repetition levels

src/parquet/schema/types.h
julienledem: 3rd-party users may be using thrift as well. This will avoid conflicts.

src/parquet/schema/printer.cc
```diff
@@ -0,0 +1,139 @@
...
+
+#include "parquet/schema/printer.h"
+
+#include <string>
+
+namespace parquet_cpp {
+
+namespace schema {
+
+class SchemaPrinter : public Node::Visitor {
```
julienledem: ideally we would have a parser/generator for the text representation of the Parquet schema as in parquet-mr.
(just a thought)

src/parquet/schema/schema-converter-test.cc
```diff
@@ -0,0 +1,208 @@
...
+  NodePtr list(new GroupNode("b", Repetition::REPEATED, {item}, LogicalType::LIST));
+  NodePtr bag(new GroupNode("bag", Repetition::OPTIONAL, {list}));
+  fields.push_back(bag);
+
+  GroupNode schema(name_, Repetition::REPEATED, fields);
+
+  ASSERT_TRUE(schema.Equals(group_));
+}
+
+TEST_F(TestSchemaConverter, InvalidRoot) {
```
julienledem: could you add a one liner describing that test?

src/parquet/schema/types.h
```diff
@@ -0,0 +1,381 @@
...
+//       <required/optional> value_type item
+//
+// 2- and 1-level encoding are respectively equivalent to 3-level encoding with
+// the non-repeated nodes set to required.
+//
+// The "official" encoding recommended in the Parquet spec is the 3-level, and
+// we use that as the default when creating list types. For semantic completeness
+// we allow the other two. Since all types of encodings will occur "in the
+// wild" we need to be able to interpret the associated definition levels in
+// the context of the actual encoding used in the file.
```
julienledem: yes unfortunately, we got to standardize on the 3-level relatively late in the project

src/parquet/schema/types.h
```diff
@@ -0,0 +1,381 @@
...
+ public:
+  PrimitiveNode(const std::string& name, Repetition::type repetition,
+      Type::type type,
+      LogicalType::type logical_type = LogicalType::NONE,
+      int id = -1) :
+      Node(Node::PRIMITIVE, name, repetition, logical_type, id),
+      physical_type_(type) {}
+
+  // FLBA ctor
+  PrimitiveNode(const std::string& name, Repetition::type repetition,
```
julienledem: maybe make it private and a static factory method with an appropriate name?

src/parquet/schema/schema-converter-test.cc
```diff
@@ -0,0 +1,208 @@
...
+  SchemaElement elements[2];
+  elements[0] = NewPrimitive("not-a-group", FieldRepetitionType::REQUIRED,
+      parquet::Type::INT32);
+  elements[0].num_children = 0;
```
julienledem: shouldn't NewPrimitive set num_children to 0 ?

src/parquet/schema/schema-builder-test.cc
```diff
@@ -0,0 +1,39 @@
...
+
+using std::string;
+using std::vector;
+
+namespace parquet_cpp {
+
+namespace test {
+
+TEST(SchemaBuilderTest, TestArray) {
+}
```
julienledem: todo?

src/parquet/schema/printer.cc
julienledem: we could reuse some of the test cases in parquet-mr to verify that the text version looks the same

wesm: yes definitely, would like to be able to generate and parse schemas formatted in this fashion

wesm: https://issues.apache.org/jira/browse/PARQUET-510

julienledem:
this looks good to me.

src/parquet/schema/converter.cc
wesm: This logic _only_ applies to the root group. Like this example schema has REQUIRED for the root group: https://github.com/apache/parquet-cpp/pull/38#issuecomment-179658491

Do you think it is safe to ignore the repetition level for the first `SchemaElement` as long as `num_children > 0`?

src/parquet/schema/schema-builder-test.cc
wesm: Yep, I'm removing it from this patch

src/parquet/schema/schema-converter-test.cc
wesm: done. I also restored the commented-out tests.

src/parquet/schema/schema-converter-test.cc
wesm: good point, will fix

src/parquet/schema/types.h
wesm: Thank you for the kick in the pants here. I'm refactoring to make the constructor private with a factory methods that constructs the node (without leaking the Thrift types) from parquet metadata

wesm: OK good enough for now

wesm:
I'm going to add a more tests for the `FlatSchemaConverter` and wrap this up. If anyone else can provide comments (especially on C++ specific stuff), will be much appreciated. Thank you kindly

wesm:
@nongli @julienledem @asandryh @majetideepak OK I am satisfied with the test coverage -- pls review and let me know any other comments or concerns. Thanks

src/parquet/schema/converter.cc
wesm: For example, Impala does not set the repetition level of the root node

https://github.com/cloudera/Impala/blob/cdh5-trunk/be/src/exec/hdfs-parquet-table-writer.cc#L733

Seems likely this will be variable depending on the origin of the file so I think best to leave as is.

wesm:
Also added a test for the schema stringifier

wesm:
@nongli @julienledem @asandryh @majetideepak any chance of completing reviews on this before we go into the weekend? I'll be available as needed until 7pm Pacific to resolve concerns so we can get to merge-ready. If not, I can base downstream patches on this branch and rebase later, it's not a big deal.

wesm:
I'd really like @nongli or @skye to take a look if possible, so no need to rush if they don't have time until next week

majetideepak:
@wesm I was away today and will be tomorrow. Sorry for the delay.

asandryh:
LGTM.

wesm:
I think this will be the target of ongoing refactoring — @julienledem if you are satisfied feel free to go ahead and merge. We can open follow-up JIRAs to tweak APIs here and there.

wesm:
Thank you very much

src/parquet/schema/types.h
julienledem: Ahah. Not intended as a kick
a gentle nudge maybe :)

src/parquet/schema/converter.cc
julienledem: Then it's probably fine. This is unnecessary info anyway.

PR #40, opened 2016-02-06
PARQUET-455: Fix OS X / Clang compiler warnings

wesm:
There actually was a legitimate bug fixed here for malformed Parquet files, but we are not yet in a position to write a decent test for it until PARQUET-497. I will make a note on that JIRA.

I also set our Travis CI build to fail on future compiler warnings. 

This also closes #15.

### DIFF
--- a/src/parquet/reader.cc
+++ b/src/parquet/reader.cc
@@ -199,7 +199,7 @@ void ParquetFileReader::ParseMetaData() {
 
   uint32_t metadata_len = *reinterpret_cast<uint32_t*>(footer_buffer);
   size_t metadata_start = filesize - FOOTER_SIZE - metadata_len;
-  if (metadata_start < 0) {
+  if (FOOTER_SIZE + metadata_len > filesize) {
     throw ParquetException("Invalid parquet file. File is less than file metadata size.");
   }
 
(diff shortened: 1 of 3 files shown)


src/parquet/reader.cc
```diff
@@ -199,7 +199,7 @@ void ParquetFileReader::ParseMetaData() {
 
   uint32_t metadata_len = *reinterpret_cast<uint32_t*>(footer_buffer);
   size_t metadata_start = filesize - FOOTER_SIZE - metadata_len;
-  if (metadata_start < 0) {
+  if (FOOTER_SIZE + metadata_len > filesize) {
```
julienledem: metadata_start < 0
<=> filesize - FOOTER_SIZE - metadata_len < 0
<=> filesize < FOOTER_SIZE + metadata_len
<=> FOOTER_SIZE + metadata_len > filesize
right?

julienledem:
LGTM, +1

src/parquet/reader.cc
wesm: yep

wesm:
Cool, fire when ready. Thank you

PR #42, opened 2016-02-06
PARQUET-513: Fail build if valgrind finds error during ctest, fix a core dump

wesm:
Also adds an option to use ccache with a custom gcc toolchain. 

Fixes an improper memory access stemming from a unit test that had a wrong parameter.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -42,6 +42,6 @@ script:
 - make
 - >
   if [ $TRAVIS_OS_NAME == linux ]; then
-    valgrind --tool=memcheck --leak-check=yes ctest;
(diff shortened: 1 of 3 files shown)


wesm:
The b46c617 build should fail (if this .travis.yml fix worked) while 43df9e7 should pass (because I fixed the core dump).

wesm:
Argh, non-deterministic behavior! Well, this patch can be merged if you are satisfied with the fix.

asandryh:
Where do you observe non-deterministic behavior?

wesm:
In the first commit I have valgrind errors locally but they did not occur on Travis CI

asandryh:
+1. LGTM

wesm:
thank you!

PR #43, opened 2016-02-08
PARQUET-502: Ensure that a scanner reads the entire column regardless…

asandryh:
… of the batch size.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -279,4 +279,5 @@ add_custom_target(clean-all
 # installation
 
 install(TARGETS parquet
-  LIBRARY DESTINATION lib)
+  LIBRARY DESTINATION lib
+  ARCHIVE DESTINATION lib)
--- a/src/parquet/column/scanner.h
+++ b/src/parquet/column/scanner.h
@@ -108,6 +108,7 @@ class TypedScanner : public Scanner {
 
       // TODO: repetition levels
 
+      value_offset_ = 0;
       level_offset_ = 0;
       if (!levels_buffered_) {
         return false;
--- a/src/parquet/reader-test.cc
+++ b/src/parquet/reader-test.cc
@@ -84,35 +84,55 @@ TEST_F(TestAllTypesPlain, TestBatchRead) {
   ASSERT_FALSE(col->HasNext());
 }
 
+void verifyScanner(Int32Scanner& scanner) {
+  int32_t val;
+  bool is_null;
+  for (size_t i = 0; i < 8; ++i) {
+    ASSERT_TRUE(scanner.HasNext());
+    ASSERT_TRUE(scanner.NextValue(&val, &is_null));
+    ASSERT_FALSE(is_null);
+  }
+  ASSERT_FALSE(scanner.HasNext());
+  ASSERT_FALSE(scanner.NextValue(&val, &is_null));
+}
+
 TEST_F(TestAllTypesPlain, TestFlatScannerInt32) {
   RowGroupReader* group = reader_.RowGroup(0);
 
   // column 0, id
   std::shared_ptr<Int32Scanner> scanner(new Int32Scanner(group->Column(0)));
-
-  int32_t val;
-  bool is_null;
-  for (size_t i = 0; i < 8; ++i) {
-    ASSERT_TRUE(scanner->HasNext());
-    ASSERT_TRUE(scanner->NextValue(&val, &is_null));
-    ASSERT_FALSE(is_null);
-  }
-  ASSERT_FALSE(scanner->HasNext());
-  ASSERT_FALSE(scanner->NextValue(&val, &is_null));
+  verifyScanner(*scanner);
 }
 
+TEST_F(TestAllTypesPlain, TestScannerSmallBatch) {
+  RowGroupReader* group = reader_.RowGroup(0);
+
+  // column 0, id
+  std::shared_ptr<Int32Scanner> scanner(new Int32Scanner(group->Column(0), 1));
+  verifyScanner(*scanner);
+}
+
+TEST_F(TestAllTypesPlain, TestScannerLargeBatch) {
+  RowGroupReader* group = reader_.RowGroup(0);
+
+  // column 0, id
+  std::shared_ptr<Int32Scanner> scanner(new Int32Scanner(group->Column(0), 1000000));
+  verifyScanner(*scanner);
+}
 
 TEST_F(TestAllTypesPlain, TestSetScannerBatchSize) {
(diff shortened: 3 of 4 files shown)


CMakeLists.txt
```diff
@@ -279,4 +279,5 @@ add_custom_target(clean-all
 # installation
 
 install(TARGETS parquet
-  LIBRARY DESTINATION lib)
+  LIBRARY DESTINATION lib
+  ARCHIVE DESTINATION lib)
```
asandryh: This is to enable static builds. Not relevant to parquet-502, but useful in general, and not worth opening a separate jira.

src/parquet/column/scanner.h
```diff
@@ -108,6 +108,7 @@ class TypedScanner : public Scanner {
 
       // TODO: repetition levels
 
+      value_offset_ = 0;
```
asandryh: That's the whole fix. :)

src/parquet/reader.cc
```diff
@@ -288,17 +288,18 @@ void ParquetFileReader::DebugPrint(std::ostream& stream, bool print_values) {
     }
     stream << "\n";
 
-    bool hasRow;
-    do {
-      hasRow = false;
+    for (size_t r = group_reader->column_metadata(0)->num_values; r >0; r--) {
```
asandryh: This is a cosmetic change, to prevent DebugPrint from printing an empty row at the end of the output.

asandryh:
Passing CI build: https://travis-ci.org/asandryh/parquet-cpp/builds/107864253

src/parquet/reader-test.cc
```diff
@@ -84,36 +84,56 @@ TEST_F(TestAllTypesPlain, TestBatchRead) {
...
+  int32_t val;
+  bool is_null;
+  for (size_t i = 0; i < 8; ++i) {
+    ASSERT_TRUE(scanner.HasNext());
+    ASSERT_TRUE(scanner.NextValue(&val, &is_null));
+    ASSERT_FALSE(is_null);
+  }
+  ASSERT_FALSE(scanner.HasNext());
+  ASSERT_FALSE(scanner.NextValue(&val, &is_null));
+}
```
wesm: It's a bummer that this only works with REQUIRED data and files with 8 values. I feel like we're digging the technical debt hole deeper by not taking this opportunity to write unit tests vs increasing our reliances on binary files. Thoughts?

src/parquet/reader-test.cc
```diff
@@ -84,36 +84,56 @@ TEST_F(TestAllTypesPlain, TestBatchRead) {
   ASSERT_FALSE(col->HasNext());
 }
 
+void verifyScanner(Int32Scanner& scanner) {
```
wesm: camel case

src/parquet/reader.cc
```diff
@@ -288,17 +288,18 @@ void ParquetFileReader::DebugPrint(std::ostream& stream, bool print_values) {
...
       for (int i = 0; i < num_columns; ++i) {
         if (scanners[i]->HasNext()) {
-          hasRow = true;
           scanners[i]->PrintNext(stream, 17);
+        } else {
+          std::string null_fmt = format_fwf<Type::BYTE_ARRAY>(17);
+          snprintf(buffer, bufsize, null_fmt.c_str(), "NULL");
+          stream << buffer;
```
wesm: When does this happen?

src/parquet/reader-test.cc
wesm: My specific point is that we really need a unit test that reproduces this bug, and so the moment we nix these temporary tests we have to go back and retroactively write the unit tests we probably should have written the first time around. I only wrote these tests here to help me do the refactoring in PARQUET-433 (which would have severely blocked progress if not gotten through quickly), but not as a long term solution.

majetideepak:
I am going spend the next couple of days adding some unit tests.

wesm:
OK, maybe combine with this patch? I'm working on other things the next 2 days so let me know if I can assist with test fixtures etc.

src/parquet/reader-test.cc
asandryh: I have started working on PARQUET-481 with the goal to remove the reliance on the files in /data. I'll submit a separate PR for that. I can take care of that first, and then rebase this PR. But until then we need to be aware of this bug in TypedScanner.

src/parquet/reader.cc
asandryh: When some of the columns have NULLs at the end.

asandryh:
@majetideepak Do you plan to add tests for FileReader and Scanner? I'm doing this as a part of PARQUET-481. Let's make sure we don't do the same work twice.

majetideepak:
@asandryh Please go ahead and implement the tests for FileReader and
Scanner.

On Mon, Feb 8, 2016 at 8:24 PM, asandryh <notifications@github.com> wrote:

> @majetideepak <https://github.com/majetideepak> Do you plan to add tests
> for FileReader and Scanner? I'm doing this as a part of PARQUET-481. Let's
> make sure we don't do the same work twice.
>
> —
> <https://github.com/apache/parquet-cpp/pull/43#issuecomment-181658740>.
>



-- 
regards,
Deepak Majeti

wesm:
@asandryh @majetideepak can you give a status update? This isn't strictly on the immediate critical path but will need to be addressed as part of more exhaustive Scanner test coverage https://issues.apache.org/jira/browse/PARQUET-526. This also overlaps with https://issues.apache.org/jira/browse/PARQUET-499

thank you

asandryh:
I've created a new PR for PARQUET-526: https://github.com/apache/parquet-cpp/pull/50. It also contains the fix for PARQUET-502.

majetideepak:
@asandryh can you please rebase your PRs. Have you started working on PARQUET-481 ? If not, we could work on it together.

wesm:
@asandryh @majetideepak I'm pretty tied up with the Spark Summit East conference this week so I will not have much time for Parquet until Friday at earliest, but we need to prioritize completing the PLAIN encoding and decoding routines and tests before continuing much further down this testing path. For example, the bug that is fixed here is difficult to test because the encoding tools for  `FIXED_LEN_BYTE_ARRAY` are incomplete (or at least untested).

wesm:
I updated the planning document to group interrelated JIRAs

https://docs.google.com/document/d/1WyquzupLc3UkErO2OhqLJNQ9a84Cccc8LVUSuLQz39o/edit#

I recommend completing PARQUET-499 before working more on this patch or #50

majetideepak:
This is a  good plan.

PR #44, opened 2016-02-09
PARQUET-505: Column reader should automatically handle large data pages

majetideepak:
This PR implements
1) PARQUET-505: Column reader should automatically handle large data pages
2) Adds support for Serialization
3) Test case for Serialization and Deserialization
4) Test case for SerializedPageReader and PARQUET-505

### DIFF
--- a/src/parquet/column/serialized-page-test.cc
+++ b/src/parquet/column/serialized-page-test.cc
@@ -1,0 +1,109 @@
+// Licensed to the Apache Software Foundation (ASF) under one
+// or more contributor license agreements.  See the NOTICE file
+// distributed with this work for additional information
+// regarding copyright ownership.  The ASF licenses this file
+// to you under the Apache License, Version 2.0 (the
+// "License"); you may not use this file except in compliance
+// with the License.  You may obtain a copy of the License at
+//
+//   http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing,
+// software distributed under the License is distributed on an
+// "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+// KIND, either express or implied.  See the License for the
+// specific language governing permissions and limitations
+// under the License.
+
+#include <cstdlib>
+#include <iostream>
+#include <sstream>
+#include <string>
+
+#include <gtest/gtest.h>
+
+#include "parquet/types.h"
+#include "parquet/thrift/parquet_types.h"
+#include "parquet/thrift/util.h"
+#include "parquet/column/serialized-page.h"
+#include "parquet/column/page.h"
+#include "parquet/column/reader.h"
+#include "parquet/column/test-util.h"
+
+
+namespace parquet_cpp {
+
+class TestSerializedPage : public ::testing::Test {
+ public:
+  void InitSerializedPageReader(const uint8_t* buffer, size_t header_size,
+      parquet::CompressionCodec::type codec) {
+    std::unique_ptr<InputStream> stream;
+    stream.reset(new InMemoryInputStream(buffer, header_size));
+    page_reader_.reset(new SerializedPageReader(std::move(stream), codec));
+  }
+
+ protected:
+  std::unique_ptr<SerializedPageReader> page_reader_;
+};
+
+TEST_F(TestSerializedPage, TestLargePageHeaders) {
+  parquet::PageHeader in_page_header;
+  parquet::DataPageHeader data_page_header;
+  parquet::PageHeader out_page_header;
+  parquet::Statistics stats;
+  int expected_header_size = 512 * 1024; //512 KB
+  int stats_size = 256 * 1024; // 256 KB
+  std::string serialized_buffer;
+  int num_values = 4141;
+
+  InitStats(stats_size, stats);
+  InitDataPage(stats, data_page_header, num_values);
+  InitPageHeader(data_page_header, in_page_header);
+
+  // Serialize the Page header
+  ASSERT_NO_THROW(serialized_buffer = SerializeThriftMsg(&in_page_header,
+      expected_header_size));
+  // check header size is between 256 KB to 16 MB
+  ASSERT_LE(stats_size, serialized_buffer.length());
+  ASSERT_GE(DEFAULT_MAX_PAGE_HEADER_SIZE, serialized_buffer.length());
+
+  InitSerializedPageReader(reinterpret_cast<const uint8_t*>(serialized_buffer.c_str()),
+      serialized_buffer.length(), parquet::CompressionCodec::UNCOMPRESSED);
+
+  std::shared_ptr<Page> current_page = page_reader_->NextPage();
+  ASSERT_EQ(parquet::PageType::DATA_PAGE, current_page->type());
+  const DataPage* page = static_cast<const DataPage*>(current_page.get());
+  ASSERT_EQ(num_values, page->num_values());
+}
+
+TEST_F(TestSerializedPage, TestFailLargePageHeaders) {
+  parquet::PageHeader in_page_header;
+  parquet::DataPageHeader data_page_header;
+  parquet::PageHeader out_page_header;
+  parquet::Statistics stats;
+  int expected_header_size = 512 * 1024; // 512 KB
+  int stats_size = 256 * 1024; // 256 KB
+  int max_header_size = 128 * 1024; // 128 KB
+  int num_values = 4141;
+  std::string serialized_buffer;
+
+  InitStats(stats_size, stats);
+  InitDataPage(stats, data_page_header, num_values);
+  InitPageHeader(data_page_header, in_page_header);
+
+  // Serialize the Page header
+  ASSERT_NO_THROW(serialized_buffer = SerializeThriftMsg(&in_page_header,
+      expected_header_size));
+  // check header size is between 256 KB to 16 MB
+  ASSERT_LE(stats_size, serialized_buffer.length());
+  ASSERT_GE(DEFAULT_MAX_PAGE_HEADER_SIZE, serialized_buffer.length());
+
+  InitSerializedPageReader(reinterpret_cast<const uint8_t*>(serialized_buffer.c_str()),
+      serialized_buffer.length(), parquet::CompressionCodec::UNCOMPRESSED);
+
+  // Set the max page header size to 128 KB, which is less than the current header size
+  page_reader_->set_max_page_header_size(max_header_size);
+
+  ASSERT_THROW(page_reader_->NextPage(), ParquetException);
+}
+} // namespace parquet_cpp
--- a/src/parquet/column/serialized-page.cc
+++ b/src/parquet/column/serialized-page.cc
@@ -33,6 +33,7 @@ namespace parquet_cpp {
 SerializedPageReader::SerializedPageReader(std::unique_ptr<InputStream> stream,
     parquet::CompressionCodec::type codec) :
     stream_(std::move(stream)) {
+  max_page_header_size_ = DEFAULT_MAX_PAGE_HEADER_SIZE;
   switch (codec) {
     case parquet::CompressionCodec::UNCOMPRESSED:
       break;
@@ -44,23 +45,42 @@ SerializedPageReader::SerializedPageReader(std::unique_ptr<InputStream> stream,
   }
 }
 
-// TODO(wesm): this may differ from file to file
-static constexpr int DATA_PAGE_SIZE = 64 * 1024;
 
 std::shared_ptr<Page> SerializedPageReader::NextPage() {
   // Loop here because there may be unhandled page types that we skip until
   // finding a page that we do know what to do with
   while (true) {
     int64_t bytes_read = 0;
-    const uint8_t* buffer = stream_->Peek(DATA_PAGE_SIZE, &bytes_read);
-    if (bytes_read == 0) {
-      return std::shared_ptr<Page>(nullptr);
+    int64_t bytes_available = 0;
+    uint32_t header_size = 0;
+    const uint8_t* buffer;
+    uint32_t allowed_page_size = DEFAULT_PAGE_HEADER_SIZE;
+    std::stringstream ss;
+
+    // Page headers can be very large because of page statistics
+    // We try to deserialize a larger buffer progressively
+    // until a maximum allowed header limit
+    while (true) {
+      buffer = stream_->Peek(allowed_page_size, &bytes_available);
+      if (bytes_available == 0) {
+        return std::shared_ptr<Page>(nullptr);
+      }
+
+      // This gets used, then set by DeserializeThriftMsg
+      header_size = bytes_available;
+      try {
+        DeserializeThriftMsg(buffer, &header_size, &current_page_header_);
+        break;
+      } catch (std::exception& e) {
+        // Failed to deserialize. Double the allowed page header size and try again
+        ss << e.what();
+        allowed_page_size *= 2;
+        if (allowed_page_size > max_page_header_size_) {
+          ss << "Deserializing page header failed.\n";
+          throw ParquetException(ss.str());
+        }
+      }
     }
-
-    // This gets used, then set by DeserializeThriftMsg
-    uint32_t header_size = bytes_read;
-    DeserializeThriftMsg(buffer, &header_size, &current_page_header_);
-
     // Advance the stream offset
     stream_->Read(header_size, &bytes_read);
 
--- a/src/parquet/column/serialized-page.h
+++ b/src/parquet/column/serialized-page.h
@@ -33,9 +33,9 @@
 namespace parquet_cpp {
 
 // 16 MB is the default maximum page header size
-static constexpr int DEFAULT_MAX_PAGE_HEADER_SIZE = 16 * 1024 * 1024;
+static constexpr uint32_t DEFAULT_MAX_PAGE_HEADER_SIZE = 16 * 1024 * 1024;
 // 16 KB is the default expected page header size
-static constexpr int DEFAULT_PAGE_HEADER_SIZE = 16 * 1024;
+static constexpr uint32_t DEFAULT_PAGE_HEADER_SIZE = 16 * 1024;
 // This subclass delimits pages appearing in a serialized stream, each preceded
 // by a serialized Thrift parquet::PageHeader indicating the type of each page
 // and the page metadata.
--- a/src/parquet/column/serialized-page.h
+++ b/src/parquet/column/serialized-page.h
@@ -32,6 +32,10 @@
 
 namespace parquet_cpp {
 
+// 16 MB is the default maximum page header size
+static constexpr int DEFAULT_MAX_PAGE_HEADER_SIZE = 16 * 1024 * 1024;
+// 16 KB is the default expected page header size
+static constexpr int DEFAULT_PAGE_HEADER_SIZE = 16 * 1024;
 // This subclass delimits pages appearing in a serialized stream, each preceded
 // by a serialized Thrift parquet::PageHeader indicating the type of each page
 // and the page metadata.
@@ -45,6 +49,10 @@ class SerializedPageReader : public PageReader {
   // Implement the PageReader interface
   virtual std::shared_ptr<Page> NextPage();
 
+  void set_max_page_header_size(uint32_t size) {
+    max_page_header_size_ = size;
+  }
+
  private:
   std::unique_ptr<InputStream> stream_;
 
@@ -54,6 +62,8 @@ class SerializedPageReader : public PageReader {
   // Compression codec to use.
   std::unique_ptr<Codec> decompressor_;
   std::vector<uint8_t> decompression_buffer_;
+  // Maximum allowed page size
+  uint32_t max_page_header_size_;
 };
 
 } // namespace parquet_cpp
--- a/src/parquet/column/test-util.h
+++ b/src/parquet/column/test-util.h
@@ -25,9 +25,9 @@
 #include <algorithm>
 #include <memory>
 #include <vector>
+#include <string>
 
 #include "parquet/column/page.h"
-
 namespace parquet_cpp {
 
 namespace test {
@@ -174,9 +174,33 @@ static std::shared_ptr<DataPage> MakeDataPage(const std::vector<T>& values,
 
   return std::make_shared<DataPage>(&(*out_buffer)[0], out_buffer->size(), page_header);
 }
-
-
 } // namespace test
+
+static inline void InitDataPage(const parquet::Statistics& stat,
+    parquet::DataPageHeader& data_page, int nvalues) {
+  data_page.encoding = parquet::Encoding::PLAIN;
+  data_page.definition_level_encoding = parquet::Encoding::RLE;
+  data_page.repetition_level_encoding = parquet::Encoding::RLE;
+  data_page.num_values = nvalues;
+  data_page.__set_statistics(stat);
+}
+
+static inline void InitStats(size_t stat_size, parquet::Statistics& stat) {
+  std::vector<char> stat_buffer;
+  stat_buffer.resize(stat_size);
+  for (int i = 0; i < stat_size; i++) {
+    (reinterpret_cast<uint8_t*>(stat_buffer.data()))[i] = i % 255;
+  }
+  stat.__set_max(std::string(stat_buffer.data(), stat_size));
+}
+
+static inline void InitPageHeader(const parquet::DataPageHeader &data_page,
+    parquet::PageHeader& page_header) {
+  page_header.__set_data_page_header(data_page);
+  page_header.uncompressed_page_size = 0;
+  page_header.compressed_page_size = 0;
+  page_header.type = parquet::PageType::DATA_PAGE;
+}
 
 } // namespace parquet_cpp
 
--- a/src/parquet/thrift/serializer-test.cc
+++ b/src/parquet/thrift/serializer-test.cc
@@ -1,0 +1,78 @@
+// Licensed to the Apache Software Foundation (ASF) under one
+// or more contributor license agreements.  See the NOTICE file
+// distributed with this work for additional information
+// regarding copyright ownership.  The ASF licenses this file
+// to you under the Apache License, Version 2.0 (the
+// "License"); you may not use this file except in compliance
+// with the License.  You may obtain a copy of the License at
+//
+//   http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing,
+// software distributed under the License is distributed on an
+// "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+// KIND, either express or implied.  See the License for the
+// specific language governing permissions and limitations
+// under the License.
+
+#include <cstdlib>
+#include <iostream>
+#include <sstream>
+#include <string>
+
+#include <gtest/gtest.h>
+
+#include "parquet/thrift/parquet_types.h"
+#include "parquet/thrift/util.h"
+#include "parquet/column/page.h"
+#include "parquet/column/reader.h"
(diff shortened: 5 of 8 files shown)


majetideepak:
@wesm feedback please. Thanks!

src/parquet/column/serialized-page-test.cc
```diff
@@ -0,0 +1,97 @@
...
+#include "parquet/thrift/util.h"
+#include "parquet/column/serialized-page.h"
+
+using std::string;
+
+namespace parquet_cpp {
+
+class TestSerializedPage : public ::testing::Test {
+ public:
+  void initSerializedPageReader(const uint8_t *buffer, size_t header_size,
```
wesm: No camel case, please =)

src/parquet/column/serialized-page-test.cc
```diff
@@ -0,0 +1,97 @@
...
+      parquet::CompressionCodec::type codec) {
+    std::unique_ptr<InputStream> stream;
+    stream.reset(new InMemoryInputStream(buffer, header_size));
+    page_reader_.reset(new SerializedPageReader(std::move(stream), codec));
+  }
+  void initStats(size_t stat_size, parquet::Statistics &stat) {
```
wesm: Generally I prefer to have a blank line in between functions.

src/parquet/column/serialized-page-test.cc
```diff
@@ -0,0 +1,97 @@
...
+    std::vector<char> stat_buffer;
+    stat_buffer.resize(stat_size);
+    for (int i = 0; i < stat_size; i++) {
+      (reinterpret_cast<uint8_t*>(stat_buffer.data()))[i] = i%255;
```
wesm: whitespace around `%`

src/parquet/column/serialized-page-test.cc
wesm: `uint8_t* buffer` instead of `uint8_t *buffer)`

src/parquet/column/serialized-page-test.cc
```diff
@@ -0,0 +1,97 @@
...
+    }
+    stat.__set_max(std::string(stat_buffer.data(), stat_size));
+  }
+  void initDataPage(parquet::DataPageHeader &data_page, parquet::Statistics &stat) {
```
wesm: `parquet::DataPageHeader& data_page`. Same with all other usages of references and pointers in this patch

wesm: This method does not need to a class instance method. Can you make this a `static inline` method in `parquet/column/test-util.h`?

src/parquet/column/serialized-page-test.cc
```diff
@@ -0,0 +1,97 @@
...
+    data_page.encoding = parquet::Encoding::PLAIN;
+    data_page.definition_level_encoding = parquet::Encoding::RLE;
+    data_page.repetition_level_encoding = parquet::Encoding::RLE;
+    data_page.num_values = 1111;
```
wesm: Make this not hardcoded per the above "make this function reusable"?

src/parquet/column/serialized-page-test.cc
```diff
@@ -0,0 +1,97 @@
...
+    data_page.__set_statistics(stat);
+  }
+  void initPageHeader(parquet::PageHeader &page_header,
+      parquet::DataPageHeader &data_page) {
+    page_header.__set_data_page_header(data_page);
+    page_header.uncompressed_page_size = 0;
+    page_header.compressed_page_size = 0;
+    page_header.type = parquet::PageType::DATA_PAGE;
```
wesm: Make this a reusable method in `column/test-util.h` also. 

`data_page` should be const here and precede the mutable argument `page_header`. Google style recommends input arguments precede output/mutated arguments

src/parquet/column/serialized-page.cc
```diff
@@ -45,23 +45,36 @@ SerializedPageReader::SerializedPageReader(std::unique_ptr<InputStream> stream,
...
+      }
 
+      // This gets used, then set by DeserializeThriftMsg
+      header_size = bytes_read;
+      if (DeserializeThriftMsg(buffer, &header_size, &current_page_header_)) {
+        break;
+      }
+      // Failed to deserialize. Double the page size and try again
+      page_sz <<= 1;
+      if (page_sz > MAX_PAGE_SIZE) {
```
wesm: Make this configurable

src/parquet/column/serialized-page.cc
```diff
@@ -45,23 +45,36 @@ SerializedPageReader::SerializedPageReader(std::unique_ptr<InputStream> stream,
   }
 }
 
-// TODO(wesm): this may differ from file to file
-static constexpr int DATA_PAGE_SIZE = 64 * 1024;
+static constexpr int PAGE_SIZE = 64 * 1024;
```
wesm: As part of making the max page serialized page header size configurable, rename this to `DEFAULT_MAX_PAGE_HEADER_SIZE` and use that as the default for the configurable variable.

src/parquet/column/serialized-page.cc
```diff
@@ -45,23 +45,36 @@ SerializedPageReader::SerializedPageReader(std::unique_ptr<InputStream> stream,
...
+static constexpr int MAX_PAGE_SIZE = 16 * 1024 * 1024;
```
wesm: 16 MB is too small (hence making this configurable); the page statistics can be arbitrarily large. Impala has observed pathological cases with page headers exceeding 2 GB.

src/parquet/column/serialized-page.cc
```diff
@@ -45,23 +45,36 @@ SerializedPageReader::SerializedPageReader(std::unique_ptr<InputStream> stream,
...
-    if (bytes_read == 0) {
-      return std::shared_ptr<Page>(nullptr);
-    }
-
-    // This gets used, then set by DeserializeThriftMsg
-    uint32_t header_size = bytes_read;
-    DeserializeThriftMsg(buffer, &header_size, &current_page_header_);
+    uint32_t header_size = 0;
+    const uint8_t* buffer;
+    int page_sz = PAGE_SIZE;
```
wesm: Use a more descriptive variable name

src/parquet/column/serialized-page.cc
```diff
@@ -45,23 +45,36 @@ SerializedPageReader::SerializedPageReader(std::unique_ptr<InputStream> stream,
...
+
+    // Retry until page header is deserialized successfully
```
wesm: Improve this comment to explain what we are retrying

src/parquet/column/serialized-page.cc
```diff
@@ -45,23 +45,36 @@ SerializedPageReader::SerializedPageReader(std::unique_ptr<InputStream> stream,
...
+    while (true) {
+      buffer = stream_->Peek(page_sz, &bytes_read);
```
wesm: `bytes_available`?

src/parquet/column/serialized-page.cc
```diff
@@ -45,23 +45,36 @@ SerializedPageReader::SerializedPageReader(std::unique_ptr<InputStream> stream,
...
+        return std::shared_ptr<Page>(nullptr);
+      }
 
+      // This gets used, then set by DeserializeThriftMsg
+      header_size = bytes_read;
+      if (DeserializeThriftMsg(buffer, &header_size, &current_page_header_)) {
+        break;
+      }
+      // Failed to deserialize. Double the page size and try again
+      page_sz <<= 1;
```
wesm: `allowed_page_size *= 2` would be clearer. I'm not a fan of bit shifting when multiplication is cleaer

src/parquet/column/serialized-page.cc
wesm: how about `allowed_page_size`

src/parquet/thrift/serializer-test.cc
```diff
@@ -0,0 +1,89 @@
...
+ void initDataPage(parquet::DataPageHeader &data_page, parquet::Statistics &stat) {
+  data_page.encoding = parquet::Encoding::PLAIN;
+  data_page.definition_level_encoding = parquet::Encoding::RLE; 
+  data_page.repetition_level_encoding = parquet::Encoding::RLE;
+  data_page.__set_statistics(stat);
+ }
+ void initPageHeader(parquet::PageHeader &page_header, 
+     parquet::DataPageHeader &data_page) {
+  page_header.__set_data_page_header(data_page);
+ }
```
wesm: Let's make these methods all reusable? Since they don't rely on any class members this only reduces our agility similar to above

src/parquet/thrift/serializer-test.cc
```diff
@@ -0,0 +1,89 @@
...
+ public:
+ void initStats(size_t stat_size, parquet::Statistics &stat) {
+  std::vector<char> stat_buffer;
+  stat_buffer.resize(stat_size);
+  for(int i = 0; i < stat_size; i++){
+    (reinterpret_cast<uint8_t*>(stat_buffer.data()))[i] = i%255;
+  }
+  stat.__set_max(std::string(stat_buffer.data(), stat_size));
+ }
+ void initDataPage(parquet::DataPageHeader &data_page, parquet::Statistics &stat) {
```
wesm: Same issue with argument ordering and const-correctness. `stat` here should be `const parquet::Statistics&` and the first argument

src/parquet/column/serialized-page-test.cc
```diff
@@ -0,0 +1,97 @@
...
+  ASSERT_LE(header_len, max_header_len);
+  ASSERT_EQ(header_len, serialized_buffer.size());
+
+  initSerializedPageReader(serialized_buffer.data(), header_len,
+      parquet::CompressionCodec::UNCOMPRESSED);
+
+  std::shared_ptr<Page> current_page = page_reader_->NextPage();
+  ASSERT_EQ(current_page->type(), parquet::PageType::DATA_PAGE);
+  const DataPage* page = static_cast<const DataPage*>(current_page.get());
+  ASSERT_EQ(page->num_values(), 1111);
```
wesm: I forgot to check in your other patch, but I realized that googletest wants you to put the expected value before the actual value. Otherwise the failure message will be unintuitive

https://github.com/google/googletest/blob/master/googletest/docs/Primer.md

So here `ASSERT_EQ(1111, page->num_values())`

src/parquet/thrift/serializer-test.cc
```diff
@@ -0,0 +1,89 @@
...
+  data_page.encoding = parquet::Encoding::PLAIN;
+  data_page.definition_level_encoding = parquet::Encoding::RLE; 
+  data_page.repetition_level_encoding = parquet::Encoding::RLE;
+  data_page.__set_statistics(stat);
+ }
+ void initPageHeader(parquet::PageHeader &page_header, 
+     parquet::DataPageHeader &data_page) {
+  page_header.__set_data_page_header(data_page);
```
wesm: set the page type?

src/parquet/thrift/serializer-test.cc
```diff
@@ -0,0 +1,89 @@
...
+  ASSERT_GE(header_len, stats_size);
+  ASSERT_LE(header_len, max_header_len);
+
+  uint32_t header_size = 1024;
+  // Deserialize the serialized page buffer
+  ASSERT_EQ(DeserializeThriftMsg(serialized_buffer.data(), &header_size, &out_page_header), true);  
+  ASSERT_GE(header_size, stats_size);
+  ASSERT_LE(header_size, max_header_len);
+
+  ASSERT_EQ(out_page_header.data_page_header.encoding, parquet::Encoding::PLAIN);
```
wesm: verify output page type is `DATA_PAGE`

src/parquet/thrift/serializer-test.cc
majetideepak: Sure! We can add a couple of more values in the header and verify.

src/parquet/column/serialized-page-test.cc
majetideepak: Will fix in all the tests

src/parquet/thrift/util.h
```diff
@@ -34,13 +35,44 @@ inline void DeserializeThriftMsg(const uint8_t* buf, uint32_t* len, T* deseriali
...
+    mem_buffer->resetBuffer();
+    obj->write(tproto.get());
+  } catch (std::exception& e) {
+    std::stringstream msg;
+    LOG(1) << "Couldn't serialize thrift object: " << e.what();
+    return false;
+  }
+  mem_buffer->getBuffer(&buf, len);
+  result->resize(*len);
+  memcpy(&((*result)[0]), buf, *len);
```
wesm: Perhaps we can avoid this copy by using `std::string`?

https://github.com/cloudera/Impala/blob/cdh5-trunk/be/src/rpc/thrift-util.h#L85

src/parquet/thrift/util.h
```diff
@@ -34,13 +35,44 @@ inline void DeserializeThriftMsg(const uint8_t* buf, uint32_t* len, T* deseriali
       tproto_factory.getProtocol(tmem_transport);
   try {
     deserialized_msg->read(tproto.get());
-  } catch (apache::thrift::protocol::TProtocolException& e) {
-    throw ParquetException("Couldn't deserialize thrift.", e);
+  } catch (std::exception& e) {
+    //TODO Determine an appproach to log exceptions
+    LOG(1) << "Couldn't deserialize thrift: " << e.what();
+    return false;
```
wesm: Hm, so this is halfway in between Impala (which uses impala::Status to communicate success/failure) and parquet_cpp (which uses exceptions). Why not raise the exception and catch it in `serialized-page.cc`, rethrowing when we exceed the maximum allowed page size?

src/parquet/column/serialized-page-test.cc
majetideepak: Reusing is a good idea. I will fix the arguments order as well.

src/parquet/thrift/util.h
```diff
@@ -34,13 +35,44 @@ inline void DeserializeThriftMsg(const uint8_t* buf, uint32_t* len, T* deseriali
...
+      apache::thrift::transport::TMemoryBuffer> tproto_factory;
+  boost::shared_ptr<apache::thrift::protocol::TProtocol> tproto =
+      tproto_factory.getProtocol(mem_buffer);
+  try {
+    mem_buffer->resetBuffer();
+    obj->write(tproto.get());
+  } catch (std::exception& e) {
+    std::stringstream msg;
+    LOG(1) << "Couldn't serialize thrift object: " << e.what();
+    return false;
```
wesm: Same issue here.

src/parquet/thrift/util.h
majetideepak: afaik Impala appends all the exceptions into `Status` and that way we can see all the exceptions generated. I feel a similar approach to provide all the exceptions generated is a better way for the end user.

src/parquet/thrift/util.h
majetideepak: We could use `std::string` too.

src/parquet/thrift/util.h
majetideepak: I will append all the exceptions inside `serialized-page.cc` and throw in this case.

src/parquet/thrift/serializer-test.cc
```diff
@@ -0,0 +1,89 @@
...
+  data_page.repetition_level_encoding = parquet::Encoding::RLE;
+  data_page.__set_statistics(stat);
+ }
+ void initPageHeader(parquet::PageHeader &page_header, 
+     parquet::DataPageHeader &data_page) {
+  page_header.__set_data_page_header(data_page);
+ }
+};
+
+TEST_F(TestThrift, TestSerializerDeserializer) {
```
wesm: BTW thanks for adding this unit test. It will also be useful to have reusable tools for microbenchmarking

src/parquet/thrift/util.h
wesm: I think this is better than copying and gives us RAII safety from memory leaks

src/parquet/column/serialized-page.cc
```diff
@@ -45,23 +45,45 @@ SerializedPageReader::SerializedPageReader(std::unique_ptr<InputStream> stream,
...
+      header_size = bytes_available;
+      try {
+        if (DeserializeThriftMsg(buffer, &header_size, &current_page_header_)) {
+          break;
+        }
+      } catch (std::exception& e) {
+        // Failed to deserialize. Double the allowed page header size and try again
+        ss << e.what();
+        allowed_page_size *= 2;
+        if (allowed_page_size > MAX_PAGE_SIZE) {
```
majetideepak: @wesm I need your help in making this configurable. Can you please elaborate ? 
I patched your other comments. Let me know if you have more feedback. Thanks!

wesm: Make the maximum allowed page size a member variable of SerializedPageReader and add a public API that sets the max allowed size. We can expose the configuration to the user via the top level `ParquetFileReader` later (and after PARQUET-497) It's OK for the default to be 16 MB.

majetideepak:
@wesm does this look okay now ?

src/parquet/column/serialized-page.cc
```diff
@@ -34,6 +34,8 @@ namespace parquet_cpp {
 SerializedPageReader::SerializedPageReader(std::unique_ptr<InputStream> stream,
     parquet::CompressionCodec::type codec) :
     stream_(std::move(stream)) {
+  max_page_header_size_ = DEFAULT_MAX_PAGE_HEADER_SIZE;
+  page_header_size_ = DEFAULT_PAGE_HEADER_SIZE;
```
wesm: This one doesn't need to be a member variable

src/parquet/column/serialized-page.cc
```diff
@@ -45,23 +47,43 @@ SerializedPageReader::SerializedPageReader(std::unique_ptr<InputStream> stream,
...
-    if (bytes_read == 0) {
-      return std::shared_ptr<Page>(nullptr);
-    }
-
-    // This gets used, then set by DeserializeThriftMsg
-    uint32_t header_size = bytes_read;
-    DeserializeThriftMsg(buffer, &header_size, &current_page_header_);
+    uint32_t header_size = 0;
+    const uint8_t* buffer;
+    int allowed_page_size = page_header_size_;
```
wesm: set the first allowed_page_size to the initial expected page header size constant here

src/parquet/column/serialized-page.h
```diff
@@ -45,6 +49,14 @@ class SerializedPageReader : public PageReader {
   // Implement the PageReader interface
   virtual std::shared_ptr<Page> NextPage();
 
+  void set_max_page_header_size(int size) {
+    max_page_header_size_ = size;
+  }
+
+  void set_page_header_size(int size) {
+    page_header_size_ = size;
+  }
```
wesm: this one isn't needed

src/parquet/column/test-util.h
```diff
@@ -174,9 +174,34 @@ class DataPageBuilder {
...
+  data_page.__set_statistics(stat);
+}
+
+static inline void InitStats(size_t stat_size, parquet::Statistics& stat) {
+  std::vector<char> stat_buffer;
+  stat_buffer.resize(stat_size);
+  for (int i = 0; i < stat_size; i++) {
+    (reinterpret_cast<uint8_t*>(stat_buffer.data()))[i] = i % 255;
+  }
+  stat.__set_max(std::string(stat_buffer.data(), stat_size));
```
wesm: This function is doing two things -- generating random string data and populating the `parquet::Statistics` object -- it would better to have two functions that do each of these tasks, but it's not a big deal right now

src/parquet/column/test-util.h
```diff
@@ -174,9 +174,34 @@ class DataPageBuilder {
...
+}
+
+static inline void InitPageHeader(const parquet::DataPageHeader &data_page,
```
wesm: `InitDataPageHeader` maybe?

src/parquet/thrift/util.h
```diff
@@ -24,7 +26,7 @@ namespace parquet_cpp {
 // all the bytes needed to store the thrift message.  On return, len will be
 // set to the actual length of the header.
 template <class T>
-inline void DeserializeThriftMsg(const uint8_t* buf, uint32_t* len, T* deserialized_msg) {
+inline bool DeserializeThriftMsg(const uint8_t* buf, uint32_t* len, T* deserialized_msg) {
```
wesm: Let's make both of these thrift utility functions `void` again

src/parquet/column/serialized-page-test.cc
```diff
@@ -0,0 +1,77 @@
...
+};
+
+TEST_F(TestSerializedPage, TestSerializedPageReader) {
+  parquet::PageHeader in_page_header;
+  parquet::DataPageHeader data_page_header;
+  parquet::PageHeader out_page_header;
+  parquet::Statistics stats;
+  uint32_t max_header_len = 16 * 1024 * 1024;
+  uint32_t header_len = 16 * 1024 * 1024;
+  int stats_size = 8 * 1024 * 1024;
```
wesm: Can you construct another test case where the `max_page_header_size_` is small and it is unable to serialize?

src/parquet/column/serialized-page-test.cc
```diff
@@ -0,0 +1,77 @@
...
+    std::unique_ptr<InputStream> stream;
+    stream.reset(new InMemoryInputStream(buffer, header_size));
+    page_reader_.reset(new SerializedPageReader(std::move(stream), codec));
+  }
+
+ protected:
+  std::unique_ptr<SerializedPageReader> page_reader_;
+};
+
+TEST_F(TestSerializedPage, TestSerializedPageReader) {
```
wesm: Change the name of this test case to something more descriptive, like "LargePageHeaders"

src/parquet/thrift/util.h
```diff
@@ -34,13 +36,41 @@ inline void DeserializeThriftMsg(const uint8_t* buf, uint32_t* len, T* deseriali
...
+      tproto_factory.getProtocol(mem_buffer);
+  try {
+    mem_buffer->resetBuffer();
+    obj->write(tproto.get());
+  } catch (std::exception& e) {
+    std::stringstream ss;
+    ss << "Couldn't serialize thrift: " << e.what() << "\n";
+    throw ParquetException(ss.str());
+  }
+  *result = mem_buffer->getBufferAsString();
```
wesm: I realized after leaving my comments yesterday that we should change this method to write into an `OutputStream` (see PARQUET-501). I think this is fine for now, but we will come back and refactor

src/parquet/column/serialized-page.cc
```diff
@@ -45,23 +47,43 @@ SerializedPageReader::SerializedPageReader(std::unique_ptr<InputStream> stream,
...
+    while (true) {
+      buffer = stream_->Peek(allowed_page_size, &bytes_available);
+      if (bytes_available == 0) {
+        return std::shared_ptr<Page>(nullptr);
+      }
 
+      // This gets used, then set by DeserializeThriftMsg
+      header_size = bytes_available;
+      try {
+        if (DeserializeThriftMsg(buffer, &header_size, &current_page_header_)) {
```
wesm: If you eliminate the return value then this `if` statement is unnecessary

src/parquet/column/serialized-page-test.cc
```diff
@@ -0,0 +1,77 @@
...
+  parquet::PageHeader in_page_header;
+  parquet::DataPageHeader data_page_header;
+  parquet::PageHeader out_page_header;
+  parquet::Statistics stats;
+  uint32_t max_header_len = 16 * 1024 * 1024;
+  uint32_t header_len = 16 * 1024 * 1024;
+  int stats_size = 8 * 1024 * 1024;
+  std::string serialized_buffer;
+  int magic_number = 4141;
```
wesm: `num_values` ?

src/parquet/column/serialized-page-test.cc
wesm: Use `ASSERT_THROW`

src/parquet/thrift/util.h
wesm: BTW this is slightly pedantic (but hey why use C++ if you don't indulge in pedantry?), but I think this calls the copy ctor whereas `return mem_buffer->getBufferAsString()` will do return-value-optiimization so a copy will not be created.

src/parquet/column/test-util.h
majetideepak: `InitDataPageHeader` will be confusing here since there is another object with that name (check `InitDataPage`) . We are just initializing the `PageHeader` object here.

src/parquet/column/test-util.h
majetideepak: I am leaving this for now. Will re-factor in the future as we add more test cases.

src/parquet/column/serialized-page.h
```diff
@@ -32,6 +32,10 @@
 
 namespace parquet_cpp {
 
+// 16 MB is the default maximum page header size
+static constexpr int DEFAULT_MAX_PAGE_HEADER_SIZE = 16 * 1024 * 1024;
+// 16 KB is the default expected page header size
+static constexpr int DEFAULT_PAGE_HEADER_SIZE = 16 * 1024;
```
wesm: Ah let's use `int64_t` here to be safer? On MSVC `int` is 32 bits so this may save us an overflow down the road.

wesm: This may not matter if Thrift can't handle sizes > UINT32_MAX

wesm:
@majetideepak after this is merged I'm going to pick up the gzip patch #11 / PARQUET-456 and enable / test all the compression codecs since I can need the test fixtures in this patch

src/parquet/column/serialized-page.h
majetideepak: Yeah, the thrift headers use `uint32_t`. `int64_t` will not help in this case.

majetideepak:
@wesm sounds good.

src/parquet/column/serialized-page-test.cc
```diff
@@ -0,0 +1,106 @@
...
+  ASSERT_EQ(num_values, page->num_values());
+}
+
+TEST_F(TestSerializedPage, TestFailLargePageHeaders) {
+  parquet::PageHeader in_page_header;
+  parquet::DataPageHeader data_page_header;
+  parquet::PageHeader out_page_header;
+  parquet::Statistics stats;
+  uint32_t header_len = 16 * 1024 * 1024;
+  // Test with 64 MB page size
```
wesm: Is this comment accurate?

majetideepak:
@nongli @julienledem can you give feedback on this patch and merge if it looks  good.

src/parquet/column/serialized-page-test.cc
majetideepak: Fixing it.
I wanted to use 64 MB initially but the test was taking too long. So I decided to decrease max page header size instead and forgot to edit this comment.

src/parquet/column/serialized-page.h
```diff
@@ -54,6 +62,8 @@ class SerializedPageReader : public PageReader {
   // Compression codec to use.
   std::unique_ptr<Codec> decompressor_;
   std::vector<uint8_t> decompression_buffer_;
+  // Maximum allowed page size
+  int max_page_header_size_;
```
majetideepak: I will change this to `uint32_t` to give us room till 4GB.

src/parquet/column/serialized-page-test.cc
wesm: You can feel free to make it much, much smaller so the test runs fast.

julienledem:
+1

julienledem:
This needs to be rebased

majetideepak:
I am squashing the commits to make rebasing easier.

majetideepak:
@julienledem This is good to go on a green build. Thank you!

PR #49, opened 2016-02-12
PARQUET-446: Hide Thrift compiled headers and Boost from public API, #include scrubbing

wesm:
This is the completion of work I started in PARQUET-442. This also resolves PARQUET-277 as no boost headers are included in the public API anymore. 

I've done some scrubbing of #includes using Google's Clang-based include-what-you-use tool. PARQUET-522 can also be resolved when this is merged.


wesm:
OK, I finished a bunch of include-what-you-use cleanup, so we can also resolve PARQUET-522 when this is merged.

Note that the wall time for compiling `libparquet.so` is about 10% faster. I'm seeing 18.2s before and 16.3s after.

src/parquet/column/scanner.h
```diff
@@ -62,14 +64,14 @@ class Scanner {
...
+  int32_t batch_size() const { return batch_size_;}
 
-  void SetBatchSize(size_t batch_size) {
+  void SetBatchSize(int32_t batch_size) {
     batch_size_ = batch_size;
   }
 
  protected:
-  size_t batch_size_;
+  int32_t batch_size_;
```
majetideepak: Can we keep this `int64_t`? Why should we limit the `batch_size` to 2GB of a byte type?

wesm: Will make `int64_t` -- data pages are limited to `int32_t` but a row group could in theory be larger.

majetideepak:
I am little concerned with having our own types for `DataPage` etc. We now have a bunch of code to construct these data pages and other structures. This code will not be useful for the write path. We will have to again use the `set` methods of the parquet types to construct objects like `parquet::DataPageHeader` etc. Can we just use the parquet types to begin with? This way we will also be heading closer to the write path.

wesm:
@majetideepak Well, we have a hard constraint, which is that exposing Thrift includes in the public API is not acceptable. What I've done is in line with parquet-mr's equivalent code (I didn't actually look until now https://github.com/apache/parquet-mr/blob/master/parquet-column/src/main/java/org/apache/parquet/column/page/DataPageV1.java)

From a software layering point of view it is better (IMHO) to decouple serialization-deserialization matters as much as possible. 

I'm happy of course to look at an alternate proposal that satisfies the constraint that developers using parquet-cpp as a 3rd-party library do not need to include Thrift libraries and headers in their build system.

wesm:
@julienledem can you review and comment? Thank you

majetideepak:
I see your point with respect to Serializer and Deserializer and `parquet-mr`.
`parquet-cpp` should be able to use any serializer and deserializer not just thrift.

src/parquet/file/reader-internal.cc
```diff
@@ -17,17 +17,138 @@
...
 #include "parquet/util/input.h"
 
 namespace parquet_cpp {
 
 // ----------------------------------------------------------------------
+// SerializedPageReader deserializes Thrift metadata and pages that have been
+// assembled in a serialized stream for storing in a Parquet files
+
+SerializedPageReader::SerializedPageReader(std::unique_ptr<InputStream> stream,
+    Compression::type codec) :
```
majetideepak: Should we name this `ThriftSerializedPageReader` since it is specific to thrift? If we want to move on the lines of `parquet-mr`, I believe there should be further de-coupling from thrift and this Reader. We probably should have another module like `parquet-cpp/parquet-thrift`.

majetideepak: This could be done in  a further downstream patch but I want to make sure we are going on the right path. @julienledem, @nongli and others, please pitch in your experience with `parquet-mr` here.

majetideepak: Alternatively, we could make `parquet-cpp` similar to `parquet-mr/parquet-thrift` (if I understood it correctly) with an additional ability to read and write "batch size" columns.

wesm: I don't agree with this line of reasoning. The Parquet specification uses Thrift for its metadata serialization, which is separate from code that can marshal materialized/reassembled records to and from Thrift or some other RPC/Serialization format (which is the role of parquet-thrift, parquet-avro, and other adapter components in parquet-mr). See for example: https://github.com/apache/parquet-mr/blob/master/parquet-thrift/src/main/java/org/apache/parquet/hadoop/thrift/ThriftToParquetFileWriter.java

julienledem: parquet-thrift in parquet-mr is not related to reading parquet metadata (which happens to use thrift) in parquet-format

We use thrift in 2 places:
 - read the parquet metadata in the file. We want that one encapsulated and invisible to the users of the library. This is why there's a separate object model additionally to the thrift idl file. The thrift IDL is here to make sure we stay backwards compatible and we have a strict spec for it. The other model is more convenient to work with and abstracts thrift out.
 - parquet-thrift is the thrift integration to let users write parquet files using a thrift IDL to define their domain specific schema. It depends on the core parquet components. Not the other way around. It is at the same level as parquet-protobuf, parquet-avro, parquet-pig ...

In particular hiding the thrift version used for the metadata will make sure that it does not conflict with the version used by the end user for their own code generation.
I hope this helps.

majetideepak: @wesm  and @julienledem thanks! for the clarification.

majetideepak:
+1 LGTM

julienledem:
+1 LGTM

wesm:
thank you!

PR #50, opened 2016-02-14
PARQUET-526: Add unit tests for Scanner.

asandryh:


src/parquet/column/scanner.h
```diff
@@ -167,6 +149,28 @@ class TypedScanner : public Scanner {
...
+  bool NextLevels(int16_t* def_level, int16_t* rep_level) {
+    bool is_required = def_levels_.size() == 0;
+    if (level_offset_ == levels_buffered_) {
+      int16_t* def_levels_data = is_required ? nullptr : &def_levels_[0];
+      levels_buffered_ = typed_reader_->ReadBatch(batch_size_, def_levels_data,
+          &rep_levels_[0], values_, &values_buffered_);
+
+      // TODO: repetition levels
+
+      value_offset_ = 0;
```
asandryh: Line 162 is the fix for PARQUET-502.

src/parquet/column/scanner-test.cc
```diff
@@ -0,0 +1,210 @@
...
+#include "parquet/types.h"
+#include "parquet/column/page.h"
+#include "parquet/column/reader.h"
+#include "parquet/column/scanner.h"
+#include "parquet/column/test-util.h"
+
+#include "parquet/util/test-common.h"
+
+using parquet::FieldRepetitionType;
+using parquet::SchemaElement;
```
wesm: These aren't used anymore but that can be addressed after rebasing #49

src/parquet/column/scanner-test.cc
```diff
@@ -0,0 +1,210 @@
...
+
+namespace parquet_cpp {
+
+using schema::NodePtr;
+
+namespace test {
+
+struct PrimitiveColumn {
+  parquet_cpp::Type::type type;
```
wesm: `parquet_cpp::` is redundant in this scope

src/parquet/column/scanner-test.cc
```diff
@@ -0,0 +1,210 @@
...
+  std::vector<int16_t> def_levels;
+  int width;
+  std::string printout;
+  PrimitiveColumn(parquet_cpp::Type::type type, const std::vector<int32_t>& values,
+      const std::vector<int16_t>& def_levels, int width, const std::string& printout) :
+    type(type), values(values), def_levels(def_levels),
+    width(width), printout(printout) {}
+};
+
+class TestPrimitiveScanner : public testing::TestWithParam<PrimitiveColumn> {
```
wesm: I am not sure how `TestWithParam` helps here -- it makes the test fixture more rigid.

src/parquet/column/scanner.h
wesm: it would be best to add a reproduction of this bug in this patch, otherwise another patch will have to be made to resolve PARQUET-502

src/parquet/column/scanner.h
```diff
@@ -167,6 +149,28 @@ class TypedScanner : public Scanner {
...
   T* values_;
+
+  bool NextLevels(int16_t* def_level, int16_t* rep_level) {
+    bool is_required = def_levels_.size() == 0;
+    if (level_offset_ == levels_buffered_) {
+      int16_t* def_levels_data = is_required ? nullptr : &def_levels_[0];
+      levels_buffered_ = typed_reader_->ReadBatch(batch_size_, def_levels_data,
+          &rep_levels_[0], values_, &values_buffered_);
+
+      // TODO: repetition levels
```
wesm: We could add this now?

wesm:
I recommend either rebasing this on the commits in #49 before that's merged or waiting for it to be merged and then rebasing, either way will work. The Thrift removal also led to some light refactoring in column/test-util.h which touches most of the test code here.

src/parquet/column/scanner-test.cc
```diff
@@ -0,0 +1,210 @@
...
+  virtual ~TestPrimitiveScanner();
+
+  TestPrimitiveScanner() {
+    auto& vals = GetParam().values;
+    auto& d_levels = GetParam().def_levels;
+    int16_t max_d_level = d_levels.size() > 0 ? 1 : 0;
+
+    std::shared_ptr<DataPage> page;
+
+    switch (GetParam().type) {
```
wesm: It might be best to refactor this to be reusable functions that are templated on Type::type -- this way it will be easier to construct test cases with multiple data pages. Since BYTE_ARRAY/FIXED_LEN_BYTE_ARRAY don't really fit this test fixture this will help make the testing approach consistent across data types

src/parquet/column/scanner-test.cc
```diff
@@ -0,0 +1,210 @@
...
+        PrimitiveColumn(Type::INT64, {11, -12, 13, -14, 15}, {1, 0, 1, 1, 0, 0, 1, 1}, 3,
+            "11 ; NULL; -12; 13 ; NULL; NULL; -14; 15 ; "),
+        PrimitiveColumn(Type::INT96, {1, 2, 3, 4}, {1, 0, 1, 1, 0, 0, 1}, 2,
+            "0 1 1 ; NULL; 0 1 2 ; 0 1 3 ; NULL; NULL; 0 1 4 ; "),
+        PrimitiveColumn(Type::BOOLEAN, {1, 1, 0, 0, 1}, {1, 0, 1, 1, 0, 0, 1, 1}, 2,
+            "1 ; NULL; 1 ; 0 ; NULL; NULL; 0 ; 1 ; "),
+        PrimitiveColumn(Type::FLOAT, {1234, -567, 89, -1}, {1, 0, 1, 0, 0, 1, 1}, 1,
+            "12.340000; NULL; -5.670000; NULL; NULL; 0.890000; -0.010000; "),
+        PrimitiveColumn(Type::DOUBLE, {1234, -567, 89, -1}, {1, 0, 1, 1, 0, 0, 1}, 1,
+            "12.340000; NULL; -5.670000; 0.890000; NULL; NULL; -0.010000; ")));
```
wesm: Testing by examining a text rendering of the output is almost guaranteed to be brittle. Can you refactor to check that values returned with each scan iteration are exactly what was inputted to the test case?

asandryh:
I will update and rebase this PR once PARQUET-499 is finished.

majetideepak:
@wesm do you have more feedback for this ?

wesm:
Yes -- high level comments

* Test data is not very diverse, and only tests with a single data page. Can you ensure that the batch size being used by the scanner to page data from the `ColumnReader` is less than the number of values being scanned? That's more important than having multiple data pages.

* Code duplication. See for example https://www.diffchecker.com/t5iymjqd. Can you: use templates to limit copy-pasting, and create a reusable scanner-tester function that can create either of the 3 cases (values only, values + def levels, values + rep + def levels). The type of approach @majetideepak used for PARQUET-515 should help here

* Does not test ByteArray / FixedLenByteArray yet, can we do that in this patch? Strictly speaking the fix for PARQUET-502 here is not complete because the original bug has not yet been reproduced. 

About to step on an airplane but will be avail throughout the day

wesm:
Lastly -- can you add test coverage for `TypedScanner<T>::FormatValue`?

asandryh:
@wesm Can you please clarify your comments?
- TestSmallBatch addresses the case when the batch size < # of read values. 
- TestSmallBatch is also the unit test for PARQUET-502. Please review the jira.
- The patch for parquet-515, and column-reader-test.cc in general, do not use any templating, and only test a few hard-coded cases for Int32 columns. What approach do you mean?

wesm:
* It would be a good idea to make sure the scanner batch-buffering works for all of the data types, especially ByteArray / FixedLenByteArray 
* The original bug report for PARQUET-502 was "Scanner segfaults when reading a FIXED_LEN_BYTE_ARRAY column". Even though the underlying cause was something else, it is a good idea in my experience to have tests cases that reproduce observed bugs if possible.
* Sorry, I meant PARQUET-499, not 515, https://github.com/apache/parquet-cpp/commit/4ae1b921bb86133f3b1be753259f20a89c128366

asandryh:
I'm adding unit tests for BA and FLBA. After that, I'll change TestSmallBatch to test with FLBA. I'll look at parquet-499 patch, too. Thanks.

src/parquet/column/scanner-test.cc
```diff
@@ -0,0 +1,386 @@
...
+  ASSERT_EQ(1000000, scanner_->batch_size());
+}
+
+// PARQUET-502
+TEST_F(TestFlatFLBAScanner, TestSmallBatch) {
+  AddPages(PAGES, 1000, {}, 0, {}, 0);
+  InitScanner(Repetition::REQUIRED);
+  scanner_->SetBatchSize(1);
+  VerifyOutput(PAGES * 1000);
+}
```
asandryh: Unit test for PARQUET-502

src/parquet/column/scanner.h
```diff
@@ -133,7 +115,7 @@ class TypedScanner : public Scanner {
     int16_t def_level;
     int16_t rep_level;
     NextLevels(&def_level, &rep_level);
-    *is_null = def_level < rep_level;
+    *is_null = def_level < descr()->max_definition_level();
```
asandryh: Fix for PARQUET-532

src/parquet/column/scanner.h
```diff
@@ -169,6 +151,31 @@ class TypedScanner : public Scanner {
...
+    bool is_required = def_levels_.size() == 0;
+    bool is_repeated = rep_levels_.size() > 0;
+    if (level_offset_ == levels_buffered_) {
+      int16_t* def_levels_data = is_required ? nullptr : &def_levels_[0];
+      int16_t* rep_levels_data = is_repeated ? &rep_levels_[0] : nullptr;
+
+      levels_buffered_ = typed_reader_->ReadBatch(batch_size_, def_levels_data,
+          rep_levels_data, values_, &values_buffered_);
+
+      value_offset_ = 0;
```
asandryh: Fix for PARQUET-502

asandryh:
@wesm @majetideepak Please review the updated PR. It contains tests for all primitive datatypes, all repetition levels, multiple pages per columns. I've fixed a few bugs in the process. This PR also incorporates fixes for PARQUET-502 and PARQUET-532.

wesm:
Can you please use `TYPED_TEST` to reduce the code duplication? https://www.diffchecker.com/xahzfgom

See

https://github.com/apache/arrow/blob/master/cpp/src/arrow/types/primitive-test.cc#L200

for an example usage. I'm reviewing the rest of the patch now.

src/parquet/column/scanner-test.cc
```diff
@@ -0,0 +1,386 @@
...
+         a.value[1] == b.value[1] &&
+         a.value[2] == b.value[2];
+}
+
+bool operator==(const ByteArray a, const ByteArray b) {
+  return a.len == b.len && 0 == memcmp(a.ptr, a.ptr, a.len);
+}
+
+static int FLBA_LENGTH = 8;
+bool operator==(const FixedLenByteArray a, const FixedLenByteArray b) {
```
wesm: should these all be const references?

src/parquet/column/scanner-test.cc
```diff
@@ -0,0 +1,386 @@
...
+
+namespace test {
+
+template <int TYPE>
+class TestFlatScanner : public ::testing::Test {
+ public:
+  void SetUp() {
+    values_.resize(0);
+    std::random_device rd;
+    SEED = rd();
```
wesm: It's better to use a deterministic seed

wesm: (more reproducible, I mean)

src/parquet/column/scanner-test.cc
```diff
@@ -0,0 +1,386 @@
...
+
+  void VerifyOutput(size_t total_values) {
+    TypedScanner<TYPE>* scanner = reinterpret_cast<TypedScanner<TYPE>* >(scanner_.get());
+    T val;
+    bool is_null;
+    size_t j = 0;
+    for (size_t i = 0; i < total_values; i++) {
+      ASSERT_TRUE(scanner->HasNext());
+      ASSERT_TRUE(scanner->NextValue(&val, &is_null));
+      if (!is_null)
```
wesm: braces here

src/parquet/column/scanner-test.cc
```diff
@@ -0,0 +1,386 @@
...
+  }
+
+  void VerifyOutput(size_t total_values) {
+    TypedScanner<TYPE>* scanner = reinterpret_cast<TypedScanner<TYPE>* >(scanner_.get());
+    T val;
+    bool is_null;
+    size_t j = 0;
+    for (size_t i = 0; i < total_values; i++) {
+      ASSERT_TRUE(scanner->HasNext());
+      ASSERT_TRUE(scanner->NextValue(&val, &is_null));
```
wesm: The `Scanner` API is deficient as you cannot scan the values and levels (as a user) at the same time if I'm reading the code right. Can you open a JIRA to add a `Scanner::Next(int16_t* rep_level, int16_t* def_level, T* val, bool* is_null)` API? We can return and improve the test case as part of that JIRA

src/parquet/column/scanner-test.cc
```diff
@@ -0,0 +1,386 @@
...
+      if (!is_null)
+        ASSERT_EQ(values_[j++], val);
+    }
+    ASSERT_FALSE(scanner->HasNext());
+  }
+
+  std::list< std::vector<uint8_t> > buffers;
```
wesm: trailing `_` on this varaible

src/parquet/column/scanner-test.cc
```diff
@@ -0,0 +1,386 @@
...
+            rep_levels, max_rep_level, &buffers.back()));
+    values_.insert(values_.end(), page.begin(), page.end());
+  }
+}
+
+template <>
+void TestFlatScanner<Type::BYTE_ARRAY>::AddPages(size_t pages, size_t num_values,
+    std::vector<int16_t> def_levels, int16_t max_def_level,
+    std::vector<int16_t> rep_levels, int16_t max_rep_level) {
+  int max_byte_array_len = 12 + sizeof(uint32_t);
```
wesm: If this is going to be hard-coded, make it a constant?

src/parquet/column/scanner.h
```diff
@@ -169,6 +151,31 @@ class TypedScanner : public Scanner {
   inline void FormatValue(void* val, char* buffer, size_t bufsize, size_t width);
 
   T* values_;
+
+  bool NextLevels(int16_t* def_level, int16_t* rep_level) {
+    bool is_required = def_levels_.size() == 0;
+    bool is_repeated = rep_levels_.size() > 0;
```
wesm: Should these flags come from the `ColumnDescriptor`?

src/parquet/column/scanner.h
```diff
@@ -169,6 +151,31 @@ class TypedScanner : public Scanner {
...
+          rep_levels_data, values_, &values_buffered_);
+
+      value_offset_ = 0;
+      level_offset_ = 0;
+      if (!levels_buffered_) {
+        return false;
+      }
+    }
+    *rep_level = is_repeated ?
+        def_levels_[level_offset_] : descr()->max_definition_level();
```
wesm: `rep_levels_` here -- I revise my prior note. We need to test the output of this method -- what do you think about the proposed API?

I believe that in the non-repeated case (`max_repetition_level == 0`), the rep_level is always 0.

src/parquet/util/test-common.h
```diff
@@ -183,6 +196,22 @@ void random_byte_array(int n, uint32_t seed, uint8_t *buf,
...
+  std::uniform_int_distribution<int> d2(0, 255);
+  for (int i = 0; i < n; ++i) {
+    out[i].len = d1(gen);
+    out[i].ptr = buf;
+    for (int j = 0; j < out[i].len; ++j) {
+      buf[j] = d2(gen) & 0xFF;
+    }
+    buf += out[i].len;
+  }
+}
```
wesm: why duplicate the existing functions? 

https://www.diffchecker.com/g9jevnz6

src/parquet/column/scanner-test.cc
wesm: per my note below, I think we need to do this in this patch to be able to verify the scanner's correctness

src/parquet/column/scanner-test.cc
wesm: This is fine, but I was confused by the presence of `sizeof(uint32_t)` -- shouldn't you pass `12` (in effect) as the last argument of `random_byte_array`? Otherwise you could possibly write past the end of `data_buffer_`.

src/parquet/column/scanner-test.cc
```diff
@@ -0,0 +1,386 @@
...
+      values_.insert(values_.end(), page.begin(), page.end());
+    }
+  }
+
+  void VerifyOutput(size_t total_values) {
+    TypedScanner<TYPE>* scanner = reinterpret_cast<TypedScanner<TYPE>* >(scanner_.get());
+    T val;
+    bool is_null;
+    size_t j = 0;
+    for (size_t i = 0; i < total_values; i++) {
```
wesm: I would prefer that the scanner batch size be set here to something smaller than `total_values` here

src/parquet/column/scanner-test.cc
wesm: I took a look at the PARQUET-499 patch and I missed this there too. Let me know if you agree on the proposed fix

majetideepak: I messed this up in PARQUET-499 while templating. I will fix it in my next patch. We should pass 12 effectively. That's the idea behind adding `sizeof(uint32_t)` explicitly.

src/parquet/column/scanner.h
majetideepak: `descr()->max_definition_level()` and `descr()->max_repetition_level()` are swapped.

wesm:
How's the timeline for this patch looking?

src/parquet/column/scanner-test.cc
```diff
@@ -0,0 +1,386 @@
...
+         a.value[2] == b.value[2];
+}
+
+bool operator==(const ByteArray a, const ByteArray b) {
+  return a.len == b.len && 0 == memcmp(a.ptr, a.ptr, a.len);
+}
+
+static int FLBA_LENGTH = 8;
+bool operator==(const FixedLenByteArray a, const FixedLenByteArray b) {
+  return 0 == memcmp(a.ptr, a.ptr, FLBA_LENGTH);
```
wesm: this should be `b.ptr`

src/parquet/column/scanner-test.cc
```diff
@@ -0,0 +1,386 @@
...
+using schema::NodePtr;
+
+bool operator==(const Int96 a, const Int96 b) {
+  return a.value[0] == b.value[0] &&
+         a.value[1] == b.value[1] &&
+         a.value[2] == b.value[2];
+}
+
+bool operator==(const ByteArray a, const ByteArray b) {
+  return a.len == b.len && 0 == memcmp(a.ptr, a.ptr, a.len);
```
wesm: `a.ptr, b.ptr`

PR #51, opened 2016-02-14
PARQUET-515: Add "SetData" to LevelDecoder

majetideepak:
This PR implements a SetData interface for the LevelDecoder class similar to existing value decoders. 
This PR also adds a test for PARQUET-523

### DIFF
--- a/src/parquet/column/column-reader-test.cc
+++ b/src/parquet/column/column-reader-test.cc
@@ -181,6 +181,62 @@ TEST_F(TestPrimitiveReader, TestInt32FlatRepeated) {
   ASSERT_EQ(0, batch_actual);
   ASSERT_EQ(0, values_read);
 }
-} // namespace test
 
+TEST_F(TestPrimitiveReader, TestInt32FlatRepeatedMultiplePages) {
+  vector<int32_t> values[2] = {{1, 2, 3, 4, 5},
+    {6, 7, 8, 9, 10}};
+  vector<int16_t> def_levels[2] = {{2, 1, 1, 2, 2, 1, 1, 2, 2, 1},
+    {2, 2, 1, 2, 1, 1, 2, 1, 2, 1}};
+  vector<int16_t> rep_levels[2] = {{0, 1, 1, 0, 0, 1, 1, 0, 0, 1},
+    {0, 0, 1, 0, 1, 1, 0, 1, 0, 1}};
+
+  std::vector<uint8_t> buffer[4];
+  std::shared_ptr<DataPage> page;
+
+  for (int i = 0; i < 4; i++) {
+    page = MakeDataPage<Type::INT32>(values[i % 2],
+        def_levels[i % 2], 2, rep_levels[i % 2], 1, &buffer[i]);
+    pages_.push_back(page);
+  }
+
+  NodePtr type = schema::Int32("a", Repetition::REPEATED);
+  ColumnDescriptor descr(type, 2, 1);
+  InitReader(&descr);
+
+  Int32Reader* reader = static_cast<Int32Reader*>(reader_.get());
+
+  size_t values_read = 0;
+  size_t batch_actual = 0;
+
+  vector<int32_t> vresult(3, -1);
+  vector<int16_t> dresult(5, -1);
+  vector<int16_t> rresult(5, -1);
+
+  for (int i = 0; i < 4; i++) {
+    batch_actual = reader->ReadBatch(5, &dresult[0], &rresult[0],
+        &vresult[0], &values_read);
+    ASSERT_EQ(5, batch_actual);
+    ASSERT_EQ(3, values_read);
+
+    ASSERT_TRUE(vector_equal(vresult, slice(values[i % 2], 0, 3)));
+    ASSERT_TRUE(vector_equal(dresult, slice(def_levels[i % 2], 0, 5)));
+    ASSERT_TRUE(vector_equal(rresult, slice(rep_levels[i % 2], 0, 5)));
+
+    batch_actual = reader->ReadBatch(5, &dresult[0], &rresult[0],
+        &vresult[0], &values_read);
+    ASSERT_EQ(5, batch_actual);
+    ASSERT_EQ(2, values_read);
+
+    ASSERT_TRUE(vector_equal(slice(vresult, 0, 2), slice(values[i % 2], 3, 5)));
+    ASSERT_TRUE(vector_equal(dresult, slice(def_levels[i % 2], 5, 10)));
+    ASSERT_TRUE(vector_equal(rresult, slice(rep_levels[i % 2], 5, 10)));
+  }
+  // EOS, pass all nullptrs to check for improper writes. Do not segfault /
+  // core dump
+  batch_actual = reader->ReadBatch(5, nullptr, nullptr,
+      nullptr, &values_read);
+  ASSERT_EQ(0, batch_actual);
+  ASSERT_EQ(0, values_read);
+}
+} // namespace test
 } // namespace parquet_cpp
--- a/src/parquet/column/levels-test.cc
+++ b/src/parquet/column/levels-test.cc
@@ -16,8 +16,8 @@
 // under the License.
 
 #include <cstdint>
-#include <string>
 #include <vector>
+#include <string>
 
 #include <gtest/gtest.h>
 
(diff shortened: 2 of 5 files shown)


majetideepak:
@wesm feedback please :)

src/parquet/util/bit-stream-utils.h
```diff
@@ -118,6 +118,8 @@ class BitReader {
     max_bytes_ = buffer_len;
     byte_offset_ = 0;
     bit_offset_ = 0;
+    int num_bytes = std::min(8, max_bytes_ - byte_offset_);
+    memcpy(&buffered_values_, buffer_ + byte_offset_, num_bytes);
   }
```
majetideepak: @wesm This is most likely a bug in the Impala code as well since we  pulled this code from there.

wesm:
Does this conflict with #49?

majetideepak:
Yes, it will conflict in the levels part. I will resolve the conflicts. My plan is to get your feedback meanwhile and keep it ready.

src/parquet/column/levels-test.cc
```diff
@@ -107,10 +150,9 @@ TEST_F(TestLevels, TestEncodeDecodeLevels) {
     parquet::Encoding::type encoding = encodings[encode];
     // BIT_PACKED requires a sequence of atleast 8
     if (encoding == parquet::Encoding::BIT_PACKED) min_repeat_factor = 3;
-
+    int num_levels_per_width = ((2 << max_repeat_factor) - (1 << min_repeat_factor));
```
wesm: Can we remove this slightly esoteric variable altogether and the `num_levels` variable too? Instead, use `push_back` to append generated levels to `input_levels`

src/parquet/column/column-reader-test.cc
```diff
@@ -183,6 +183,62 @@ TEST_F(TestPrimitiveReader, TestInt32FlatRepeated) {
...
+    {0, 0, 1, 0, 1, 1, 0, 1, 0, 1}};
+
+  std::vector<uint8_t> buffer[4];
+  std::shared_ptr<DataPage> page;
+
+  for (int i = 0; i < 4; i++) {
+    page = MakeDataPage<Type::INT32>(values[i % 2],
+        def_levels[i % 2], 2, rep_levels[i % 2], 1, &buffer[i]);
+    pages_.push_back(page);
+  }
```
wesm: This is definitely the limit of what we should do for ad hoc test page generation (except when fixing specific bugs is called for). If not here, the next step should be to add a simple algorithm to "paginate" otherwise contiguous arrays of levels and values into data pages. For example, if you had 10,000 randomly-generated values (some percentage null, so there would be 10,000 levels and fewer actual non-null values), then the paginator could break this into pages of size 1000 or less.

src/parquet/column/levels-test.cc
```diff
@@ -121,8 +163,42 @@ TEST_F(TestLevels, TestEncodeDecodeLevels) {
...
+  int min_repeat_factor = 3;
+  int max_repeat_factor = 7; // 128
+  int bit_width = 8;
+  int max_level = (1 << bit_width) - 1;
+  std::vector<int16_t> input_levels;
+  std::vector<std::vector<uint8_t>> bytes;
+  parquet::Encoding::type encodings[2] = {parquet::Encoding::RLE,
+      parquet::Encoding::BIT_PACKED};
+  int num_levels_per_width = ((2 << max_repeat_factor) - (1 << min_repeat_factor));
+  int num_levels = (bit_width + 1) * num_levels_per_width;
```
wesm: same comment re removing these variables if possible

src/parquet/column/levels.h
```diff
@@ -95,23 +95,32 @@ class LevelDecoder {
  public:
   LevelDecoder() {}
 
-  // Initialize the LevelDecoder and return the number of bytes consumed
-  size_t Init(parquet::Encoding::type encoding, int16_t max_level,
+  // Initialize the LevelDecoder
+  void Init(int16_t max_level) {
+    bit_width_ = BitUtil::Log2(max_level + 1);
+    rle_decoder_.reset(new RleDecoder(NULL, 0, 0));
+    bit_packed_decoder_.reset(new BitReader(NULL, 0));
```
wesm: Perhaps only initialize the appropriate decoder, and only if `max_level > 0` -- does the Parquet spec indicate whether data pages could switch encodings mid-stream -- in DataPageV2 there is no bit-packed encoding anymore so this is less of a concern there

src/parquet/column/levels.h
```diff
@@ -123,8 +132,8 @@ class LevelDecoder {
   // Decodes a batch of levels into an array and returns the number of levels decoded
   size_t Decode(size_t batch_size, int16_t* levels) {
     size_t num_decoded = 0;
-    if (!rle_decoder_ && !bit_packed_decoder_) {
-      throw ParquetException("Level decoders are not initialized.");
+    if (num_values_remaining_ == 0) {
+      throw ParquetException("Level decoders are not initialized with data.");
     }
```
wesm: Hypothetically we could have empty data pages, so the original if statement was checking something different

src/parquet/column/reader.cc
```diff
@@ -33,7 +33,10 @@ ColumnReader::ColumnReader(const ColumnDescriptor* descr,
   : descr_(descr),
     pager_(std::move(pager)),
     num_buffered_values_(0),
-    num_decoded_values_(0) {}
+    num_decoded_values_(0) {
+      repetition_level_decoder_.Init(descr->max_repetition_level());
+      definition_level_decoder_.Init(descr->max_definition_level());
```
wesm: Thoughts on combining `Init` and `Reset` (so this could be removed here), and `Reset` would have to add a null check on the decoder of interest? Then these could be removed

src/parquet/column/column-reader-test.cc
majetideepak: Agreed!

src/parquet/column/levels.h
majetideepak: You are right, the previous if check for null still should be there.

src/parquet/util/bit-stream-utils.h
wesm: Good catch, I agree that it looks like a bug. I will look at reporting on the Impala JIRA. I looked through the RLE encoding code and the reason that this bug was never hit in in real code was that other parts of the RleDecoder code path (in BitReader::GetAligned) trigger the `buffered_values_` value to get data copied into it. 

Separately, I looked at parquet-format more carefully and it appears that BIT_PACKED uses MSB ordering while the RLE encoding uses LSB bit-packing for its bit-packed literal runs. This means we would be unable to correctly read files generated using the correct MSB-bit-packing. 

But since BIT_PACKED is deprecated this all suggests that it never saw much production use. Let's leave this as is for now and we can investigate further.

src/parquet/column/levels-test.cc
majetideepak: done!

src/parquet/column/levels.h
majetideepak: I couldn't find any details about the encoding switch.
However, the issue here is that the level encoding is specified in the `DataPageHeader`. This is probably to allow different encodings mid-stream. The `Init` is called once during the instantiation of the column reader. I will have a check for `max_level > 0` there.

src/parquet/column/reader.cc
majetideepak: The `Init` also creates a decoder object(instance). The `Reset` does not do this.

wesm: Right, what I was saying is that the first time Reset is called it would have to create the right decoder instance if it is null (i.e. on the first invocation). When the ColumnReader ctor is called the level encoding type is not known.

src/parquet/column/levels.h
majetideepak: If the level decoders are not initialized either due to empty data pages or "required" data, the semantics of `Decode` should simply return 0 elements. If on the other hand, the decoders are initialized, then an exception must be thrown for `Decode` on zero remaining elements. If we want to simplify, we could return zero on both these cases.  Does this sound reasonable ?

src/parquet/column/levels.h
wesm: If you move the initialization-from-null to Reset as discussed below then there won't be an issue (if the encoding were to change mid-stream then the then-null decoder would be initialized). But per my comment at the end BIT_PACKED may not be used anywhere at the moment out in the wild

src/parquet/column/levels.h
wesm: That seems fine. Can you add a test case with an empty data page (num_values == 0) to make sure this returns 0?

src/parquet/column/reader.cc
majetideepak: I prefer to have clean semantics for the decoder without bothering about how the upper layers call `Reset` and  `Decode` for various cases. The Reset should not be bothered about creating an instance. It should simply re-initialize the state of the decoder.  Similarly the `Decode` should return values only if there are values to decode. Else it will simply return zero elements.

wesm: Well, I don't think we should be creating a `BitReader` instance if we don't ever use it.

wesm: How about calling `Reset` instead `SetData` like the value decoder classes to avoid hazy semantics?

majetideepak: Essentially, the upper layers should have the flexibility to create an instance or reset the state of the instance.

src/parquet/column/levels.h
majetideepak: Sure!

src/parquet/column/reader.cc
wesm: I think the SetData name (and removing Init) mitigates both our concerns and also makes this consistent with the value decoders.

majetideepak: `SetData` is a good idea. I checked in the modified code with these changes. It includes a test for checking 0 values.

wesm:
Let me know when this is rebased on #49 -- up to you whether you wait til that lands in master.

majetideepak:
@wesm  I am merging my commits to simply the merge

majetideepak:
@julienledem this is good to go. Thanks you!

src/parquet/column/levels.h
```diff
@@ -23,6 +23,7 @@
 #include "parquet/exception.h"
 #include "parquet/types.h"
 #include "parquet/util/rle-encoding.h"
+#include <algorithm>
```
wesm: Missed this before. System includes go before library includes

wesm:
+1 (outside the minor include ordering issue)

src/parquet/column/levels.h
majetideepak: done!

majetideepak:
the include order is fixed

src/parquet/util/bit-stream-utils.h
wesm: reported this in https://issues.cloudera.org/browse/IMPALA-3000.

majetideepak: Yea, let's investigate this later.

PR #52, opened 2016-02-17
PARQUET-499: Complete PlainEncoder implementation for all primitive types and test end to end

majetideepak:
Includes tests for end to end plain encoding and decoding of all data types.

### DIFF
--- a/src/parquet/column/reader.h
+++ b/src/parquet/column/reader.h
@@ -95,6 +95,12 @@ class ColumnReader {
   // The number of values from the current data page that have been decoded
   // into memory
   int num_decoded_values_;
+
+  // We need to store the data page for BYTE_ARRAY and FLBA since the corresponding types
+  // store pointers into these pages
+  // The user is responsible to free these DataPages
+  // todo(mdeepak) figure out a way to efficiently discard these pages
+  std::vector<std::shared_ptr<DataPage>> byte_pages_;
 };
 
 // API to read values from a single column. This is the main client facing API.
--- a/src/parquet/column/reader.h
+++ b/src/parquet/column/reader.h
@@ -103,9 +103,8 @@ class ColumnReader {
 
   // We need to store the data page for BYTE_ARRAY and FLBA since the corresponding types
   // store pointers into these pages
-  // The user is responsible to free these DataPages
-  // todo(mdeepak) figure out a way to efficiently discard these pages
-  std::vector<std::shared_ptr<DataPage> > byte_pages_;
+  // The user is responsible for freeing these DataPages by calling ClearDataPages()
+  std::vector<std::shared_ptr<Page> > byte_pages_;
 };
 
 // API to read values from a single column. This is the main client facing API.
--- a/src/parquet/column/reader.h
+++ b/src/parquet/column/reader.h
@@ -61,10 +61,6 @@ class ColumnReader {
     return descr_;
   }
 
-  void ClearDataPages() {
-    byte_pages_.clear();
-  }
-
  protected:
   virtual bool ReadNewPage() = 0;
 
@@ -100,11 +96,6 @@ class ColumnReader {
   // The number of values from the current data page that have been decoded
   // into memory
   int num_decoded_values_;
-
-  // We need to store the data page for BYTE_ARRAY and FLBA since the corresponding types
-  // store pointers into these pages
-  // The user is responsible for freeing these DataPages by calling ClearDataPages()
-  std::vector<std::shared_ptr<Page> > byte_pages_;
 };
 
 // API to read values from a single column. This is the main client facing API.
--- a/src/parquet/column/reader.h
+++ b/src/parquet/column/reader.h
@@ -23,7 +23,6 @@
 #include <cstring>
 #include <memory>
 #include <unordered_map>
-#include <vector>
 
 #include "parquet/column/levels.h"
 #include "parquet/column/page.h"
--- a/src/parquet/column/reader.h
+++ b/src/parquet/column/reader.h
@@ -23,6 +23,7 @@
 #include <cstring>
 #include <memory>
 #include <unordered_map>
+#include <vector>
 
 #include "parquet/column/levels.h"
 #include "parquet/column/page.h"
@@ -58,6 +59,10 @@ class ColumnReader {
 
   const ColumnDescriptor* descr() const {
     return descr_;
+  }
+
+  void ClearDataPages() {
+    byte_pages_.clear();
   }
 
  protected:
@@ -100,7 +105,7 @@ class ColumnReader {
   // store pointers into these pages
   // The user is responsible to free these DataPages
   // todo(mdeepak) figure out a way to efficiently discard these pages
-  std::vector<std::shared_ptr<DataPage>> byte_pages_;
+  std::vector<std::shared_ptr<DataPage> > byte_pages_;
 };
 
 // API to read values from a single column. This is the main client facing API.
--- a/src/parquet/encodings/plain-encoding-test.cc
+++ b/src/parquet/encodings/plain-encoding-test.cc
@@ -253,6 +253,7 @@ TEST(FLBATest, TestEncodeDecode) {
   size_t nbytes = nvalues * flba_length;
 
   vector<FLBA> draws;
+  draws.resize(nbytes);
   std::vector<uint8_t> data_buffer;
   data_buffer.resize(nbytes);
   // seed the prng so failure is deterministic
--- a/src/parquet/encodings/plain-encoding-test.cc
+++ b/src/parquet/encodings/plain-encoding-test.cc
@@ -67,187 +67,153 @@ TEST(BooleanTest1, TestEncodeDecode) {
   }
 }
 
-TEST(BooleanTest2, TestEncodeDecode) {
-  size_t nvalues = 10000;
-  size_t nbytes = BitUtil::Ceil(nvalues, 8);
+template<typename T, int TYPE>
+class EncodeDecode{
+ public:
+  void init_data(int nvalues) {
+    num_values_ = nvalues;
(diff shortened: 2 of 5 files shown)


majetideepak:
@wesm Please suggest feedback. Thanks!

src/parquet/column/reader.h
```diff
@@ -95,6 +100,11 @@ class ColumnReader {
   // The number of values from the current data page that have been decoded
   // into memory
   int num_decoded_values_;
+
+  // We need to store the data page for BYTE_ARRAY and FLBA since the corresponding types
+  // store pointers into these pages
+  // The user is responsible for freeing these DataPages by calling ClearDataPages()
+  std::vector<std::shared_ptr<Page> > byte_pages_;
```
wesm: I'm unconvinced by this argument. The `Page` objects do not own their memory either

src/parquet/encodings/plain-encoding-test.cc
```diff
@@ -63,10 +63,281 @@ TEST(BooleanTest, TestEncodeDecode) {
...
+  size_t values_decoded = decoder.Decode(
+      reinterpret_cast<int32_t*>(decode_buffer.data()), nvalues);
+  ASSERT_EQ(nvalues, values_decoded);
+
+  for (size_t i = 0; i < nvalues; ++i) {
+    ASSERT_EQ(draws[i], decode_data[i]) << i;
+  }
+}
+
+TEST(Int64Test, TestEncodeDecode) {
```
wesm: Any possibility of using templates to reduce code duplication between this and the other primitive type functions?

src/parquet/util/test-common.h
```diff
@@ -104,8 +115,72 @@ void random_bytes(int n, uint32_t seed, std::vector<uint8_t>* out) {
   }
 }
 
-} // namespace test
+template <typename T>
+void random_int_numbers(int n, uint32_t seed, std::vector<T>* out) {
+  std::mt19937 gen(seed);
+  std::uniform_int_distribution<T> d(-10000, 10000);
```
wesm: The behavior of this template is not defined for `Int96` if I understand the C++11 documentation correctly. I think we're going to need a separate data generator for Int96 that generates 3 random `uint32_t` values

wesm: It would be better to use more precise limits using `std::numeric_limits`

src/parquet/encodings/plain-encoding-test.cc
```diff
@@ -63,10 +63,281 @@ TEST(BooleanTest, TestEncodeDecode) {
...
+
+  decoder.SetData(nvalues, &encode_buffer[0], encode_buffer.size());
+  size_t values_decoded = decoder.Decode(
+      reinterpret_cast<ByteArray*>(decode_buffer.data()), nvalues);
+  ASSERT_EQ(nvalues, values_decoded);
+
+  for (size_t i = 0; i < nvalues; ++i) {
+    ASSERT_EQ(draws[i].len, decode_data[i].len) << i;
+    for (size_t j = 0; j < draws[i].len; ++j) {
+      ASSERT_EQ(draws[i].ptr[j], decode_data[i].ptr[j]) << i;
```
wesm: Maybe use `memcmp` here?

src/parquet/encodings/plain-encoding.h
```diff
@@ -72,11 +72,11 @@ inline int PlainDecoder<Type::BYTE_ARRAY>::Decode(ByteArray* buffer,
     int max_values) {
   max_values = std::min(max_values, num_values_);
   for (int i = 0; i < max_values; ++i) {
-    buffer[i].len = *reinterpret_cast<const uint32_t*>(data_);
-    if (len_ < sizeof(uint32_t) + buffer[i].len) ParquetException::EofException();
+    int len = buffer[i].len = *reinterpret_cast<const uint32_t*>(data_);
```
wesm: `uint32_t` here maybe

src/parquet/util/test-common.h
```diff
@@ -95,6 +96,16 @@ static inline vector<bool> flip_coins(size_t n, double p) {
   return draws;
 }
 
+void random_bools(int n, uint32_t seed, bool* out) {
+  std::mt19937 gen(seed);
+  std::uniform_int_distribution<int> d(0, 1);
+
+  for (int i = 0; i < n; ++i) {
+    out[i] = (d(gen) & 0x1);
```
wesm: Use `std::bernoulli_distribution` here

src/parquet/util/test-common.h
wesm: Oops I missed the specialization below

src/parquet/util/test-common.h
```diff
@@ -104,8 +115,72 @@ void random_bytes(int n, uint32_t seed, std::vector<uint8_t>* out) {
...
+  }
+}
+
+template <>
+void random_int_numbers(int n, uint32_t seed, std::vector<Int96>* out) {
+  std::mt19937 gen(seed);
+  std::uniform_int_distribution<int> d(-10000, 10000);
+
+  for (int i = 0; i < n; ++i) {
+    Int96 temp;
```
wesm: Resize `out` first then assign into a `const Int96&` to avoid copies

src/parquet/util/test-common.h
wesm: (unnecessary parens)

src/parquet/util/test-common.h
```diff
@@ -104,8 +115,72 @@ void random_bytes(int n, uint32_t seed, std::vector<uint8_t>* out) {
...
+    temp.value[0] = (d(gen));
+    temp.value[1] = (d(gen));
+    temp.value[2] = (d(gen));
```
wesm: unneeded parens

src/parquet/util/test-common.h
```diff
@@ -104,8 +115,72 @@ void random_bytes(int n, uint32_t seed, std::vector<uint8_t>* out) {
...
+    }
+    buf += len;
+    out->push_back(temp);
+  }
+}
+
+void random_byte_array(int n, uint32_t seed, uint8_t *buf,
+    std::vector<ByteArray>* out) {
+  std::mt19937 gen(seed);
+  std::uniform_int_distribution<int> d1(1, 12);
```
wesm: Make the max length configurable — also 0-length should be a valid option

wesm:
I'll finish reviewing a bit later — if we can figure out a way to enhance code reuse in the unit tests that would be great; I haven't looked close enough to see how hard that would be.

src/parquet/column/reader.h
majetideepak: Another alternative is to keep a `vector<uint8_t>` and copy the DataPages to this. We can resize the vector based on the `uncompressed_page_size`.

src/parquet/encodings/plain-encoding-test.cc
majetideepak: I can template INT32, INT64 and FLOAT, DOUBLE.

majetideepak:
Thanks! let me know if you can come with a better strategy to handle buffers for ByteArray and FLBA.

wesm:
I don't think we can (or should) easily extend the lifetime of the pointers inside the `ByteArray` and `FLBA` instances in this patch without copying the deserialized data page (and then managing that copy's lifetime). For the moment we should briefly document that the returned values become invalid when the column reader is garbage collected (which at the moment is...never, or at least not while the `ParquetFileReader` is alive). I only see these values as temporary objects in flight to some other persistent data structure (like Arrow http://arrow.apache.org/).

I am going to overhaul our memory management (which is currently quite poor) in general as part of PARQUET-520 (and follow-up JIRAs) and I suggest we revisit this issue after this refactoring takes place.

src/parquet/column/reader.h
wesm: per the top level comment let's remove this and add a TODO to `plain-encoding.h` around BA/FLBA that the returned objects do not currently own their data (and thus have lifetime tied to the input stream). Fixing this in a clean way that can yield good performance is too far out of scope for this JIRA

majetideepak: I did not look into the lifetime of these DataPages carefully. The above vector is actually redundant.

majetideepak:
@wesm Your comments are addressed. 
As part of PARQUET-520, are you planning to add a `MemoryPool` like class? The idea is to allocate from an application provided buffer.

wesm:
On PARQUET-520, I'll let you know when I dig in. The key thing there is being able to switch between memory maps and regular file (e.g. `fread`) into newly allocated memory.

I still think we can do better on the templating and code reuse. 

Here is the diff between the template and Int96:

https://www.diffchecker.com/sq0lb8i2

And ByteArray:

https://www.diffchecker.com/my1hmm9b

With this code being nearly identical (except the data generation and equality-comparison), we can put the differences in a templated struct with static methods but otherwise use the same code

* generate data
* encode
* decode
* compare results

src/parquet/encodings/plain-encoding-test.cc
```diff
@@ -63,10 +63,192 @@ TEST(BooleanTest, TestEncodeDecode) {
   ASSERT_EQ(nvalues, values_decoded);
 
   for (size_t i = 0; i < nvalues; ++i) {
-    ASSERT_EQ(BitUtil::GetArrayBit(decode_data, i), draws[i]) << i;
+    ASSERT_EQ(draws[i], BitUtil::GetArrayBit(decode_data, i)) << i;
   }
 }
 
-} // namespace test
+TEST(BooleanTest2, TestEncodeDecode) {
```
wesm: I also wonder whether the code duplication here can be reduced

majetideepak:
@wesm does this look good now ?

wesm:
Looks good, thank you. +1 LGTM

majetideepak:
@julienledem This is can be merged. Thank you!

PR #53, opened 2016-02-18
add googletest information to README

Mistobaan:


wesm:
We aren't supporting any version of googletest for development outside the one that is built by the thirdparty scripts, so documenting a different process creates extra work for the project. Let me think about how to improve the README.

Mistobaan:
Using cmake will fail if googletest is not found. One solution is to not require the gtest on the default target and add another `test` target that requires the googletest libraries.

I solved using the provided development script `setup_development_script.sh`.
I am on Mac.

wesm:
I'm working on a patch right now for PARQUET-471 with some better instructions about the thirdparty stuff, stay tuned

PR #55, opened 2016-02-18
PARQUET-468: Use thirdparty Thrift compiler to compile parquet.thrift at make time

wesm:
This also adds a `#define` working around the googletest incompatibility described in PARQUET-470.


wesm:
@asandryh @majetideepak can you take a look? This can be merged after PARQUET-471

wesm:
rebased

majetideepak:
+1 LGTM

majetideepak:
Were you able to verify if the fix for Thrift 0.9.3 works with this patch ?

wesm:
I was able to build and run the test suite locally on Thrift 0.9.3. We could upgrade the dev sandbox to use 0.9.3 or leave it at 0.9.1, I don't have a strong opinion

majetideepak:
Cool! Let's leave it at 0.9.1 for the first release. We can upgrade later.

wesm:
thank you!

PR #56, opened 2016-02-18
PARQUET-516: Better error handling in LocalFileSource

asandryh:
Also, a fix for PARQUET-537: Ensure LocalFileSource is closed properly.


src/parquet/file/reader-internal.cc
```diff
@@ -217,6 +217,10 @@ std::unique_ptr<ParquetFileReader::Contents> SerializedFile::Open(
   return result;
 }
 
+SerializedFile::~SerializedFile() {
+  Close();
+}
+
```
asandryh: This addresses PARQUET-537

src/parquet/file/reader.cc
```diff
@@ -74,13 +74,16 @@ RowGroupStatistics RowGroupReader::GetColumnStats(int i) const {
 // ParquetFileReader public API
 
 ParquetFileReader::ParquetFileReader() : schema_(nullptr) {}
-ParquetFileReader::~ParquetFileReader() {}
+ParquetFileReader::~ParquetFileReader() {
+  Close();
+}
```
asandryh: This addresses PARQUET-537

asandryh:
@wesm Can you look at this PR and let me know whether this is what you had in mind for PARQUET-516? Thanks.

wesm:
Yes, but we need to write unit tests to verify the correct functioning of these classes and their methods.

asandryh:
@wesm Right, will do. Just wanted to check first whether you had anything else in mind.

wesm:
In the tests the main thing will be failing in the expected ways (e.g. exceptions raised instead of core dump / segfault later).

wesm:
I'm working on a patch for PARQUET-520 and will add unit tests for the rest of input.h/cc.

asandryh:
@wesm I've updated this PR with unit tests. Please take a look.

src/parquet/util/input.cc
```diff
@@ -54,23 +58,47 @@ void LocalFileSource::Close() {
...
   return size_;
 }
 
+void LocalFileSource::SeekFile(int64_t pos, int origin) const {
+  if (origin == SEEK_SET && pos >= size_) {
+    std::stringstream ss;
+    ss << "Cannot seek to position " << pos << " in file [" << path_
+       << "] of size " << size_;
+    throw ParquetException(ss.str());
+  }
```
asandryh: Some implementations of fseek() will not produce an error if you try to seek past the end of the file. Hence, we should track this explicitly.

wesm:
Argh, we have crossed streams.

wesm:
Can you create a separate patch for PARQUET-537? I will incorporate the additional unit tests here into my patch-in-progress (which also includes a MemoryMapSource).

wesm:
Please see #66 and let me know if I missed any test cases here

PR #58, opened 2016-02-19
PARQUET-457: Verify page deserialization for GZIP and SNAPPY codecs, related refactoring

wesm:
This also restores passing on user's `CMAKE_CXX_FLAGS`, which had unfortunately led some compiler warnings to creep into our build.

PR #59, opened 2016-02-20
PARQUET-533: Add a Buffer abstraction, refactor input/output classes to be simpler using Buffers

wesm:
I have also removed all RowGroupReader and ColumnReader caching until we have an idea of how the caching will help users while also keeping memory-use under control.

The goal with this patch is to encapsulate the "data pointer and size" concept and lightly abstract away buffer ownership. The particular motivation is being able to deal with both normal files (bytes arriving via `fread`) and memory-mapped files (copying of bytes into new memory not required). It also helps do away with a bunch of functions that write output into a `std::vector<uint8_t>` now in favor of the `OwnedMutableBuffer`. Feedback welcome; plenty more work that can be done here. 

Requires PARQUET-457. Will rebase once that's merged.

### DIFF
--- a/src/parquet/encodings/plain-encoding-test.cc
+++ b/src/parquet/encodings/plain-encoding-test.cc
@@ -28,6 +28,7 @@
 #include "parquet/types.h"
 #include "parquet/schema/types.h"
 #include "parquet/util/bit-util.h"
+#include "parquet/util/buffer.h"
 #include "parquet/util/output.h"
 #include "parquet/util/test-common.h"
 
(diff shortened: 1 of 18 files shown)


majetideepak:
Finally, it will be useful to add a MemoryPool class as well. We should be able to allocate from a application provided buffer. The `locality` advantage from this feature will be significant.

wesm:
Can you open a JIRA about having an externalized memory allocator API? Out of scope for this JIRA but I have also been thinking about that.

majetideepak:
PARQUET-542 to track the external memory manager

wesm:
Great thanks. I would prefer to do some design work on that before anyone starts on a patch. I will mention this on JIRA

wesm:
Let me know any comments on this patch in particular such as it is, can be merged right after PARQUET-457 (and then it's required for #60)

src/parquet/encodings/plain-encoding-test.cc
```diff
@@ -141,7 +144,7 @@ void EncodeDecode<ByteArray, Type::BYTE_ARRAY>::generate_data() {
   int max_byte_array_len = 12 + sizeof(uint32_t);
   size_t nbytes = num_values_ * max_byte_array_len;
   data_buffer_.resize(nbytes);
-  random_byte_array(num_values_, 0.5, data_buffer_.data(), draws_,
+  random_byte_array(num_values_, 0, data_buffer_.data(), draws_,
```
majetideepak: Thanks! for changing this. This was throwing warnings in the travis ci build for clang. Should we enable `-Werror`  ?

majetideepak:
+1 LGTM

src/parquet/encodings/plain-encoding-test.cc
wesm: We are, but it was being lost in the CMakeLists.txt. I fixed it in PARQUET-457 https://github.com/apache/parquet-cpp/pull/58/files#diff-af3b638bc2a3e6c650974192a53c7291R238

julienledem:
+1
needs to be rebased

wesm:
Rebased

wesm:
thank you!

PR #60, opened 2016-02-20
PARQUET-525: Add test coverage for failure modes in ParseMetaData

wesm:


wesm:
I asked ASF INFRA to disable build failure when the coverage decreases

majetideepak:
This has patch  #58 as well ?

majetideepak:
I see, you are just rebasing off that.

wesm:
Yeah, if you anticipate any conflicts it's better to base patches off each other rather than all off master

src/parquet/file/reader-internal.cc
```diff
@@ -254,11 +248,8 @@ void SerializedFile::ParseMetaData() {
   uint8_t footer_buffer[FOOTER_SIZE];
   source_->Seek(filesize - FOOTER_SIZE);
   size_t bytes_read = source_->Read(FOOTER_SIZE, footer_buffer);
-
-  if (bytes_read != FOOTER_SIZE) {
-    throw ParquetException("Invalid parquet file. Corrupt footer.");
-  }
-  if (memcmp(footer_buffer + 4, PARQUET_MAGIC, 4) != 0) {
+  if (bytes_read != FOOTER_SIZE ||
```
majetideepak: do we need the `bytes_read ! = FOOTER_SIZE` check ?

wesm: Hypothetically, since `Read` is a virtual function, it could return fewer than `FOOTER_SIZE` bytes (or, if for some odd reason `fread` failed). It's a pretty pathological issue, though

src/parquet/file/reader-internal.cc
```diff
@@ -268,15 +259,10 @@ void SerializedFile::ParseMetaData() {
...
-
   source_->Seek(metadata_start);
 
   std::vector<uint8_t> metadata_buffer(metadata_len);
-  bytes_read = source_->Read(metadata_len, &metadata_buffer[0]);
-  if (bytes_read != metadata_len) {
-    throw ParquetException("Invalid parquet file. Could not read metadata bytes.");
-  }
-
+  source_->Read(metadata_len, &metadata_buffer[0]);
```
majetideepak: We have to keep the check here then ? since `bytes_read` could hypothetically be less than `metadata_len`

wesm: I just changed the unit test to definitely read past the end of the file here and will restore the check (the exception is being thrown in `DeserializeThriftMsg`)

wesm:
Coveralls flaked again. Let me know when it looks good and we can try to get this stack of PRs merged

majetideepak:
+1 LGTM

julienledem:
+1
duuuuude. You decreased the test coverage by 0.1%
needs to be rebased

wesm:
@julienledem rebased and ready to go

wesm:
thank you!

PR #62, opened 2016-02-22
PARQUET-538: Improve ColumnReader Tests

majetideepak:
closes #43 and closes #50
This PR also implements
1) PARQUET-532: Null values detection needs to be fixed and tested
2) PARQUET-502: Scanner segfaults when its batch size is smaller than the number of rows
3) PARQUET-526: Add more complete unit test coverage for column Scanner implementations
4) PARQUET-531: Can't read past first page in a column

### DIFF
--- a/src/parquet/column/column-reader-test.cc
+++ b/src/parquet/column/column-reader-test.cc
@@ -99,8 +99,8 @@ class TestPrimitiveReader : public ::testing::Test {
     int32_t batch_size = 8;
     size_t batch = 0;
     // This will cover both the cases
-    // 1) batch_size < page_size (multiple ReadBatch form a single page)
-    // 2) batch_size > page_size (read from multiple pages)
+    // 1) batch_size < page_size (multiple ReadBatch from a single page)
+    // 2) batch_size > page_size (BatchRead from multiple pages)
     do {
       batch = reader->ReadBatch(batch_size, &dresult[0] + batch_actual,
           &rresult[0] + batch_actual, &vresult[0] + total_values_read, &values_read);
--- a/src/parquet/column/column-reader-test.cc
+++ b/src/parquet/column/column-reader-test.cc
@@ -76,7 +76,7 @@ class TestPrimitiveReader : public ::testing::Test {
     values_.resize(num_values_);
     random_numbers(num_values_, seed, std::numeric_limits<int32_t>::min(),
        std::numeric_limits<int32_t>::max(), values_.data());
-    RandomPaginate<Type::INT32, int32_t>(d, values_, def_levels_, max_def_level_,
+    Paginate<Type::INT32, int32_t>(d, values_, def_levels_, max_def_level_,
         rep_levels_, max_rep_level_, levels_per_page, values_per_page, pages_);
   }
 
--- a/src/parquet/column/column-reader-test.cc
+++ b/src/parquet/column/column-reader-test.cc
@@ -44,29 +44,45 @@ namespace test {
 
 class TestPrimitiveReader : public ::testing::Test {
  public:
-  void init_def_levels() {
+  void make_pages() {
     num_values_ = 0;
-    def_levels_.resize(num_levels_);
-    random_levels(num_levels_, 0, max_def_level_, def_levels_.data());
-    for (int i = 0; i < num_levels_; i++) {
-      if (def_levels_[i] == max_def_level_) num_values_++;
+    int levels_per_page = num_levels_ / num_pages_;
+    vector<int16_t> defs;
+    vector<int16_t> reps;
+    vector<int32_t> vals;
+    int values_per_page = 0;
+    for (int p = 0; p < num_pages_; p++) {
+      values_per_page = 0;
+      // Create definition levels
+      if (max_def_level_ > 0) {
+        defs.resize(levels_per_page);
+        random_levels(levels_per_page, 0, max_def_level_, defs.data());
+        for (int i = 0; i < levels_per_page; i++) {
+          if (defs[i] == max_def_level_) values_per_page++;
+        }
+        def_levels_.insert(def_levels_.end(), defs.begin(), defs.end());
+      } else {
+        values_per_page = levels_per_page;
+      }
+      // Create repitition levels
+      if (max_rep_level_ > 0) {
+        reps.resize(levels_per_page);
+        random_levels(levels_per_page, 0, max_rep_level_, reps.data());
+        rep_levels_.insert(rep_levels_.end(), reps.begin(), reps.end());
+      }
+      // Create values
+      vals.resize(values_per_page);
+      random_numbers(values_per_page, 0, vals.data());
+      values_.insert(values_.end(), vals.begin(), vals.end());
+      std::shared_ptr<DataPage> page = MakeDataPage<Type::INT32>(vals, defs,
+          max_def_level_, reps, max_rep_level_);
+      pages_.push_back(page);
+
+      num_values_ += values_per_page;
     }
   }
 
-  void init_rep_levels() {
-    rep_levels_.resize(num_levels_);
-    random_levels(num_levels_, 0, max_rep_level_, rep_levels_.data());
-  }
-
-  void init_values() {
-    values_.resize(num_values_);
-    random_numbers(num_values_, 0, values_.data());
-  }
-
   void init_reader() {
-    std::shared_ptr<DataPage> page = MakeDataPage<Type::INT32>(values_, def_levels_,
-        max_def_level_, rep_levels_, max_rep_level_);
-    pages_.push_back(page);
     pager_.reset(new test::MockPageReader(pages_));
     reader_ = ColumnReader::Make(descr_, std::move(pager_));
   }
@@ -77,19 +93,24 @@ class TestPrimitiveReader : public ::testing::Test {
     vector<int16_t> rresult(num_levels_, -1);
     size_t values_read = 0;
     size_t total_values_read = 0;
+    size_t batch_actual = 0;
 
     Int32Reader* reader = static_cast<Int32Reader*>(reader_.get());
-    int offset = num_levels_ / 2;
-    size_t batch_actual = reader->ReadBatch(num_levels_, &dresult[0], &rresult[0],
-        &vresult[0], &values_read);
-    total_values_read += values_read;
-    batch_actual += reader->ReadBatch(offset, &dresult[0] + offset, &rresult[0] + offset,
-        &vresult[0] + values_read, &values_read);
-    total_values_read += values_read;
+    int32_t batch_size = 8;
+    size_t batch = 0;
+    // This will cover both the cases
+    // 1) batch_size < page_size (multiple ReadBatch form a single page)
+    // 2) batch_size > page_size (read from multiple pages)
+    do {
+      batch = reader->ReadBatch(batch_size, &dresult[0] + batch_actual,
+          &rresult[0] + batch_actual, &vresult[0] + total_values_read, &values_read);
+      total_values_read += values_read;
+      batch_actual += batch;
+      batch_size = batch_size * 2;
+    } while (batch > 0);
 
     ASSERT_EQ(num_levels_, batch_actual);
     ASSERT_EQ(num_values_, total_values_read);
-
     ASSERT_TRUE(vector_equal(values_, vresult));
     if (max_def_level_ > 0) {
       ASSERT_TRUE(vector_equal(def_levels_, dresult));
@@ -100,7 +121,7 @@ class TestPrimitiveReader : public ::testing::Test {
   }
 
   void execute() {
-    init_values();
+    make_pages();
     init_reader();
     check_results();
   }
@@ -108,6 +129,7 @@ class TestPrimitiveReader : public ::testing::Test {
  protected:
   int num_levels_;
   int num_values_;
+  int num_pages_;
   int16_t max_def_level_;
   int16_t max_rep_level_;
   const ColumnDescriptor *descr_;
@@ -120,7 +142,8 @@ class TestPrimitiveReader : public ::testing::Test {
 };
 
 TEST_F(TestPrimitiveReader, TestInt32FlatRequired) {
-  num_levels_ = num_values_ = 1000;
+  num_levels_ = 5000;
+  num_pages_ = 50;
   max_def_level_ = 0;
   max_rep_level_ = 0;
   NodePtr type = schema::Int32("a", Repetition::REQUIRED);
@@ -130,25 +153,24 @@ TEST_F(TestPrimitiveReader, TestInt32FlatRequired) {
 }
 
 TEST_F(TestPrimitiveReader, TestInt32FlatOptional) {
-  num_levels_ = 1000;
+  num_levels_ = 5000;
+  num_pages_ = 50;
   max_def_level_ = 4;
   max_rep_level_ = 0;
   NodePtr type = schema::Int32("b", Repetition::OPTIONAL);
   const ColumnDescriptor descr(type, max_def_level_, max_rep_level_);
   descr_ = &descr;
-  init_def_levels();
   execute();
 }
 
 TEST_F(TestPrimitiveReader, TestInt32FlatRepeated) {
-  num_levels_ = 1000;
+  num_levels_ = 5000;
+  num_pages_ = 50;
   max_def_level_ = 4;
   max_rep_level_ = 2;
   NodePtr type = schema::Int32("c", Repetition::REPEATED);
   const ColumnDescriptor descr(type, max_def_level_, max_rep_level_);
   descr_ = &descr;
-  init_def_levels();
-  init_rep_levels();
   execute();
 }
 
--- a/src/parquet/column/column-reader-test.cc
+++ b/src/parquet/column/column-reader-test.cc
@@ -80,7 +80,7 @@ class TestPrimitiveReader : public ::testing::Test {
         rep_levels_, max_rep_level_, levels_per_page, values_per_page, pages_);
   }
 
-  void InitReader(const ColumnDescriptor *d) {
+  void InitReader(const ColumnDescriptor* d) {
     std::unique_ptr<PageReader> pager_;
     pager_.reset(new test::MockPageReader(pages_));
     reader_ = ColumnReader::Make(d, std::move(pager_));
--- a/src/parquet/column/column-reader-test.cc
+++ b/src/parquet/column/column-reader-test.cc
@@ -44,195 +44,113 @@ namespace test {
 
 class TestPrimitiveReader : public ::testing::Test {
  public:
-  void SetUp() {}
+  void init_def_levels() {
+    num_values_ = 0;
+    def_levels_.resize(num_levels_);
+    random_levels(num_levels_, 0, max_def_level_, def_levels_.data());
+    for (int i = 0; i < num_levels_; i++) {
+      if (def_levels_[i] == max_def_level_) num_values_++;
+    }
+  }
 
-  void TearDown() {}
+  void init_rep_levels() {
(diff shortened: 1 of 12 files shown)


majetideepak:
@wesm feedback please!

src/parquet/column/column-reader-test.cc
```diff
@@ -44,195 +44,135 @@ namespace test {
...
+      std::shared_ptr<DataPage> page = MakeDataPage<Type::INT32>(vals, defs,
+          max_def_level_, reps, max_rep_level_);
+      pages_.push_back(page);
+
+      num_values_ += values_per_page;
+    }
+  }
 
-  void InitReader(const ColumnDescriptor* descr) {
+  void init_reader() {
```
wesm: Google C++ style function names, please

src/parquet/column/column-reader-test.cc
```diff
@@ -44,195 +44,135 @@ namespace test {
...
+      }
+      // Create values
+      vals.resize(values_per_page);
+      random_numbers(values_per_page, 0, vals.data());
+      values_.insert(values_.end(), vals.begin(), vals.end());
+      std::shared_ptr<DataPage> page = MakeDataPage<Type::INT32>(vals, defs,
+          max_def_level_, reps, max_rep_level_);
+      pages_.push_back(page);
+
+      num_values_ += values_per_page;
```
wesm: This function is combining two tasks which ought to be separate for better code reuse

- Generating random data
- "Paginating" data.

I'd prefer to see a function that accepts pre-generated arrays of data, along with the max definition level and a page size, and returns a vector of pages, and not as part of a test fixture.

wesm: For example, if you have in essence 10,000 data points, you could pass `page_size = 1000` and get back a vector of 10 pages.

src/parquet/column/column-reader-test.cc
```diff
@@ -44,195 +44,135 @@ namespace test {
...
+    // This will cover both the cases
+    // 1) batch_size < page_size (multiple ReadBatch from a single page)
+    // 2) batch_size > page_size (BatchRead from multiple pages)
+    do {
+      batch = reader->ReadBatch(batch_size, &dresult[0] + batch_actual,
+          &rresult[0] + batch_actual, &vresult[0] + total_values_read, &values_read);
+      total_values_read += values_read;
+      batch_actual += batch;
+      batch_size = batch_size * 2;
+    } while (batch > 0);
```
wesm: Can you preserve the final `ReadBatch` after this with all `nullptr` arguments? This way valgrind will catch improper writes when we are at EOS. 

```c++
batch_actual = reader->ReadBatch(5, nullptr, nullptr, nullptr, &values_read);
ASSERT_EQ(0, batch_actual);
ASSERT_EQ(0, values_read);
```

src/parquet/column/column-reader-test.cc
wesm: (For example, we can use this paginate function for benchmarking)

src/parquet/column/column-reader-test.cc
```diff
@@ -44,195 +44,135 @@ namespace test {
...
-
-  size_t values_read = 0;
-  size_t batch_actual = reader->ReadBatch(10, nullptr, nullptr,
-      &result[0], &values_read);
-  ASSERT_EQ(10, batch_actual);
-  ASSERT_EQ(10, values_read);
-
-  ASSERT_TRUE(vector_equal(result, values));
+  const ColumnDescriptor descr(type, max_def_level_, max_rep_level_);
+  descr_ = &descr;
```
wesm: This anti-pattern suggests we should be passing `std::shared_ptr<ColumnDescriptor>` to `ColumnReader::Make`. What do you think?

src/parquet/reader-test.cc
```diff
@@ -111,7 +110,6 @@ TEST_F(TestAllTypesPlain, TestSetScannerBatchSize) {
   ASSERT_EQ(1024, scanner->batch_size());
 }
 
-
```
wesm: can you unstage this whitespace diff, unless it's causing `make lint` to fail?

src/parquet/util/test-common.h
```diff
@@ -183,6 +183,15 @@ void random_byte_array(int n, uint32_t seed, uint8_t *buf,
...
   }
 }
+
+void random_levels(int n, uint32_t seed, int16_t max_level, int16_t* out) {
+  std::mt19937 gen(seed);
+  std::uniform_int_distribution<uint16_t> d(0, max_level);
+  for (int i = 0; i < n; ++i) {
+    out[i] = d(gen);
+  }
+}
```
wesm: Perhaps we can refactor `random_numbers` to use another function that accepts upper and lower bounds for the type `T`? Looking again at the data generation in `test-common.h` the code duplication here shows that things are factored suboptimally:

```c++
template <>
void random_numbers(int n, uint32_t seed, int32_t* out) {
  std::mt19937 gen(seed);
  std::uniform_int_distribution<int32_t> d(std::numeric_limits<int32_t>::lowest(),
      std::numeric_limits<int32_t>::max());
  for (int i = 0; i < n; ++i) {
    out[i] = d(gen);
  }
}

template <>
void random_numbers(int n, uint32_t seed, int64_t* out) {
  std::mt19937 gen(seed);
  std::uniform_int_distribution<int64_t> d(std::numeric_limits<int64_t>::lowest(),
      std::numeric_limits<int64_t>::max());
  for (int i = 0; i < n; ++i) {
    out[i] = d(gen);
  }
}
```

I propose:

```c++
template <typename T>
void random_int(int n, uint32_t seed, T lower_bound, T upper_bound, T* out) {
  std::mt19937 gen(seed);
  std::uniform_int_distribution<T> d(lower_bound, upper_bound);
  for (int i = 0; i < n; ++i) {
    out[i] = d(gen);
  }
}
```

Then you can use this for level generation, too.

src/parquet/column/column-reader-test.cc
wesm: In general, to the extent that you can make functions like these less dependent on member variables (and preferably not embedded in the test fixture at all), the better.

src/parquet/column/reader.h
```diff
@@ -155,50 +155,59 @@ inline size_t TypedColumnReader<TYPE>::ReadValues(size_t batch_size, T* out) {
...
-    // TODO(wesm): this tallying of values-to-decode can be performed with better
-    // cache-efficiency if fused with the level decoding.
-    for (size_t i = 0; i < num_def_levels; ++i) {
-      if (def_levels[i] == descr_->max_definition_level()) {
-        ++values_to_read;
+  *values_read = 0;
+  size_t total_values = 0;
+  int32_t level_offset = 0;
+  int32_t value_offset = 0;
+  size_t current_values_read = 0;
```
wesm: use signed integers for all of these

wesm: I think `int` is fine

src/parquet/column/reader.h
```diff
@@ -155,50 +155,59 @@ inline size_t TypedColumnReader<TYPE>::ReadValues(size_t batch_size, T* out) {
...
+      }
     }
-  }
-
-  *values_read = ReadValues(values_to_read, values);
-  size_t total_values = std::max(num_def_levels, *values_read);
-  num_decoded_values_ += total_values;
 
+    current_values_read = ReadValues(values_to_read, values + value_offset);
+    size_t current_values = std::max(num_def_levels, current_values_read);
```
wesm: If there are definition levels, `num_def_levels` is always greater than `current_values_read`. I would suggest setting this variable right after reading the definition levels.

wesm:
This is looking a lot better overall, thanks for doing the cleanup. how do you want to handle reconciling PARQUET-526 with this? You could combine the patches, that would be fine with me.

src/parquet/column/column-reader-test.cc
```diff
@@ -44,195 +44,135 @@ namespace test {
...
+
+    Int32Reader* reader = static_cast<Int32Reader*>(reader_.get());
+    int32_t batch_size = 8;
+    size_t batch = 0;
+    // This will cover both the cases
+    // 1) batch_size < page_size (multiple ReadBatch from a single page)
+    // 2) batch_size > page_size (BatchRead from multiple pages)
+    do {
+      batch = reader->ReadBatch(batch_size, &dresult[0] + batch_actual,
+          &rresult[0] + batch_actual, &vresult[0] + total_values_read, &values_read);
```
wesm: You need to check somehow that this continues reading pages when a current page is exhausted. But, in thinking more about this, what happens to `ByteArray` and `FixedLenByteArray` values when a batch read crosses a page boundary? I'm slightly rethinking whether a batch read should span pages. Let me know what you think

majetideepak:
I will make the above changes. I am planning to combine PARQUET-526 on top of this. I have a question on `TYPED_TEST`. Is there an alternative for tests templated on values ?

src/parquet/column/column-reader-test.cc
majetideepak: Since the lifetime of the `DataPages` is until the `RowGroupReader` get destroyed (if I understood it correctly), will BA and FBLA cause problems ?

src/parquet/column/reader.h
majetideepak: This works in the case when definition levels are absent too. Did you mean to have two assignments ? one after reading values and one after reading definition levels?

src/parquet/column/reader.h
majetideepak: The parquet types use `int32_t`. I can use `int` too. I don't have a strong opinion.

wesm: Either way is fine.

src/parquet/util/test-common.h
majetideepak: changing it to `random_int` breaks the templates in the tests. Ideally we have to combine `real` and `int` generators into a single `random_numbers` template where `std::uniform_int_distribution` is called for `int16_t` `int32_t` `int64_t`and `std::uniform_real_distribution` is called for `float` and `double`.

src/parquet/column/reader.h
wesm: Read through this again. I think it's okay as is. I'm more concerned about the page boundary question

src/parquet/column/column-reader-test.cc
wesm: I think it's only an issue for compressed data pages (because of the `decompression_buffer_`). I think until we figure out a more flexible way to manage decompressed data lifetime we should return the batch when we reach the end of the data page, and document that users should keep calling `ReadBatch` until it returns 0

wesm:
`TYPED_TEST` is one way to address the copypasta problem; I am fine with any option that results in maximum code reuse.

src/parquet/column/column-reader-test.cc
majetideepak: Yeah! compressed data will cause additional buffer management problems here. Let's postpone reading multiple data pages as you suggested.

src/parquet/util/test-common.h
wesm: What about refactoring `random_number` to call `random_int` except for float and double? 

```
template <>
void random_number(int n, uint32_t seed, double* out) {
  random_real(n, seed, -INF, INF, out);
}
```

majetideepak: Yes! I am specializing float and double.

src/parquet/column/column-reader-test.cc
majetideepak: "Paginating" gets complicated when you have OPTIONAL and REPEATED data. We will have to pass the number of values per page too.

src/parquet/column/column-reader-test.cc
majetideepak: I agree.

src/parquet/column/column-reader-test.cc
wesm: You can examine the definition levels (using the max definition level) to compute the number of values in each page.

wesm:
What's your gameplan with PARQUET-526 -- are you going to combine the patches or rebase PARQUET-526 on this (and open a new pull request)?

majetideepak:
I am planning to combine patches on top of this. I could re-use the Pageinate and extend the column-`reader-test` to `scanner-test`. If any API changes are needed, I can change for both.

wesm:
Very good, thanks

majetideepak:
@wesm this is pretty much complete. I just have to add a test for boolean scanner.  `vector<bool>` is implemented differently and it breaks our API when we try to get the underlying memory. I am planning to specialize a test.

wesm:
Cool, I will review in a little while.

src/parquet/column/column-reader-test.cc
```diff
@@ -44,195 +44,132 @@ namespace test {
...
+    // Create values
+    values_.resize(num_values_);
+    random_numbers(num_values_, seed, std::numeric_limits<int32_t>::min(),
+       std::numeric_limits<int32_t>::max(), values_.data());
+    Paginate<Type::INT32, int32_t>(d, values_, def_levels_, max_def_level_,
+        rep_levels_, max_rep_level_, levels_per_page, values_per_page, pages_);
+  }
 
-  void InitReader(const ColumnDescriptor* descr) {
+  void InitReader(const ColumnDescriptor *d) {
```
wesm: ColumnDescriptor* d

src/parquet/column/scanner-test.cc
```diff
@@ -0,0 +1,271 @@
...
+}
+
+static int FLBA_LENGTH = 12;
+bool operator==(const FixedLenByteArray& a, const FixedLenByteArray& b) {
+  return 0 == memcmp(a.ptr, b.ptr, FLBA_LENGTH);
+}
+
+namespace test {
+
+template <int TYPE, typename T>
```
wesm: `typename T` adds extra complexity here because you can use `type_traits<TYPE>::value_type`

wesm: i.e. putting in the public: blog

```c++
typedef typename type_traits<TYPE>::value_type T;
```

src/parquet/column/scanner-test.cc
```diff
@@ -0,0 +1,271 @@
...
+  ExecuteAll(num_pages, num_levels_per_page, batch_size);
+}
+
+TEST_F(TestFlatByteArrayScanner, TestScanner) {
+  ExecuteAll(num_pages, num_levels_per_page, batch_size);
+}
+
+TEST_F(TestFlatFLBAScanner, TestScanner) {
+  ExecuteAll(num_pages, num_levels_per_page, batch_size);
+}
```
wesm: I really don't think `TYPED_TEST` is all that scary but I'll let this slide for now.

wesm: =)

src/parquet/column/test-util.h
```diff
@@ -191,6 +193,36 @@ static std::shared_ptr<DataPage> MakeDataPage(const std::vector<T>& values,
       page_builder.rep_level_encoding());
 }
 
+template <int TYPE, typename T>
```
wesm: same comment as above re: the `typename T` -- better to use `type_traits`

src/parquet/column/scanner-test.cc
```diff
@@ -0,0 +1,271 @@
...
+  void CheckResults(int batch_size) {
+    TypedScanner<TYPE>* scanner = reinterpret_cast<TypedScanner<TYPE>* >(scanner_.get());
+    T val;
+    bool is_null;
+    size_t j = 0;
+    scanner->SetBatchSize(batch_size);
+    for (size_t i = 0; i < num_levels_; i++) {
+      ASSERT_TRUE(scanner->NextValue(&val, &is_null)) << i <<"NV"<< j;
+      if (!is_null) {
+        ASSERT_EQ(values_[j++], val) << i <<"V"<< j;
```
wesm: We still aren't verifying that the repetition/definition levels scanned are correct per the discussion in #50 . 

I think the right way to handle this may be to instead change the scanner API to have a method:

```c++
bool Next(T* val, int16_t* rep_level, int16_t* def_level, bool* is_null)
```

what do you think?

src/parquet/column/scanner.h
```diff
@@ -108,21 +108,21 @@ class TypedScanner : public Scanner {
...
+      value_offset_ = 0;
       level_offset_ = 0;
       if (!levels_buffered_) {
         return false;
       }
     }
-    *def_level = def_levels_[level_offset_++];
-    *rep_level = 1;
+    *rep_level = descr()->is_repeated() ? rep_levels_[level_offset_] : 0;
+    *def_level = descr()->is_optional() ? def_levels_[level_offset_] : 1;
```
wesm: Isn't the definition level always zero if `max_definition_level == 0`? These results aren't being tested yet

wesm:
If we can fix the correctness issues with the repetition/definition levels using the new `Scanner::Next` API I think we are good to go. Let me know when that's done and I'll have a last look

src/parquet/column/scanner-test.cc
majetideepak: I will give it another shot. What types will the `Primitives` array contain ?

src/parquet/column/scanner-test.cc
majetideepak: Yes! I will make that change now.

src/parquet/column/scanner-test.cc
majetideepak: We already test the scanning of repetition and definition levels in the column-reader-test.
I see the `Scanner` as a higher level API which gives us `values` as its output. The checking of values and nulls should be sufficient here I feel.

src/parquet/column/scanner.h
majetideepak: Yea, you are right. I will fix this.
 I think this still passes because `def_level > max_definition_level()` and it becomes same as REQUIRED.

src/parquet/column/scanner-test.cc
wesm: It should be all of the Parquet primitive types

wesm: This is less urgent than the correctness issues, we can leave it for cleaning later

src/parquet/column/scanner-test.cc
majetideepak: Actually, its better we do a check here as well. I will implement this API. Sorry for the confusion.

majetideepak:
@wesm Does this look good ?

src/parquet/column/test-util.h
majetideepak: Forgot this. Doing now.

wesm:
yep this looks great -- +1 LGTM (when the build is green)

wesm:
If you put "closes #43 and closes #50" in the description it will tidy those PRs when this is merged

majetideepak:
Sure! I am just trying the `TYPED_TEST` :). Will try another 15 mins and push a patch for green build.

majetideepak:
@julienledem this is good to go on a green build. Thank you!
@wesm I got `TYPED_TEST` working.

majetideepak:
https://travis-ci.org/apache/parquet-cpp/jobs/111389933
failed with a strange CMake flag error.

wesm:
The apt-get commands failed -- I've seen this happen 1 or 2 other times so no worries

wesm:
See https://coveralls.io/builds/5182275/source?filename=src%2Fparquet%2Fcolumn%2Fscanner.h

as we're scrubbing the codebase for release I suggest we return and add test coverage for the missed lines

wesm:
follow up patch is fine

majetideepak:
I will add tests for missed lines in both the reader and the scanner.

nongli:
Merged into master. Thanks!

PR #63, opened 2016-02-22
PARQUET-518: Remove -Wno-sign-compare and scrub integer signedness

wesm:
This patch removes compiler warning suppresses, fixes signed-unsigned integer comparisons, and scrubs most usages of `size_t` from the codebase in favor of signed integer types.

### DIFF
--- a/src/parquet/column/test-util.h
+++ b/src/parquet/column/test-util.h
@@ -47,7 +47,7 @@ class MockPageReader : public PageReader {
 
   // Implement the PageReader interface
   virtual std::shared_ptr<Page> NextPage() {
-    if (page_index_ == pages_.size()) {
+    if (page_index_ == static_cast<int>(pages_.size())) {
       // EOS to consumer
       return std::shared_ptr<Page>(nullptr);
     }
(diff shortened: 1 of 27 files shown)


wesm:
I'll rebase and complete this after #64 is merged

wesm:
@majetideepak if you could review, thank you!

src/parquet/column/test-util.h
```diff
@@ -186,7 +186,7 @@ static std::shared_ptr<DataPage> MakeDataPage(const ColumnDescriptor *d,
     const std::vector<T>& values,
     const std::vector<int16_t>& def_levels, int16_t max_def_level,
     const std::vector<int16_t>& rep_levels, int16_t max_rep_level) {
-  size_t num_values = values.size();
+  int num_values = values.size();
```
majetideepak: This should be `size_t` ?

majetideepak:
+1 LGTM

src/parquet/column/test-util.h
wesm: It is okay to cast to int unless there is concern about an overflow -- data pages are limited to INT32_MAX anyway

wesm:
Rebased

wesm:
Travis CI appears to be having problem again this morning, I just pinged them about it

wesm:
green build: https://travis-ci.org/wesm/parquet-cpp/builds/112679440

julienledem:
+1

PR #64, opened 2016-02-25
PARQUET-494: Implement DictionaryEncoder and test dictionary decoding

wesm:
I incorporated quite a bit of code from Impala for this patch, but did a bunch of work to get everything working. In particular, I wasn't happy with the hash table implementation in `dict-encoder.h` and so have written a simple new one that we can benchmark and tune as necessary. 

The simplest way to pull in the DictEncoder (PARQUET-493) was to also bring in the `MemPool` implementation, suitably trimmed down. We can continue to refactor this as needed for parquet-cpp.

I also did some light refactoring using `TYPED_TEST` in `plain-encoding-test` (now `encoding-test`).

PR #65, opened 2016-02-25
PARQUET-545: Improve API to support decimal type

majetideepak:
This PR also
1) Improves scanner coverage.
2) Implements PARQUET-546
3) Adds tests for Types to string


majetideepak:
@wesm I don't see much to do for `DECIMAL` types except to check that the `scale` and `precision` are set properly in the `ColumnDescriptor`. I modified the `ColumnDescriptor` and I am yet to add some test cases.
Do you have other thoughts here?
I also implemented PARQUET-546. Can you have a look ? I will add test cases if you are okay with the new API.
Thanks!

src/parquet/encodings/plain-encoding-test.cc
```diff
@@ -221,8 +221,8 @@ TEST(BAEncodeDecode, TestEncodeDecode) {
 
 TEST(FLBAEncodeDecode, TestEncodeDecode) {
   schema::NodePtr node;
-  node = schema::PrimitiveNode::MakeFLBA("name", Repetition::OPTIONAL,
-      flba_length, LogicalType::UTF8);
```
majetideepak: We should have a single API to create Primitive nodes.

majetideepak:
I just looked at  `parquet-mr` and it looks like they check if the `length` `precision` and `scale` validate the decimal type. I will add this check.

wesm:
I commented on https://issues.apache.org/jira/browse/PARQUET-545. Impala has a lot of decimal-related code, but it is mostly used for computations in the runtime. I'm not sure if having container types for decimal data (and stuff like coercing to/from double -- that would be useful IMHO) would be helpful overall. Being able to print decimals would be nice. It's a can of worms though

majetideepak:
@wesm I agree. I just commented on  https://issues.apache.org/jira/browse/PARQUET-545 too. In this patch, I want to extend the ColumnDescriptor API to be able to extract the `scale` and `precision` values required to interpret decimals.

src/parquet/schema/descriptor.h
```diff
@@ -54,24 +53,32 @@ class ColumnDescriptor {
...
+    return primitive_node_->logical_type();
+  }
+
   const std::string& name() const {
     return primitive_node_->name();
   }
 
   bool is_required() const {
-    return max_definition_level_ == 0;
+    return primitive_node_->is_required();
```
majetideepak: We should be using the Type values set in the PrimitiveNode.

src/parquet/column/scanner.h
```diff
@@ -114,7 +114,7 @@ class TypedScanner : public Scanner {
         return false;
       }
     }
-    *def_level = descr()->is_optional() ?
+    *def_level = !descr()->is_required() ?
```
majetideepak: This fix is important.

src/parquet/schema/descriptor.cc
```diff
@@ -84,10 +84,15 @@ const ColumnDescriptor* SchemaDescriptor::Column(size_t i) const {
...
 
+int ColumnDescriptor::type_scale() const {
+  return primitive_node_->decimal_metadata().scale;
+}
+
+int ColumnDescriptor::type_precision() const {
+  return primitive_node_->decimal_metadata().precision;
+}
+
 int ColumnDescriptor::type_length() const {
```
majetideepak: ColumnDescriptor API extensions.

majetideepak:
@wesm I feel this patch is important for the release as well. Can you please give feedback. I will rebase this off #64.

src/parquet/schema/descriptor.h
```diff
@@ -54,24 +53,32 @@ class ColumnDescriptor {
...
 
   bool is_optional() const {
-    return max_definition_level_ > 0;
+    return primitive_node_->is_optional();
   }
 
   bool is_repeated() const {
-    return max_repetition_level_ > 0;
+    return primitive_node_->is_repeated();
   }
```
wesm: These changes have fundamentally changed the nature of their results -- they can't be used like they were in the scanner any more, because you could have required types with definition level > 0:

```
optional group bag
  repeated group list
    required int32 item
```

I suggest using the max repetition/definition levels in the scanner to avoid this issue

src/parquet/column/scanner.h
```diff
@@ -45,7 +45,7 @@ class Scanner {
       values_buffered_(0),
       reader_(reader) {
     // TODO: don't allocate for required fields
-    def_levels_.resize(reader->descr()->is_optional() ? batch_size_ : 0);
+    def_levels_.resize(!reader->descr()->is_required() ? batch_size_ : 0);
```
wesm: If the `ColumnDescriptor` is REQUIRED inside a nested schema tree, it will have definition levels

src/parquet/column/scanner.h
wesm: See above

src/parquet/types.h
```diff
@@ -280,6 +280,73 @@ static inline std::string type_to_string(Type::type t) {
...
+      return "JSON";
+      break;
+    case LogicalType::BSON:
+      return "BSON";
+      break;
+    case LogicalType::INTERVAL:
+      return "INTERVAL";
+      break;
+    default:
+      return "UNKNOWN";
```
wesm: Test these?

src/parquet/schema/descriptor.h
majetideepak: makes sense. I did not think about the groups. Will revert this.

src/parquet/types.h
majetideepak: Will do.

src/parquet/schema/descriptor.h
wesm: Since this didn't cause test failures, it may be worth trying to add a failing test case

src/parquet/column/scanner.h
majetideepak: Here  `is_optional()` (`max_definition_level_ > 0`) and `! is_required()` (`!(max_definition_level_ == 0)`) are equivalent statements (given precondition that `max_definition_level_` is never less than 0). I want to use `! is_required()` for clarity. Technically, definition values exist for both `optional` and `repeated` values. Just checking for `optional` here is less clear.

src/parquet/schema/descriptor.h
majetideepak: We don't have any groups tests yet. This code works for all flat schemas.

majetideepak: I will add some `GroupNode` tests in the Scanner and ColumnReader

src/parquet/column/scanner.h
majetideepak: Infact, when I changed the implementation to return the `RepetitionType` in the `PrimitiveNode`, the previous code breaks even for flat schema.

src/parquet/schema/schema-descriptor-test.cc
```diff
@@ -73,6 +73,37 @@ TEST_F(TestSchemaDescriptor, InitNonGroup) {
   ASSERT_THROW(descr_.Init(node), ParquetException);
 }
 
+TEST_F(TestSchemaDescriptor, DescriptorRepetitionValues) {
```
majetideepak: Tests for repetition values of schemas with GroupNodes

majetideepak:
The changes are complete. I will wait for the test coverage report.

wesm:
Cool I can review in a little bit

src/parquet/column/scanner-test.cc
```diff
@@ -203,42 +207,65 @@ void TestFlatScanner<FLBAType>::InitValues() {
...
+  NodePtr type = schema::PrimitiveNode::Make("c1", Repetition::OPTIONAL,
+      Type::FIXED_LEN_BYTE_ARRAY, LogicalType::DECIMAL, FLBA_LENGTH, 10, 2);
+  const ColumnDescriptor d(type, 4, 0);
+  MakePages(&d, 1, 100);
+  InitScanner(&d);
+  TypedScanner<FLBAType::type_num>* scanner =
+    reinterpret_cast<TypedScanner<FLBAType::type_num>* >(scanner_.get());
+  ASSERT_EQ(10, scanner->descr()->type_precision());
+  ASSERT_EQ(2, scanner->descr()->type_scale());
+  ASSERT_EQ(FLBA_LENGTH, scanner->descr()->type_length());
```
majetideepak: Test for ColumnDescriptor API extensions in this patch.

src/parquet/column/scanner-test.cc
```diff
@@ -203,42 +207,65 @@ void TestFlatScanner<FLBAType>::InitValues() {
...
-      FLBA_LENGTH, LogicalType::UTF8);
+  NodePtr type = schema::PrimitiveNode::Make("c1", Repetition::REQUIRED,
+      Type::FIXED_LEN_BYTE_ARRAY, LogicalType::DECIMAL, FLBA_LENGTH, 10, 2);
   const ColumnDescriptor d(type, 0, 0);
   MakePages(&d, 1, 100);
   InitScanner(&d);
   CheckResults(1, &d);
 }
 
+TEST_F(TestFlatFLBAScanner, TestScannerCoverage) {
```
wesm: Can you add a comment explaining what this test case verifies? If it tests multiple things, can you break it up into multiple test cases with clear descriptions?

src/parquet/column/scanner-test.cc
wesm: These should probably go in their own test case.

src/parquet/column/scanner.h
```diff
@@ -114,7 +114,7 @@ class TypedScanner : public Scanner {
         return false;
       }
     }
-    *def_level = descr()->is_optional() ?
+    *def_level = !descr()->is_required() ?
       def_levels_[level_offset_] : descr()->max_definition_level();
```
wesm: In the false branch, the output should always be zero, using `max_definition_level()` obscures that fact. I think this code would be much clearer as

```c++
*def_level = descr->max_definition_level() > 0? def_levels_[level_offset_] : 0;
```

As is, it's confusing because the leaf node can actually *be* required, but implicitly null if any of its optional ancestors is null. If that makes sense? Same comment for `*rep_level` below

src/parquet/parquet-types-test.cc
```diff
@@ -0,0 +1,82 @@
+// Licensed to the Apache Software Foundation (ASF) under one
```
wesm: Can you call this file `types-test.cc`?

src/parquet/schema/schema-descriptor-test.cc
```diff
@@ -73,6 +73,37 @@ TEST_F(TestSchemaDescriptor, InitNonGroup) {
...
+
+  const ColumnDescriptor* col = descr_.Column(0);
+  EXPECT_TRUE(col->is_required());
+  col = descr_.Column(1);
+  EXPECT_TRUE(!col->is_required());
+  EXPECT_TRUE(col->is_repeated());
+  col = descr_.Column(2);
+  EXPECT_TRUE(!col->is_required());
+  EXPECT_TRUE(col->is_optional());
+  EXPECT_TRUE(col->is_repeated());
```
wesm: It would be better to put these tests in the `scanner-test.cc` to check that the scanner logic is correct for leaf nodes in nested schemas -- I believe these tests are already overlapping with the other schema tests from the earlier patch. 

Per the discussion above I'm still unsure about how useful these methods for the potential confusion they introduce.

src/parquet/schema/types.cc
```diff
@@ -40,6 +40,103 @@ bool Node::EqualsInternal(const Node* other) const {
...
+    case LogicalType::NONE:
+      // Logical type not set
+      // Clients should be able to read these values
+      decimal_metadata_.precision = precision;
+      decimal_metadata_.scale = scale;
+      break;
+    case LogicalType::UTF8:
+    case LogicalType::JSON:
+    case LogicalType::BSON:
+      if (!(type == Type::BYTE_ARRAY)) {
```
wesm: `!=`

wesm:
Looks like we're blocked on Travis builds again for the next few hours. Let me know on these relatively minor issues and I'll look again a bit later. I'm going to close out the remaining clean-up patches as quickly as I can

wesm:
Also, the PR title needs to be changed to `PARQUET-545`

src/parquet/column/scanner.h
majetideepak: I am fine with this suggestion :)

src/parquet/schema/schema-descriptor-test.cc
majetideepak: This won't be required with the above changes.

majetideepak:
I am making these changes now.

wesm:
+1, thank you!

majetideepak:
@wesm Thanks! 
On a separate note, will it help if we test the column reader and scanner for Dictionary pages too ?

wesm:
Yes, that would be a good idea -- I only tested the lowest tier of encoding. Can you open a JIRA?

wesm:
This issue title needs to start with `PARQUET-545:` (capital letters) or the merge tool will fail

majetideepak:
https://issues.apache.org/jira/browse/PARQUET-549

julienledem:
+1

PR #66, opened 2016-02-29
PARQUET-520: Add MemoryMapSource and add unit tests for both it and LocalFileSource

wesm:
I also added the `file_descriptor` API so that we can verify that dtors elsewhere successfully close open files. Closes #56

### DIFF
--- a/src/parquet/file/reader.cc
+++ b/src/parquet/file/reader.cc
@@ -68,7 +68,7 @@ ParquetFileReader::ParquetFileReader() : schema_(nullptr) {}
 ParquetFileReader::~ParquetFileReader() {}
 
 std::unique_ptr<ParquetFileReader> ParquetFileReader::OpenFile(const std::string& path) {
-  std::unique_ptr<LocalFileSource> file(new LocalFileSource());
+  std::unique_ptr<MemoryMapSource> file(new MemoryMapSource());
   file->Open(path);
 
   auto contents = SerializedFile::Open(std::move(file));
--- a/src/parquet/file/reader.cc
+++ b/src/parquet/file/reader.cc
@@ -67,8 +67,14 @@ RowGroupStatistics RowGroupReader::GetColumnStats(int i) const {
 ParquetFileReader::ParquetFileReader() : schema_(nullptr) {}
 ParquetFileReader::~ParquetFileReader() {}
 
-std::unique_ptr<ParquetFileReader> ParquetFileReader::OpenFile(const std::string& path) {
-  std::unique_ptr<MemoryMapSource> file(new MemoryMapSource());
+std::unique_ptr<ParquetFileReader> ParquetFileReader::OpenFile(const std::string& path,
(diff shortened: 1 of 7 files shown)


wesm:
@asandryh let me know if there's anything additional from #56 that I can add here; we should prefer this patch as it runs the same set of tests both on memory-mapped / non-memory-mapped file readers. We need a test case reproduction of PARQUET-537 before fixing that bug.

src/parquet/util/input-output-test.cc
```diff
@@ -43,4 +49,59 @@ TEST(TestInMemoryOutputStream, Basics) {
...
+
+TYPED_TEST(TestFileReaders, NonExistentFile) {
+  ASSERT_THROW(this->source.Open("0xDEADBEEF.txt"), ParquetException);
+}
+
+TYPED_TEST(TestFileReaders, FileDisappeared) {
+  this->source.Open(this->test_path_);
+  this->source.Seek(4);
+  this->DeleteTestFile();
+  this->source.Close();
```
asandryh: Do you delete the test file before closing it on purpose?

asandryh: Argh, nevermind, I've just read the test name. :)

src/parquet/util/input-output-test.cc
```diff
@@ -43,4 +49,59 @@ TEST(TestInMemoryOutputStream, Basics) {
...
+}
+
+TYPED_TEST(TestFileReaders, BadSeek) {
+  this->source.Open(this->test_path_);
+
+  ASSERT_THROW(this->source.Seek(this->filesize_ + 1), ParquetException);
+}
+
```
asandryh: We should also test Read() functionality. See an example in #56.

wesm: Will do

src/parquet/file/reader.cc
```diff
@@ -68,7 +68,7 @@ ParquetFileReader::ParquetFileReader() : schema_(nullptr) {}
 ParquetFileReader::~ParquetFileReader() {}
 
 std::unique_ptr<ParquetFileReader> ParquetFileReader::OpenFile(const std::string& path) {
-  std::unique_ptr<LocalFileSource> file(new LocalFileSource());
+  std::unique_ptr<MemoryMapSource> file(new MemoryMapSource());
```
asandryh: We should not hard-code using memory mapping (it is not that advantageous in our use case). You can add a flag to OpenFile() to select between LocalFileSource and MemoryMapSource.

wesm:
All set. let me know any more comments, thank you!

asandryh:
+1 LGTM.

wesm:
green build: https://travis-ci.org/wesm/parquet-cpp/builds/112676184

julienledem:
+1

wesm:
thank you!

PR #68, opened 2016-02-29
PARQUET-537: Ensure that LocalFileSource is properly closed.

asandryh:


wesm:
Unit test please.

asandryh:
Not sure what a unit test in this case will look like. I've just added explicit closing of sources in dtors.

wesm:
Examine the file descriptor after invoking the dtor -- you can use `fcntl` to see if it is still open http://man7.org/linux/man-pages/man2/fcntl.2.html

wesm:
I would just use `alltypes_plain.parquet` for this -- the test case can get refactored later to work on files that we generate using parquet-cpp

asandryh:
@wesm I've added a unit test. Please take a look.

julienledem:
+1

PR #69, opened 2016-03-01
PARQUET-519: Remove last of suppressed compiler warnings

wesm:


src/parquet/util/rle-test.cc
```diff
@@ -218,13 +218,13 @@ bool CheckRoundTrip(const vector<int>& values, int bit_width) {
       return false;
     }
   }
-  int encoded_len = encoder.Flush();
+  encoder.Flush();
```
asandryh: Add a check here? `EXPECT_EQ(..., encoder.Flush())`

wesm: The encoded length is data-dependent, but I set it to initialize the decoder with the encoded length

wesm:
green build https://travis-ci.org/wesm/parquet-cpp/builds/112921791

julienledem:
+1

PR #70, opened 2016-03-01
PARQUET-482: Organize public API headers

wesm:
Feedback requested. I'll add some simple Markdown documentation for the public API if we agree on the general approach.


asandryh:
Looks good. Since we still have example/decode_benchmark.cc around, its included headers should be replaced with parquet/api ones, like you did for parquet_reader and parquet-dump-schema.

asandryh:
+1

julienledem:
+1

PR #71, opened 2016-03-01
PARQUET-549: Add column reader tests for dictionary pages

majetideepak:


src/parquet/column/test-util.h
```diff
@@ -165,9 +166,108 @@ class DataPageBuilder {
...
+        {
+          for (int i = 0; i < values.size(); ++i) {
+            encoder.Put(values[i]);
+          }
+        });
+  // Dictionary encoding
+    dict_buffer_->Resize(encoder.dict_encoded_size());
+    encoder.WriteDict(dict_buffer_->mutable_data());
+    rle_indices_->Resize(encoder.EstimatedDataEncodedSize());
+    int actual_bytes = encoder.WriteIndices(rle_indices_->mutable_data(),
```
majetideepak: @wesm This took much longer than I expected.  This currently works for REQUIRED data.
 I am now blocked here for OPTIONAL data with `actual_bytes` being set to `-1`. The dictionary encoder says this is possible `If the supplied buffer is not big enough`. However, this should not happen if we use `EstimatedDataEncodedSize()`.
 I am looking into why the data being generated in the OPTIONAL case is causing this. Let me know if you have other comments. Thanks!

wesm: I haven't looked deeply at the rest of this patch but I don't see how required/optional is relevant here as this only impacts the repetition/definition levels.

src/parquet/column/column-reader-test.cc
```diff
@@ -124,8 +88,22 @@ class TestPrimitiveReader : public ::testing::Test {
     ASSERT_EQ(0, values_read);
   }
 
-  void execute(int num_pages, int levels_page, const ColumnDescriptor *d) {
-    MakePages(d, num_pages, levels_page);
+  void executePlain(int num_pages, int levels_per_page, const ColumnDescriptor *d) {
```
wesm: `ExecutePlain`

src/parquet/column/test-util.h
```diff
@@ -240,6 +352,126 @@ static void Paginate(const ColumnDescriptor *d,
...
+  if (encoding == Encoding::PLAIN) {
+    InitValues<typename Type::c_type>(num_values, values, buffer);
+    Paginate<Type>(d, values, def_levels, max_def_level,
+        rep_levels, max_rep_level, levels_per_page, values_per_page, pages);
+  } else if (encoding == Encoding::RLE_DICTIONARY) {
+    // Calls InitValues and repeats the data
+    InitDictValues<typename Type::c_type>(num_values, levels_per_page, values, buffer);
+    vector<uint8_t> rle_indices;
+
+    shared_ptr<DictionaryPage> dict_page = MakeDictPage<Type>(d, values, encoding,
```
wesm: This function is trying to do a large number of things. See the "single responsibility principle" applied to functions. 

For example, what if we wanted to have multiple dict-encoded data pages (1 dict page followed by many data pages)? It would be better to break down the tasks in this function into a set of reusable components.

src/parquet/column/test-util.h
```diff
@@ -165,9 +166,108 @@ class DataPageBuilder {
...
+    DictEncoder<TC> encoder(&pool_, type_length);
+    ASSERT_NO_THROW(
+        {
+          for (int i = 0; i < values.size(); ++i) {
+            encoder.Put(values[i]);
+          }
+        });
+  // Dictionary encoding
+    dict_buffer_->Resize(encoder.dict_encoded_size());
+    encoder.WriteDict(dict_buffer_->mutable_data());
```
wesm: Perhaps this can be refactored to return a buffer of RLE-encoded dict indices for each input chunk of values, then the dict can be written at the end for serialization purposes. This way it will be easier to add tests with multiple dict-encoded data pages (there may be new previously unobserved dict values from page to page)

src/parquet/column/column-reader-test.cc
```diff
@@ -45,42 +45,6 @@ namespace test {
 
 class TestPrimitiveReader : public ::testing::Test {
  public:
-  void MakePages(const ColumnDescriptor *d, int num_pages, int levels_per_page) {
```
wesm: Definitely good to factor this code out of here and `scanner-test` for better reuse

src/parquet/column/test-util.h
```diff
@@ -165,9 +166,108 @@ class DataPageBuilder {
...
+  void AppendValues(const ColumnDescriptor *d, const vector<TC>& values,
+      Encoding::type encoding = Encoding::RLE_DICTIONARY) {
+    int type_length = 0;
+    dict_buffer_ = std::make_shared<OwnedMutableBuffer>();
+    rle_indices_ = std::make_shared<OwnedMutableBuffer>();
+    if (TN == Type::FIXED_LEN_BYTE_ARRAY) {
+      type_length = d->type_length();
+    }
+    DictEncoder<TC> encoder(&pool_, type_length);
+    ASSERT_NO_THROW(
```
wesm: This `ASSERT_NO_THROW` wrapper can be omitted

src/parquet/column/test-util.h
```diff
@@ -165,9 +166,108 @@ class DataPageBuilder {
...
+
+  int32_t rle_indices_size() const {
+    return rle_indices_->size();
+  }
+
+ private:
+  InMemoryOutputStream* sink_;
+  shared_ptr<OwnedMutableBuffer> rle_indices_;
+  shared_ptr<OwnedMutableBuffer> dict_buffer_;
+  MemPool pool_;
```
wesm: You will have to call `FreeAll` on the MemPool at some point

src/parquet/column/test-util.h
majetideepak: I agree. The plan is to use `Paginate` here for the indices. But I could not think of an easy way to break the indices to generate multiple rle buffers.

src/parquet/column/test-util.h
majetideepak: Yeah! I can try doing this.

src/parquet/column/test-util.h
wesm: If you store the encoded indices in a `shared_ptr<Buffer>` object it will be more manageable (end up with `vector<shared_ptr<Buffer> >`. It would be better to migrate most of our test code to use buffers instead of raw `std::vector` for consistency

majetideepak: I will change them once I get the test running.

src/parquet/column/test-util.h
```diff
@@ -163,9 +164,109 @@ class DataPageBuilder {
...
+    }
+    int num_values = values.size();
+    DictEncoder<TC> encoder(&pool_, type_length);
+    for (int i = 0; i < num_values; ++i) {
+      encoder.Put(values[i]);
+    }
+    // Dictionary encoding
+    dict_buffer_->Resize(encoder.dict_encoded_size());
+    encoder.WriteDict(dict_buffer_->mutable_data());
+    rle_indices_->Resize(4*encoder.EstimatedDataEncodedSize());
```
majetideepak: @wesm this is buggy. I had to multiply it by a factor to get it working.

majetideepak:
I will work on creating multiple index pages tomorrow and re-factor with `shared_ptr<Buffer>`

majetideepak:
https://travis-ci.org/apache/parquet-cpp/jobs/113044090 clang reports incorrect error as specialized functions not being used.

src/parquet/column/test-util.h
majetideepak: I have a fix for this.

src/parquet/column/test-util.h
```diff
@@ -163,9 +166,109 @@ class DataPageBuilder {
...
+  test::DictionaryPageBuilder<Type> page_builder(&page_stream);
+
+  page_builder.AppendValues(d, values, encoding);
+
+  rle_indices.assign(page_builder.rle_indices(),
+      page_builder.rle_indices() + page_builder.rle_indices_size());
+
+  auto buffer = page_stream.GetBuffer();
+
+  page_builder.free_pool();
```
majetideepak: Frees the pool

src/parquet/column/test-util.h
```diff
@@ -177,39 +246,171 @@ void DataPageBuilder<Type::BOOLEAN>::AppendValues(const ColumnDescriptor *d,
...
+    return num_dict_values_;
+  }
+
+  Encoding::type encoding() const {
+    return encoding_;
+  }
+
+  int32_t rle_indices(shared_ptr<OwnedMutableBuffer>& rle_indices) {
+    rle_indices->Resize(sizeof(int) * encoder_->EstimatedDataEncodedSize());
+    int actual_bytes = encoder_->WriteIndices(rle_indices->mutable_data(),
```
majetideepak: @wesm I don't see a way to get a batch of indices rle-encoded without changing this API.

src/parquet/column/test-util.h
```diff
@@ -177,39 +246,171 @@ void DataPageBuilder<Type::BOOLEAN>::AppendValues(const ColumnDescriptor *d,
...
+  }
+
+  void AppendValues(const vector<TC>& values) {
+    int num_values = values.size();
+    // Dictionary encoding
+    for (int i = 0; i < num_values; ++i) {
+      encoder_->Put(values[i]);
+    }
+    dict_buffer_->Resize(encoder_->dict_encoded_size());
+    encoder_->WriteDict(dict_buffer_->mutable_data());
```
majetideepak: If we want a batch of dictionary data, we have to change this API to take an offset.

wesm: You only write the dictionary *after* you have computed the encoded indices for all the batches (at which point you have observed all of the dictionary values).

src/parquet/column/test-util.h
wesm: Could you be more specific? How about having `AppendValues` return a buffer containing the encoded dictionary indices? `WriteIndices` resets the state of the internal RleEncoder.

wesm: Or rather, it clears the `buffered_indices_` variable

src/parquet/column/test-util.h
majetideepak: I meant, if we want to split the dictionary across multiple dictionary pages

src/parquet/column/test-util.h
majetideepak: I did not see the `ClearIndices()` in the DictEncoder. This should be doable without any API changes.

src/parquet/column/test-util.h
wesm: I see. My understanding is that the format only permits one dictionary page per column chunk inside a row group.

majetideepak:
@wesm This is complete. Let me know if you have more feedback. Thank you!

src/parquet/column/test-util.h
```diff
@@ -234,6 +418,58 @@ static void Paginate(const ColumnDescriptor *d,
   }
 }
 
+// Generates plain pages from randomly generated data
```
majetideepak: will fix this comment.

src/parquet/column/scanner-test.cc
```diff
@@ -178,62 +181,50 @@ class TestFlatScanner : public ::testing::Test {
...
+  this->ExecuteAll(num_pages, num_levels_per_page, batch_size, 0, Encoding::PLAIN);
+}
+
+TYPED_TEST(TestFlatScanner, TestDictScanner) {
+  this->ExecuteAll(num_pages, num_levels_per_page, batch_size, 0,
+      Encoding::RLE_DICTIONARY);
+}
+
+TEST_F(TestBooleanFlatScanner, TestPlainScanner) {
+  this->ExecuteAll(num_pages, num_levels_per_page, batch_size, FLBA_LENGTH);
```
wesm: 0 instead of `FLBA_LENGTH`?

src/parquet/column/test-util.h
```diff
@@ -177,39 +203,196 @@ void DataPageBuilder<Type::BOOLEAN>::AppendValues(const ColumnDescriptor *d,
...
+  test::DictionaryPageBuilder<Type> page_builder(d, &page_stream, encoding);
+  int num_pages = values_per_page.size();
+  int value_start = 0;
+
+  for (int i = 0; i < num_pages; i++) {
+    page_builder.AppendValues(slice(values, value_start,
+          value_start + values_per_page[i]));
+    shared_ptr<OwnedMutableBuffer> indices_buff = std::make_shared<OwnedMutableBuffer>();
+    page_builder.rle_indices(indices_buff);
+    rle_indices.push_back(indices_buff);
```
wesm: Any reason not to have `AppendValues` return `std::shared_ptr<Buffer>` then this code as well as the `DictionaryPageBuilder` can be collapsed?

src/parquet/column/test-util.h
```diff
@@ -177,39 +203,196 @@ void DataPageBuilder<Type::BOOLEAN>::AppendValues(const ColumnDescriptor *d,
...
+    value_start += values_per_page[i];
+    page_builder.clear_indices();
+  }
+
+  page_builder.WriteDict();
+
+  auto buffer = page_stream.GetBuffer();
+
+  page_builder.free_pool();
```
wesm: Maybe put this in the page_builder dtor?

src/parquet/column/test-util.h
```diff
@@ -177,39 +203,196 @@ void DataPageBuilder<Type::BOOLEAN>::AppendValues(const ColumnDescriptor *d,
...
+    shared_ptr<OwnedMutableBuffer> indices_buff = std::make_shared<OwnedMutableBuffer>();
+    page_builder.rle_indices(indices_buff);
+    rle_indices.push_back(indices_buff);
+    value_start += values_per_page[i];
+    page_builder.clear_indices();
+  }
+
+  page_builder.WriteDict();
+
+  auto buffer = page_stream.GetBuffer();
```
wesm: You could consolidate some steps by having `WriteDict` return a Buffer directly instead of Buffer -> OutputStream -> Buffer, then you don't need to pass an outputstream into the ctor at all

src/parquet/column/test-util.h
```diff
@@ -177,39 +203,196 @@ void DataPageBuilder<Type::BOOLEAN>::AppendValues(const ColumnDescriptor *d,
...
+
+  page_builder.free_pool();
+  return std::make_shared<DictionaryPage>(buffer, page_builder.num_values(),
+      page_builder.encoding());
```
wesm: Should this always be `Encoding::PLAIN`? I believe that the `encoding_` bit can be removed from the page builder, too. Let me know what you think (I haven't scrutinized parquet-mr):

https://github.com/apache/parquet-format/blob/parquet-format-2.0.0/src/thrift/parquet.thrift#L158

wesm:
Thank you for doing all this -- as we continue writing tests we can expand upon the page generation utilities.

majetideepak:
All valid comments. Will fix them. Thanks!

src/parquet/column/test-util.h
majetideepak: One dis-advantage is that we cannot call  `AppendValues` multiple times for a single rle-buffer.
Should I keep this instead ?

wesm: I'm not sure it's worth the extra boilerplate, up to you

majetideepak: I removed it for now. Let's keep it simple.

wesm:
Looks good -- I can merge whenever we have a green build, I can pull this up on my mac in a bit and see if I can repro the clang failure locally

wesm:
Ah the clang bug is because of a `vector<bool>` issue. For some reason this works on my Mac (Apple LLVM 7.0.2, so newer than Travis) but must be inconsistent across C++ stdlib implementations

majetideepak:
Thanks! this patch should fix the `vector<bool>` issue.

majetideepak:
@wesm this is good to merge. Thank you!

wesm:
+1, thank you

PR #72, opened 2016-03-02
PARQUET-542: Support custom memory allocators

asandryh:
Added `MemoryAllocator` interface and a default implementation, ensured `MemPool` can use a custom allocator, added `Vector<T>` to replace `std::vector<T>`.

### DIFF
--- a/src/parquet/column/scanner.cc
+++ b/src/parquet/column/scanner.cc
@@ -25,24 +25,25 @@
 namespace parquet_cpp {
 
 std::shared_ptr<Scanner> Scanner::Make(std::shared_ptr<ColumnReader> col_reader,
-    int64_t batch_size) {
+    int64_t batch_size, MemoryAllocator* allocator) {
   switch (col_reader->type()) {
     case Type::BOOLEAN:
-      return std::make_shared<BoolScanner>(col_reader, batch_size);
+      return std::make_shared<BoolScanner>(col_reader, batch_size, allocator);
     case Type::INT32:
-      return std::make_shared<Int32Scanner>(col_reader, batch_size);
+      return std::make_shared<Int32Scanner>(col_reader, batch_size, allocator);
     case Type::INT64:
-      return std::make_shared<Int64Scanner>(col_reader, batch_size);
+      return std::make_shared<Int64Scanner>(col_reader, batch_size, allocator);
     case Type::INT96:
-      return std::make_shared<Int96Scanner>(col_reader, batch_size);
+      return std::make_shared<Int96Scanner>(col_reader, batch_size, allocator);
     case Type::FLOAT:
-      return std::make_shared<FloatScanner>(col_reader, batch_size);
+      return std::make_shared<FloatScanner>(col_reader, batch_size, allocator);
     case Type::DOUBLE:
-      return std::make_shared<DoubleScanner>(col_reader, batch_size);
+      return std::make_shared<DoubleScanner>(col_reader, batch_size, allocator);
     case Type::BYTE_ARRAY:
-      return std::make_shared<ByteArrayScanner>(col_reader, batch_size);
+      return std::make_shared<ByteArrayScanner>(col_reader, batch_size, allocator);
     case Type::FIXED_LEN_BYTE_ARRAY:
-      return std::make_shared<FixedLenByteArrayScanner>(col_reader, batch_size);
+      return std::make_shared<FixedLenByteArrayScanner>(col_reader,
+          batch_size, allocator);
     default:
       ParquetException::NYI("type reader not implemented");
   }
--- a/src/parquet/column/scanner.h
+++ b/src/parquet/column/scanner.h
@@ -29,6 +29,7 @@
 #include "parquet/exception.h"
 #include "parquet/schema/descriptor.h"
 #include "parquet/types.h"
+#include "parquet/util/mem-allocator.h"
 
 namespace parquet_cpp {
 
@@ -37,10 +38,12 @@ static constexpr int64_t DEFAULT_SCANNER_BATCH_SIZE = 128;
 class Scanner {
  public:
   explicit Scanner(std::shared_ptr<ColumnReader> reader,
-      int64_t batch_size = DEFAULT_SCANNER_BATCH_SIZE) :
+      int64_t batch_size = DEFAULT_SCANNER_BATCH_SIZE,
+      MemoryAllocator* allocator = default_allocator()) :
       batch_size_(batch_size),
       level_offset_(0),
       levels_buffered_(0),
+      value_buffer_(0, allocator),
       value_offset_(0),
       values_buffered_(0),
       reader_(reader) {
@@ -52,7 +55,8 @@ class Scanner {
   virtual ~Scanner() {}
 
   static std::shared_ptr<Scanner> Make(std::shared_ptr<ColumnReader> col_reader,
-      int64_t batch_size = DEFAULT_SCANNER_BATCH_SIZE);
+      int64_t batch_size = DEFAULT_SCANNER_BATCH_SIZE,
+      MemoryAllocator* allocator = default_allocator());
 
   virtual void PrintNext(std::ostream& out, int width) = 0;
 
@@ -78,7 +82,7 @@ class Scanner {
   int level_offset_;
   int levels_buffered_;
 
-  std::vector<uint8_t> value_buffer_;
+  OwnedMutableBuffer value_buffer_;
   int value_offset_;
   int64_t values_buffered_;
 
@@ -93,11 +97,12 @@ class TypedScanner : public Scanner {
   typedef typename type_traits<TYPE>::value_type T;
 
   explicit TypedScanner(std::shared_ptr<ColumnReader> reader,
-      int64_t batch_size = DEFAULT_SCANNER_BATCH_SIZE) :
-      Scanner(reader, batch_size) {
+      int64_t batch_size = DEFAULT_SCANNER_BATCH_SIZE,
+      MemoryAllocator* allocator = default_allocator()) :
+      Scanner(reader, batch_size, allocator) {
     typed_reader_ = static_cast<TypedColumnReader<TYPE>*>(reader.get());
     int value_byte_size = type_traits<TYPE>::value_byte_size;
-    value_buffer_.resize(batch_size_ * value_byte_size);
+    value_buffer_.Resize(batch_size_ * value_byte_size);
     values_ = reinterpret_cast<T*>(&value_buffer_[0]);
   }
 
--- a/src/parquet/column/test-util.h
+++ b/src/parquet/column/test-util.h
@@ -252,7 +252,7 @@ class DictionaryPageBuilder {
         if (TN == Type::FIXED_LEN_BYTE_ARRAY) {
           type_length = d->type_length();
         }
-        encoder_.reset(new DictEncoder<TC>(&pool_, type_length));
+        encoder_.reset(new DictEncoder<TC>(&pool_, default_allocator(), type_length));
   }
 
   ~DictionaryPageBuilder() {
@@ -260,7 +260,6 @@ class DictionaryPageBuilder {
   }
 
   shared_ptr<Buffer> AppendValues(const vector<TC>& values) {
-    shared_ptr<OwnedMutableBuffer> rle_indices = std::make_shared<OwnedMutableBuffer>();
     int num_values = values.size();
     // Dictionary encoding
     for (int i = 0; i < num_values; ++i) {
@@ -268,7 +267,8 @@ class DictionaryPageBuilder {
     }
     num_dict_values_ = encoder_->num_entries();
     have_values_ = true;
-    rle_indices->Resize(sizeof(int) * encoder_->EstimatedDataEncodedSize());
+    shared_ptr<OwnedMutableBuffer> rle_indices = std::make_shared<OwnedMutableBuffer>(
+        sizeof(int) * encoder_->EstimatedDataEncodedSize());
     int actual_bytes = encoder_->WriteIndices(rle_indices->mutable_data(),
         rle_indices->size());
     rle_indices->Resize(actual_bytes);
@@ -277,8 +277,8 @@ class DictionaryPageBuilder {
   }
 
   shared_ptr<Buffer> WriteDict() {
-    shared_ptr<OwnedMutableBuffer> dict_buffer = std::make_shared<OwnedMutableBuffer>();
-    dict_buffer->Resize(encoder_->dict_encoded_size());
+    shared_ptr<OwnedMutableBuffer> dict_buffer = std::make_shared<OwnedMutableBuffer>(
+        encoder_->dict_encoded_size());
     encoder_->WriteDict(dict_buffer->mutable_data());
     return dict_buffer;
   }
--- a/src/parquet/encodings/dictionary-encoding.h
+++ b/src/parquet/encodings/dictionary-encoding.h
@@ -27,8 +27,10 @@
 #include "parquet/encodings/decoder.h"
 #include "parquet/encodings/encoder.h"
 #include "parquet/encodings/plain-encoding.h"
+#include "parquet/util/buffer.h"
 #include "parquet/util/cpu-info.h"
 #include "parquet/util/hash-util.h"
+#include "parquet/util/mem-allocator.h"
 #include "parquet/util/mem-pool.h"
 #include "parquet/util/rle-encoding.h"
 
@@ -42,9 +44,10 @@ class DictionaryDecoder : public Decoder<TYPE> {
   // Initializes the dictionary with values from 'dictionary'. The data in
   // dictionary is not guaranteed to persist in memory after this call so the
   // dictionary decoder needs to copy the data out if necessary.
-  explicit DictionaryDecoder(const ColumnDescriptor* descr)
-      : Decoder<TYPE>(descr, Encoding::RLE_DICTIONARY) {
-  }
+  explicit DictionaryDecoder(const ColumnDescriptor* descr,
+      MemoryAllocator* allocator = default_allocator()):
+      Decoder<TYPE>(descr, Encoding::RLE_DICTIONARY), dictionary_(0, allocator),
+      byte_array_data_(0, allocator) {}
 
   // Perform type-specific initiatialization
   void SetDict(Decoder<TYPE>* dictionary);
@@ -77,11 +80,11 @@ class DictionaryDecoder : public Decoder<TYPE> {
   }
 
   // Only one is set.
-  std::vector<T> dictionary_;
+  Vector<T> dictionary_;
 
   // Data that contains the byte array data (byte_array_dictionary_ just has the
   // pointers).
-  std::vector<uint8_t> byte_array_data_;
+  OwnedMutableBuffer byte_array_data_;
 
   RleDecoder idx_decoder_;
 };
@@ -89,7 +92,7 @@ class DictionaryDecoder : public Decoder<TYPE> {
 template <int TYPE>
 inline void DictionaryDecoder<TYPE>::SetDict(Decoder<TYPE>* dictionary) {
   int num_dictionary_values = dictionary->values_left();
-  dictionary_.resize(num_dictionary_values);
+  dictionary_.Resize(num_dictionary_values);
   dictionary->Decode(&dictionary_[0], num_dictionary_values);
 }
 
@@ -103,14 +106,14 @@ template <>
 inline void DictionaryDecoder<Type::BYTE_ARRAY>::SetDict(
     Decoder<Type::BYTE_ARRAY>* dictionary) {
   int num_dictionary_values = dictionary->values_left();
-  dictionary_.resize(num_dictionary_values);
+  dictionary_.Resize(num_dictionary_values);
   dictionary->Decode(&dictionary_[0], num_dictionary_values);
 
   int total_size = 0;
   for (int i = 0; i < num_dictionary_values; ++i) {
     total_size += dictionary_[i].len;
   }
-  byte_array_data_.resize(total_size);
+  byte_array_data_.Resize(total_size);
   int offset = 0;
   for (int i = 0; i < num_dictionary_values; ++i) {
     memcpy(&byte_array_data_[offset], dictionary_[i].ptr, dictionary_[i].len);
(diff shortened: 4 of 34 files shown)


src/parquet/util/memory-pool.h
```diff
@@ -0,0 +1,44 @@
...
+#include "parquet/util/logging.h"
+#include "parquet/util/bit-util.h"
+
+namespace parquet_cpp {
+
+class MemoryPool {
+ public:
+  virtual ~MemoryPool();
+
+  virtual uint8_t* Allocate(uint64_t size) = 0;
```
wesm: Use `int64_t` for the memory size

src/parquet/util/memory-pool.h
```diff
@@ -0,0 +1,44 @@
...
+  virtual void Free(const uint8_t* p) = 0;
```
wesm: Should `Free` also be passed the buffer size? Otherwise memory usage is hard to track.

src/parquet/util/memory-pool-test.cc
```diff
@@ -0,0 +1,39 @@
...
+// specific language governing permissions and limitations
+// under the License.
+
+// Initially imported from Apache Impala on 2016-02-23, and has been modified
+// since for parquet-cpp
+
+#include <gtest/gtest.h>
+
+#include "parquet/exception.h"
+#include "parquet/util/memory-pool.h"
```
wesm: This include should go first in the file

wesm: oops, wrong file, see below

src/parquet/util/memory-pool.cc
```diff
@@ -0,0 +1,54 @@
...
+// software distributed under the License is distributed on an
+// "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+// KIND, either express or implied.  See the License for the
+// specific language governing permissions and limitations
+// under the License.
+
+#include <cstdlib>
+
+#include "parquet/exception.h"
+#include "parquet/util/memory-pool.h"
```
wesm: This include should go first in the file

src/parquet/util/buffer.h
```diff
@@ -127,12 +130,14 @@ class ResizableBuffer : public MutableBuffer {
 // garbage-collected
 class OwnedMutableBuffer : public ResizableBuffer {
  public:
-  OwnedMutableBuffer();
+  explicit OwnedMutableBuffer(MemoryPool* pool = get_default_pool());
```
wesm: You *could* forward-declare `class MemoryPool;` then handle getting the default pool in `buffer.cc`, but not a big deal either way

src/parquet/util/memory-pool.h
wesm: Also why `const uint8_t* p` instead of `uint8_t* p`? It would be better for callers to have to explicitly remove const if they are trying to free a const pointer (which in theory should rarely occur)

src/parquet/util/buffer.cc
```diff
@@ -34,18 +34,33 @@ std::shared_ptr<Buffer> MutableBuffer::GetImmutableView() {
...
+    memory_pool_->Free(data_);
+  }
+}
+
+void OwnedMutableBuffer::Reserve(int64_t new_capacity) {
+  if (!data_ || new_capacity > capacity_) {
+    if (data_) {
+      uint8_t* new_data = memory_pool_->Allocate(new_capacity);
+      memcpy(new_data, data_, size_);
+      memory_pool_->Free(data_);
```
wesm: I feel it would be better to call `Free(mutable_data_)` here, then assign `data_ = mutable_data_` at the end and avoid `const_const` altogether

src/parquet/util/CMakeLists.txt
```diff
@@ -40,6 +41,7 @@ add_library(parquet_util STATIC
   cpu-info.cc
   input.cc
   mem-pool.cc
+  memory-pool.cc
```
wesm: Perhaps we could refactor the Impala `mem-pool.cc` (whose API is different from the simpler Allocate/Free memory pool) to be called `ChunkedMemPool` or something and use the DefaultMemoryPool for its chunk malloc/free calls? We could also nix the chunked pool altogether in favor of quadratically-growing allocations in `DictEncoder`

asandryh: Yes, that's on my to-do list.

src/parquet/util/memory-pool.h
```diff
@@ -0,0 +1,44 @@
...
+#define PARQUET_UTIL_MEMORY_POOL_H
+
+#include <stdio.h>
+#include <algorithm>
+#include <cstdint>
+#include <vector>
+#include <string>
+
+#include "parquet/util/logging.h"
+#include "parquet/util/bit-util.h"
```
wesm: most of these includes are unneeded

asandryh:
@wesm Thank you for the comments! I'll be updating this patch to address feedback and implement further functionality.

src/parquet/util/memory-pool.h
asandryh: Argh, I copy-pasted this and forgot to clean up. Thanks!

src/parquet/column/scanner.h
```diff
@@ -78,7 +78,7 @@ class Scanner {
   int level_offset_;
   int levels_buffered_;
 
-  std::vector<uint8_t> value_buffer_;
+  OwnedMutableBuffer value_buffer_;
```
wesm: I figure most of these will have to be revisited as we thread a top-level MemoryPool through each of these classes (any allocated memory would be requested from the active MemoryPool in that context rather than the default one)

asandryh: Yes, that's an item on the above to-do list.

wesm:
Great, I will stay tuned, thank you

src/parquet/util/buffer.cc
```diff
@@ -34,18 +35,68 @@ std::shared_ptr<Buffer> MutableBuffer::GetImmutableView() {
...
 
+template class OwnedMutableBuffer<uint8_t>;
+template class OwnedMutableBuffer<int32_t>;
+template class OwnedMutableBuffer<int64_t>;
+template class OwnedMutableBuffer<bool>;
+template class OwnedMutableBuffer<float>;
+template class OwnedMutableBuffer<double>;
+template class OwnedMutableBuffer<Int96>;
+template class OwnedMutableBuffer<ByteArray>;
+template class OwnedMutableBuffer<FixedLenByteArray>;
```
wesm: I'm not sure about this. I would prefer that buffers only be concerned with raw memory but otherwise data types are handled elsewhere. Having a replacement for `std::vector<T>` that allocated memory from the pool seems reasonable, but that seems like a separate abstraction.

asandryh: Functionality-wise, `OwnedMutableBuffer<T>` is precisely a replacement of `std::vector<T,allocator>`. Implementing it elsewhere would either require inheriting from `OwnedMutableBuffer` or unnecessary code duplication. IMHO, this is the most concise implementation.

wesm: At the moment, the only place where `T` is not `uint8_t` is in the dict encoder hash table implementation. I'm not arguing against having an `std::vector` replacement, only that it would be better to do it using object composition (so no inheritance) or implementing a custom STL allocator. This doesn't strike me as code duplication since it is introducing a new functional concept (typed value access and other APIs) to the API.

asandryh: Ok, I've moved this functionality to a separate DataBuffer<T> class.

src/parquet/util/buffer.h
```diff
@@ -127,12 +129,39 @@ class ResizableBuffer : public MutableBuffer {
...
   // TODO: aligned allocations
-  std::vector<uint8_t> buffer_owner_;
+  MemoryPool* memory_pool_;
+
+  OwnedMutableBuffer(OwnedMutableBuffer& buffer) = delete;
+  OwnedMutableBuffer& operator=(OwnedMutableBuffer& buffer) = delete;
+};
+
+template <class T>
+class DataBuffer {
```
wesm: Let's call this something semantically distinct, like `Vector` -- my objection to the prior layering was about object semantics.

src/parquet/util/buffer.cc
```diff
@@ -34,18 +35,91 @@ std::shared_ptr<Buffer> MutableBuffer::GetImmutableView() {
...
+}
+
+template class DataBuffer<int32_t>;
+template class DataBuffer<int64_t>;
+template class DataBuffer<bool>;
+template class DataBuffer<float>;
+template class DataBuffer<double>;
+template class DataBuffer<Int96>;
+template class DataBuffer<ByteArray>;
+template class DataBuffer<FixedLenByteArray>;
```
wesm: Is there a reason to force-instantiate these templates?

src/parquet/util/buffer.h
```diff
@@ -127,12 +129,39 @@ class ResizableBuffer : public MutableBuffer {
...
+ public:
+  explicit DataBuffer(int64_t size = 0, MemoryPool* pool = get_default_pool());
+  void Resize(int64_t new_size);
+  void Reserve(int64_t new_capacity);
+  void Clear();
+  void Assign(int64_t size, const T val);
+  T& operator[](int64_t i);
```
wesm: By not inlining `[]`, I suspect it would introduce a measurable performance impact on the dictionary encoding hot path, but we don't have benchmarks to see one way or the other.

src/parquet/util/buffer.cc
wesm: I see; it's because the methods are not inlined (per below)

asandryh:
@wesm This PR is complete. Can you please review?

src/parquet/column/scanner.cc
```diff
@@ -25,24 +25,24 @@
...
+      return std::make_shared<FloatScanner>(col_reader, pool, batch_size);
     case Type::DOUBLE:
-      return std::make_shared<DoubleScanner>(col_reader, batch_size);
+      return std::make_shared<DoubleScanner>(col_reader, pool, batch_size);
     case Type::BYTE_ARRAY:
-      return std::make_shared<ByteArrayScanner>(col_reader, batch_size);
+      return std::make_shared<ByteArrayScanner>(col_reader, pool, batch_size);
     case Type::FIXED_LEN_BYTE_ARRAY:
-      return std::make_shared<FixedLenByteArrayScanner>(col_reader, batch_size);
+      return std::make_shared<FixedLenByteArrayScanner>(col_reader, pool, batch_size);
```
wesm: Do you envision that Scanners would use a different MemoryPool from the ColumnReader?

src/parquet/column/scanner.h
```diff
@@ -37,10 +38,13 @@ static constexpr int64_t DEFAULT_SCANNER_BATCH_SIZE = 128;
 class Scanner {
  public:
   explicit Scanner(std::shared_ptr<ColumnReader> reader,
+      MemoryPool* pool,
       int64_t batch_size = DEFAULT_SCANNER_BATCH_SIZE) :
       batch_size_(batch_size),
       level_offset_(0),
       levels_buffered_(0),
+      value_buffer_(0, pool),
+//      value_buffer_(0, get_default_pool()),
```
wesm: Delete this

src/parquet/column/test-util.h
```diff
@@ -316,7 +315,7 @@ template <typename Type>
 static shared_ptr<DictionaryPage> MakeDictPage(const ColumnDescriptor *d,
     const vector<typename Type::c_type>& values, const vector<int>& values_per_page,
     Encoding::type encoding, vector<shared_ptr<Buffer> >& rle_indices) {
-  InMemoryOutputStream page_stream;
+  InMemoryOutputStream page_stream(get_default_pool());
```
wesm: All of these `get_default_pool` calls are very intrusive — why not set the default argument value to `nullptr` and use the default pool if it is not set by the user?

src/parquet/encodings/dictionary-encoding.h
```diff
@@ -418,6 +432,23 @@ inline void DictEncoder<FixedLenByteArray>::WriteDict(uint8_t* buffer) {
...
+    pool_->Free(const_cast<uint8_t*>(value.ptr));
+  }
+}
+
+template<>
+inline void DictEncoder<FixedLenByteArray>::FreeMemory() {
+  for (const auto& value : uniques_) {
+    pool_->Free(const_cast<uint8_t*>(value.ptr));
+  }
+}
```
wesm: This code is concerning given that we don't have performance benchmarks. In the prior version, we were requesting memory from a chunked allocator, then `FreeAll` was deallocating chunks. Here, if we have a lot of small strings, all of these small frees (virtual function calls, plus `std::free`) are likely to have much worse performance.

src/parquet/encodings/encoding-test.cc
```diff
@@ -301,7 +300,7 @@ TYPED_TEST(TestDictionaryEncoding, BasicRoundTrip) {
 
 TEST(TestDictionaryEncoding, CannotDictDecodeBoolean) {
   PlainDecoder<Type::BOOLEAN> dict_decoder(nullptr);
-  DictionaryDecoder<Type::BOOLEAN> decoder(nullptr);
+  DictionaryDecoder<Type::BOOLEAN> decoder(nullptr, get_default_pool());
```
wesm: Per above I don't think it makes sense to sprinkle `get_default_pool` everywhere — it adds complexity where not strictly necessary.

src/parquet/file/file-deserialize-test.cc
```diff
@@ -275,7 +277,7 @@ TEST_F(TestParquetFileReader, InvalidFooter) {
 }
 
 TEST_F(TestParquetFileReader, IncompleteMetadata) {
-  InMemoryOutputStream stream;
+  InMemoryOutputStream stream(get_default_pool());
```
wesm: ditto

src/parquet/util/memory-pool.h
```diff
@@ -0,0 +1,39 @@
...
+
+namespace parquet_cpp {
+
+class MemoryPool {
+ public:
+  virtual ~MemoryPool();
+
+  // Returns nullptr if size is 0
+  virtual uint8_t* Allocate(int64_t size) = 0;
+  virtual void Free(uint8_t* p) = 0;
```
wesm: I still think we should be keeping track of the total allocated memory in the pool, which means we'd need to make this `Free(uint8_t* data, int64_t size)`, what do you think?

wesm:
Almost all of the `get_default_pool()` calls should be replaced with `pool = nullptr` default arguments, which will make things easier for both users and developers — to be clear these should only be the points of contact with the public API.

src/parquet/encodings/dictionary-encoding.h
wesm: I would suggest that `ChunkedMemoryPool` not be a subclass of `MemoryPool` and instead take a memory pool as a parameter. That way we won't have this problem. The MemPool from Impala only really makes sense with a bunch of allocations followed by `FreeAll` (in the context of a step in a data flow system / SQL engine)

src/parquet/column/test-util.h
asandryh: I feel strongly against default args for memory pools - it's too easy to forget to specify a memory pool. However, these `get_default_pool()` do pollute the code - I'll try to come up with a better solution.

src/parquet/encodings/dictionary-encoding.h
asandryh: `ChunkedMemoryPool::Free()` is a no-op; it's only there to implement `MemoryPool::Free()` method. All deallocation takes places in the dtor. We don't take a performance hit here.

src/parquet/util/memory-pool.h
asandryh: I have implemented memory usage tracking in my PARQUET-552 patch. I'll move this functionality to this patch.

src/parquet/encodings/dictionary-encoding.h
wesm: Not sure I follow. Doesn't this hit the vtable for each Free call? Even if the function doesn't do anything, this does add overhead.

asandryh: Ah, in that sense, yes, we do a vtable look-up there.
I'll add a `bool MemoryPool::NeedToFree()` method that indicates whether `Free()` needs to be called explicitly for a pool.

wesm: I am still pretty uncomfortable with this. I don't think we should ever use other memory allocators than the chunked one in dictionary encoding -- if DictEncoder is initialized with a normal memory pool, it can initialize the ChunkedMemoryPool (which has the passed MemoryPool as a member variable, rather than using inheritance, to chunk memory allocations, then freeing them all once complete).

asandryh: `ChunkedMemoryPool`is only one possible implementation of a memory pool. It does not free specified memory and cannot reuse freed memory. It is not a cure-all solution.
IMHO, we should only provide a memory pool interface (which includes at least `Allocate` and `Free` methods) and leave the implementation trade-offs to users. They can use the implemented `TrackingMemoryPool` or `ChunkedMemoryPool`, or implement their own pool.
To alleviate your concern, I propose to modify `get_default_pool()` to return `ChunkedMemoryPool` rather than `TrackedMemoryPool`.

wesm: We have a naming problem -- the `MemPool` that I imported from Impala is a different kind of abstraction, and we should not conflate them here. Perhaps we can find a different name rather than thinking of it as an implementation of the `MemoryPool` interface (which I think we should do)?

asandryh: How about rename `MemoryPool` to `MemoryAllocator` and `TrackingMemoryPool` to `TrackingMemoryAllocator`(because they are basically only allocators)? Then use a `MemoryAllocator` in `ChunkedMemoryPool`.

wesm: Sounds good to me.

asandryh:
@wesm I've updated the PR following our latest discussions. (I've also rolled back some changes, such as renaming `MemPool` to `ChunkedMemoryPool`). Please have a look.

src/parquet/encodings/dictionary-encoding.h
```diff
@@ -360,7 +367,12 @@ inline void DictEncoder<T>::DoubleTableSize() {
 
   hash_table_size_ = new_size;
   mod_bitmask_ = new_size - 1;
-  new_hash_slots.swap(hash_slots_);
+
+  hash_slots_.Clear();
+  hash_slots_.Reserve(hash_table_size_);
+  for (int i = 0; i < hash_table_size_; i++) {
+    hash_slots_[i] = new_hash_slots[i];
+  }
```
wesm: This has a distinctly different performance profile than using `std::vector::swap`. Perhaps we should implement `swap` for our `Vector`?

src/parquet/util/buffer.h
```diff
@@ -127,12 +129,41 @@ class ResizableBuffer : public MutableBuffer {
...
+  }
+
+ private:
+  OwnedMutableBuffer buffer_;
+  int64_t size_;
+  int64_t capacity_;
+  T* data_;
+
+  Vector(Vector& buffer) = delete;
+  Vector& operator=(Vector& buffer) = delete;
```
wesm: You can use the `DISALLOW_COPY_AND_ASSIGN` macro for this in the future

src/parquet/util/input-output-test.cc
```diff
@@ -115,10 +116,9 @@ TYPED_TEST(TestFileReaders, FileDisappeared) {
   this->DeleteTestFile();
   this->source.Close();
 }
-
+//
```
wesm: ?

src/parquet/util/mem-allocator.cc
```diff
@@ -0,0 +1,67 @@
...
+namespace parquet_cpp {
+
+MemoryAllocator::~MemoryAllocator() {}
+
+uint8_t* TrackingAllocator::Malloc(int64_t size) {
+  if (0 == size) {
+    return nullptr;
+  }
+
+  // store the size at the beginning of the block
```
wesm: Interesting. One consequence of this approach is that allocators cannot necessarily take ownership of a piece of memory. I'm curious what is the resistance (if any) to adding `int64_t size` to the `Free` function. In every usage of `MemoryAllocator::Free` in this patch we have the size accessible. Is there a concern that passing the buffer size with `Free` might be error prone or cause other complications?

src/parquet/util/mem-allocator.h
```diff
@@ -0,0 +1,61 @@
...
+  virtual ~TrackingAllocator();
+
+  uint8_t* Malloc(int64_t size) override;
+  void Free(uint8_t* p) override;
+
+  inline int64_t TotalMemory() {
+    return total_memory_;
+  }
+
+  inline int64_t MaxMemory() {
```
wesm: `inline` is redundant here

src/parquet/util/mem-allocator.h
```diff
@@ -0,0 +1,61 @@
...
+    return max_memory_;
+  }
+
+ private:
+  static const uint64_t OFFSET = sizeof(int64_t);
```
wesm: Perhaps make this a `static constexpr` in `mem-allocator.cc`. But as above I'm somewhat unsure about whether embedding the buffer size in the memory is a good idea.

src/parquet/util/mem-pool.cc
```diff
@@ -108,7 +109,8 @@ bool MemPool::FindChunk(int64_t min_size) {
     chunk_size = std::max<int64_t>(min_size, next_chunk_size_);
 
     // Allocate a new chunk. Return early if malloc fails.
-    uint8_t* buf = reinterpret_cast<uint8_t*>(malloc(chunk_size));
+//    uint8_t* buf = reinterpret_cast<uint8_t*>(malloc(chunk_size));
```
wesm: Delete this

wesm:
This looks good outside my comments re: the TrackingAllocator. 

Note that Travis CI is checking for memory leaks now, but it looks like we need to make it print out the leak details. If you figure out how to do that in ctest feel free to make a tweak

src/parquet/encodings/dictionary-encoding.h
asandryh: Excellent point. I'll add a `Swap` method.

src/parquet/util/buffer.h
asandryh: Will do, for consistency.

src/parquet/util/mem-allocator.cc
asandryh: Other than the preference to hide as many implementation details from users as possible, I don't have a strong objection against passing `size` to `Free()`. I'll add this parameter; it does not look error-prone.

asandryh:
@wesm I've updated the PR following your comments.

asandryh:
Regarding memory leak reporting in Travis CI: It looks like valgrind does not catch memory leaks in ctest. I've tried deliberately adding a memory leak to mem-allocator-test.cc. Running 
`valgrind --tool=memcheck --leak-check=full ctest`
did not catch the leak; while
`valgrind --tool=memcheck --leak-check=full debug/mem-allocator-test` 
caught it. We should create a jira to fix our Travis CI script to ensure that it catches memory leaks.

xhochy:
@asandryh Note that your first command only checks `ctest` for memory leaks, not your code. See the recent changes in master regarding valgrind. There is now a `cmake` flag for this: `-DPARQUET_TEST_MEMCHECK=ON`.

asandryh:
@xhochy Thank you, I missed setting this flag.

wesm:
+1, thank you very much

PR #73, opened 2016-03-03
PARQUET-555: Dictionary page metadata handling inconsistencies

majetideepak:
Includes tests

### DIFF
--- a/src/parquet/column/column-reader-test.cc
+++ b/src/parquet/column/column-reader-test.cc
@@ -155,5 +155,37 @@ TEST_F(TestPrimitiveReader, TestInt32FlatRepeated) {
   ExecuteDict(num_pages, levels_per_page, &descr);
 }
 
+TEST_F(TestPrimitiveReader, TestPageReader) {
+  max_def_level_ = 0;
+  max_rep_level_ = 0;
+  NodePtr type = schema::Int32("a", Repetition::REQUIRED);
+  const ColumnDescriptor descr(type, max_def_level_, max_rep_level_);
+  shared_ptr<OwnedMutableBuffer> dummy = std::make_shared<OwnedMutableBuffer>();
+  shared_ptr<DataPage> data_page = MakeDataPage<Int32Type>(&descr, {}, 0,
+      Encoding::RLE_DICTIONARY, {}, 0, {}, 0, {}, 0);
+  pages_.push_back(data_page);
+  InitReader(&descr);
+  // Tests dictionary page must occur before data page
+  ASSERT_THROW(reader_->HasNext(), ParquetException);
+  shared_ptr<DictionaryPage> dict_page = std::make_shared<DictionaryPage>(dummy,
+      0, Encoding::RLE);
+  pages_.clear();
+  pages_.push_back(dict_page);
+  InitReader(&descr);
+  // Tests only RLE_DICTIONARY is supported
+  ASSERT_THROW(reader_->HasNext(), ParquetException);
+
+  shared_ptr<DictionaryPage> dict_page1 = std::make_shared<DictionaryPage>(dummy,
+      0, Encoding::PLAIN_DICTIONARY);
+  shared_ptr<DictionaryPage> dict_page2 = std::make_shared<DictionaryPage>(dummy,
+      0, Encoding::PLAIN);
+  pages_.clear();
+  pages_.push_back(dict_page1);
+  pages_.push_back(dict_page2);
+  InitReader(&descr);
+  // Column cannot have more than one dictionary
+  ASSERT_THROW(reader_->HasNext(), ParquetException);
+}
+
 } // namespace test
 } // namespace parquet_cpp
--- a/src/parquet/column/column-reader-test.cc
+++ b/src/parquet/column/column-reader-test.cc
@@ -161,30 +161,63 @@ TEST_F(TestPrimitiveReader, TestPageReader) {
   NodePtr type = schema::Int32("a", Repetition::REQUIRED);
   const ColumnDescriptor descr(type, max_def_level_, max_rep_level_);
   shared_ptr<OwnedMutableBuffer> dummy = std::make_shared<OwnedMutableBuffer>();
+
+  shared_ptr<DictionaryPage> dict_page = std::make_shared<DictionaryPage>(dummy,
+      0, Encoding::PLAIN);
   shared_ptr<DataPage> data_page = MakeDataPage<Int32Type>(&descr, {}, 0,
+      Encoding::RLE_DICTIONARY, {}, 0, {}, 0, {}, 0);
+  pages_.push_back(dict_page);
+  pages_.push_back(data_page);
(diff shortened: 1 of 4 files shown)


majetideepak:
@wesm let me know your feedback. Thank you!

src/parquet/column/column-reader-test.cc
```diff
@@ -155,5 +155,37 @@ TEST_F(TestPrimitiveReader, TestInt32FlatRepeated) {
...
+  const ColumnDescriptor descr(type, max_def_level_, max_rep_level_);
+  shared_ptr<OwnedMutableBuffer> dummy = std::make_shared<OwnedMutableBuffer>();
+  shared_ptr<DataPage> data_page = MakeDataPage<Int32Type>(&descr, {}, 0,
+      Encoding::RLE_DICTIONARY, {}, 0, {}, 0, {}, 0);
+  pages_.push_back(data_page);
+  InitReader(&descr);
+  // Tests dictionary page must occur before data page
+  ASSERT_THROW(reader_->HasNext(), ParquetException);
+  shared_ptr<DictionaryPage> dict_page = std::make_shared<DictionaryPage>(dummy,
+      0, Encoding::RLE);
```
wesm: `RLE` is not a valid encoding for dictionary pages -- perhaps use one of the parquet 2.0 value encodings? Also, since Impala generates dictionary pages with `PLAIN_DICTIONARY` (https://github.com/cloudera/Impala/blob/cdh5-trunk/be/src/exec/hdfs-parquet-table-writer.cc#L445) we should make sure this does not throw an exception

src/parquet/column/reader.cc
```diff
@@ -37,32 +37,42 @@ ColumnReader::ColumnReader(const ColumnDescriptor* descr,
...
-
-  // The dictionary is fully decoded during DictionaryDecoder::Init, so the
-  // DictionaryPage buffer is no longer required after this step
-  //
-  // TODO(wesm): investigate whether this all-or-nothing decoding of the
-  // dictionary makes sense and whether performance can be improved
-
-  auto decoder = std::make_shared<DictionaryDecoder<TYPE> >(descr_);
-  decoder->SetDict(&dictionary);
+  if (encode_type == Encoding::RLE_DICTIONARY) {
```
wesm: This is misleading -- we could check for `Encoding::PLAIN` or `Encoding::PLAIN_DICTIONARY` here -- however we _do_ want to store the decoder in `decoders_` under the `RLE_DICTIONARY` key

src/parquet/column/reader.cc
```diff
@@ -124,12 +134,16 @@ bool TypedColumnReader<TYPE>::ReadNewPage() {
       // first page with this encoding.
       Encoding::type encoding = page->encoding();
 
-      if (IsDictionaryIndexEncoding(encoding)) {
+      if (IsRleDictionaryIndexEncoding(encoding)) {
```
wesm: this could just be `IsDictionaryEncoding` (or the original name is fine too) since either PLAIN_DICTIONARY or RLE_DICTIONARY is valid

src/parquet/column/reader.cc
```diff
@@ -124,12 +134,16 @@ bool TypedColumnReader<TYPE>::ReadNewPage() {
...
         encoding = Encoding::RLE_DICTIONARY;
       }
 
       auto it = decoders_.find(static_cast<int>(encoding));
       if (it != decoders_.end()) {
+        if (encoding == Encoding::RLE_DICTIONARY &&
+            current_decoder_->encoding() != Encoding::RLE_DICTIONARY) {
+            throw ParquetException("Dictionary decoder is not RLE_DICTIONARY.");
+        }
```
wesm: I don't think this is needed if we have enough tests to feel comfortable decoder was correctly constructed -- alternately this can be made a DCHECK

src/parquet/column/test-util.h
```diff
@@ -444,7 +444,8 @@ static int MakePages(const ColumnDescriptor *d, int num_pages, int levels_per_pa
     InitValues<typename Type::c_type>(num_values, values, buffer);
     PaginatePlain<Type>(d, values, def_levels, max_def_level,
         rep_levels, max_rep_level, levels_per_page, values_per_page, pages);
-  } else if (encoding == Encoding::RLE_DICTIONARY) {
+  } else if (encoding == Encoding::RLE_DICTIONARY
+      || encoding == Encoding::PLAIN_DICTIONARY) {
```
wesm: The case that is not checked (unless I missed it) is

- Dictionary page: `Encoding::PLAIN`
- Data pages: `Encoding::RLE_DICTIONARY`

src/parquet/column/scanner-test.cc
```diff
@@ -217,6 +217,12 @@ TEST_F(TestFLBAFlatScanner, TestDictScanner) {
       Encoding::RLE_DICTIONARY);
 }
 
+TEST_F(TestFLBAFlatScanner, TestPlainDictScanner) {
+  this->ExecuteAll(num_pages, num_levels_per_page, batch_size, FLBA_LENGTH,
+      Encoding::PLAIN_DICTIONARY);
```
majetideepak: Test for PLAIN_DICTIONARY.  I will add another in the column reader.

src/parquet/column/test-util.h
majetideepak: Added this test

majetideepak:
@wesm please let me know if you have other comments. Thanks!

wesm:
I will review first thing in the morning, thank you

src/parquet/column/column-reader-test.cc
```diff
@@ -155,5 +155,70 @@ TEST_F(TestPrimitiveReader, TestInt32FlatRepeated) {
   ExecuteDict(num_pages, levels_per_page, &descr);
 }
 
+TEST_F(TestPrimitiveReader, TestPageReader) {
```
wesm: Can you change this test name to something descriptive, like "DictionaryEncodedPages"

src/parquet/column/reader.cc
```diff
@@ -37,26 +37,35 @@ ColumnReader::ColumnReader(const ColumnDescriptor* descr,
...
+    // DictionaryPage buffer is no longer required after this step
+    //
+    // TODO(wesm): investigate whether this all-or-nothing decoding of the
+    // dictionary makes sense and whether performance can be improved
+
+    auto decoder = std::make_shared<DictionaryDecoder<TYPE> >(descr_);
+    decoder->SetDict(&dictionary);
+    decoders_[encoding] = decoder;
+  } else {
+    throw ParquetException("Only RLE_DICTIONARY is supported.");
```
wesm: Maybe make this `ParquetException::NYI("only plain dictionary encoding has been implemented")`

wesm:
Thanks -- I can merge with these minor fixes

majetideepak:
Thank you!

PR #74, opened 2016-03-06
PARQUET-488: Add SSE cmake toggle, fix build on systems without SSE

wesm:
I added an option to make SSE strictly opt-in for now. As a side effect of this, parquet-cpp now builds and the test suite passes out of the box on 32-bit ARMv7 (I tried it on my RaspberryPi Model B 2).

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -64,6 +64,9 @@ string (TOLOWER ${CMAKE_BUILD_TYPE} BUILD_SUBDIR_NAME)
 
 # Top level cmake file, set options
 if ("${CMAKE_SOURCE_DIR}" STREQUAL "${CMAKE_CURRENT_SOURCE_DIR}")
+  option(PARQUET_USE_SSE
+	"Build with SSE4 optimizations"
+	OFF)
   option(PARQUET_BUILD_TESTS
 	"Build the libparquet test suite"
 	ON)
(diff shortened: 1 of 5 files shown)


wesm:
@asandryh @majetideepak can you take a look? Thank you

CMakeLists.txt
```diff
@@ -244,7 +247,12 @@ message(STATUS "Build Type: ${CMAKE_BUILD_TYPE}")
 
 # Build with C++11 and SSE3 by default
 # TODO(wesm): These compiler warning suppressions should be removed one by one
-SET(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -std=c++11 -msse3 -Wall")
+SET(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -std=c++11 -Wall")
+
+if (PARQUET_USE_SSE)
+  SET(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -msse4.2")
```
asandryh: Instead of hard-coding this parameter, it is better to use `-march=native`.

CMakeLists.txt
```diff
@@ -244,7 +247,12 @@ message(STATUS "Build Type: ${CMAKE_BUILD_TYPE}")
 
 # Build with C++11 and SSE3 by default
```
asandryh: The comment about SSE3 does not reflect the compiler flag.

CMakeLists.txt
wesm: Users (me, for example) may need to build portable binaries -- my understanding is that `-march=native` can cause problems for that use case (but I don't know for sure)? My intent here was to build with SSE instructions that can be optionally disabled at runtime (using `CpuInfo`) -- this is what Impala does to generate portable x86 binaries that use SSE4 on machines that have it but not requiring separate binaries for earlier CPUs.

asandryh: AFAIR, `-march=native` is more general: it checks what SSE are supported and enables corresponding flags (such as `-msse`, `-msse2`, etc.).
Anyway, if you are interested specifically in SSE4, then `-msse4.2` does the job.

wesm: I'll ask around and see if I can learn more

wesm:
@asandryh I wasn't able to get a clear picture of what compiler arg makes the most sense. There is also https://gcc.gnu.org/wiki/FunctionMultiVersioning, but the downside of this approach is that it disables inlining. We should probably institute some microbenchmarks before going down that route

wesm:
@asandryh is this OK to merge? thank you

asandryh:
Oops, didn't see this PR was updated. Yes, looks good.

PR #75, opened 2016-03-06
PARQUET-558: Add zsh support to build script

xhochy:


wesm:
+1. if we end up writing a lot more bash/zsh compatibility code we can perhaps factor some of this out into utility modules

PR #76, opened 2016-03-09
PARQUET-556:Extend RowGroupStatistics to include "min" "max" statistics 

majetideepak:
Also includes a patch to extend the GroupReader API with num_rows()

### DIFF
--- a/src/parquet/file/reader.h
+++ b/src/parquet/file/reader.h
@@ -34,6 +34,8 @@ struct RowGroupStatistics {
   int64_t num_values;
   int64_t null_count;
   int64_t distinct_count;
+  std::string min;
+  std::string max;
 };
 
 class RowGroupReader {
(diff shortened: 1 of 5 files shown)


majetideepak:
@wesm let me know your feedback. Thanks!

src/parquet/file/reader.h
```diff
@@ -34,6 +34,8 @@ struct RowGroupStatistics {
   int64_t num_values;
   int64_t null_count;
   int64_t distinct_count;
+  std::string min;
+  std::string max;
```
wesm: should these perhaps be `const std::string*`? The questions is whether 

* the `RowGroupStatistics` might have a lifetime independent of the deserialized file metadata or
* copying the min/max is viewed as expensive

In general it's probably not worth worrying too much about large page statistics but curious what you think

majetideepak: I think your suggestion to make it `const std::string*` makes more sense. We can document that the `RowGroupStatistics` has to be deep copied if it is required past the lifetime of the `RowGroupReader`.

wesm: Sounds good. Unit testing this might be too onerous since we cannot write fully-formed files yet. I hope to find some time to build out the write path later this month.

wesm:
+1, thank you

PR #77, opened 2016-03-12
PARQUET-561: Add destructor to PIMPL

xhochy:
Without this destuctor, the SerializedFile class cannot be destructed and leaks memory.

### DIFF
--- a/src/parquet/file/reader.h
+++ b/src/parquet/file/reader.h
@@ -69,6 +69,7 @@ class ParquetFileReader {
  public:
   // Forward declare the PIMPL
   struct Contents {
+    virtual ~Contents() {}
     // Perform any cleanup associated with the file contents
     virtual void Close() = 0;
 


xhochy:
Two small notes:
 * Could not create an corresponding issue, Apache JIRA fails for me with Error 502
 * The README section about memory leak detection with ctest is wrong. With that command you actually check ctest for memory leaks, not parquet-cpp. There is an option ` -T MemCheck` to run tests with Valgrind but that will then check `build-support/run-test.sh` for memory leaks.

wesm:
@xhochy thanks for the note re: ctest/valgrind. Can you fix the `.travis.yml` in this patch to make the correct invocation on Linux? Presumably this detects the memory leak, then?

wesm:
Also, is there any way to detect this resource management issue with unit tests? I thought we had fixed this in https://github.com/apache/parquet-cpp/commit/46e40e20e7afa1fe493ea927d5a9f689be71dbf8

wesm:
Also, can you open a JIRA and prefix your PR title with "PARQUET-XXX:"? Thank you

xhochy:
@wesm Updated the title to included the opened JIRA ticket. Fixing `.travis.yml` or the unit tests in general won't be easy as we cannot use `build-support/run-test.sh` together with CTests memory checking automatisms. We would either need to get rid of `build-support/run-test.sh` or introduce our own memory check magic to this script.

wesm:
Got it. I will open a JIRA to add a memcheck option to `run-test.sh` and do that. We should also get ASAN working at some point.

wesm:
+1, I verified that the memory leaks are gone (I see them on master) with `$ valgrind --tool=memcheck --leak-check=full ./debug/reader-test`

PR #78, opened 2016-03-13
PARQUET-562: Simplified ZSH support in build scripts

xhochy:
The original proposed patch can be implemented in a much simpler way.
The complicated approach is only needed when used inside a function.

Also some scripts in thirdparty did not support zsh which this patch
adds.


wesm:
+1, thank you

PR #79, opened 2016-03-13
PARQUET-564: Add cmake option to run valgrind on each unit test executable

wesm:
This also enables the current test suite to pass cleanly under `ctest` with this option enabled.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -132,6 +132,8 @@ function(ADD_PARQUET_TEST REL_TEST_NAME)
   endif()
 
   if (PARQUET_TEST_MEMCHECK)
+	SET_TARGET_PROPERTIES(${TEST_NAME}
+	  PROPERTIES COMPILE_FLAGS -DPARQUET_VALGRIND)
 	add_test(${TEST_NAME}
       valgrind --tool=memcheck --leak-check=full --error-exitcode=1 ${TEST_PATH})
   else()
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -25,13 +25,6 @@ if (NOT "$ENV{PARQUET_GCC_ROOT}" STREQUAL "")
   set(CMAKE_CXX_COMPILER ${GCC_ROOT}/bin/g++)
 endif()
 
-if ("$ENV{PARQUET_USE_CCACHE}")
-  set(CMAKE_C_COMPILER_ARG1 ${CMAKE_C_COMPILER})
-  set(CMAKE_CXX_COMPILER_ARG1 ${CMAKE_CXX_COMPILER})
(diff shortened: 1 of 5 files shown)


wesm:
cc @xhochy @asandryh for review

CMakeLists.txt
```diff
@@ -128,8 +131,15 @@ function(ADD_PARQUET_TEST REL_TEST_NAME)
     set(TEST_PATH ${CMAKE_CURRENT_SOURCE_DIR}/${REL_TEST_NAME})
   endif()
 
-  add_test(${TEST_NAME}
-    ${BUILD_SUPPORT_DIR}/run-test.sh ${TEST_PATH})
+  if (PARQUET_TEST_MEMCHECK)
+	SET_TARGET_PROPERTIES(${TEST_NAME}
```
xhochy: This does not work on Linux as it will override all previously set target properties and thus compiling will fail due to the wrong `tr1/tuple` handling. The following line worked for me and seems to be the common way to append to target properties (note the explicit space between `"` and `-D`):

```
SET_PROPERTY(TARGET ${TEST_NAME} APPEND_STRING PROPERTY COMPILE_FLAGS " -DPARQUET_VALGRIND")
```

CMakeLists.txt
```diff
@@ -128,8 +131,15 @@ function(ADD_PARQUET_TEST REL_TEST_NAME)
...
+	  PROPERTIES COMPILE_FLAGS -DPARQUET_VALGRIND)
+	add_test(${TEST_NAME}
+      valgrind --tool=memcheck --leak-check=full --error-exitcode=1 ${TEST_PATH})
```
xhochy: You could also `include(CTest)` and then use `${MEMORYCHECK_COMMAND}` which will resolve to the installation of `valgrind`.

CMakeLists.txt
wesm: Cool, thank you, will remember this for the future

CMakeLists.txt
wesm: Cool, I will check that this substitution does what we want

wesm:
@xhochy can you take a look and let me know if this looks good?

I tried using `${MEMORYCHECK_COMMAND}` but it does not report leaks as errors (`--error-exitcode`), so I left the explicit valgrind call as is.

xhochy:
Looking good! `${MEMORYCHECK_COMMAND}` just evaluates to `/usr/bin/valgrind` for me. As it is a very unlikely case that someone wants to use a valgrind installation that is not in `PATH`, we probably should not care about `${MEMORYCHECK_COMMAND}`.

wesm:
thank you.

PR #80, opened 2016-03-16
PARQUET-565: Use PATH instead of DIRECTORY in get_filename_component

xhochy:
Fixes support for CMake<=2.8.11


wesm:
Thank you. Do you get the sense that supporting older cmake versions is worthwhile? I don't know what kind of bugs have been fixed or other divergence between >= 3.2.x and old versions.

xhochy:
I would support it for now as long as we support Linux distributions that old (i.e. ones that ship with gcc 4.7+). As long as we do not need any special features, it should work. Otherwise there are always ways to get a newer CMake version but that would make complicated to build parquet-cpp initially on a slightly older system. (All under the assumption that there are not serious problems in CMake < 3.0).

PR #81, opened 2016-03-21
PARQUET-568: Enable top-level column selection.

asandryh:


asandryh:
@wesm Can you review this patch? Thanks.

wesm:
Can you modify the JIRA to reflect this this is limited only to the `DebugPrint` method? I'll review the code now but I don't believe we should be continuing to invest energy in this method vs. a more versatile "ParquetFilePrinter"

src/parquet/file/reader.cc
```diff
@@ -133,11 +133,23 @@ std::shared_ptr<RowGroupReader> ParquetFileReader::RowGroup(int i) {
...
 
-void ParquetFileReader::DebugPrint(std::ostream& stream, bool print_values) {
+void ParquetFileReader::DebugPrint(std::ostream& stream,
+    std::list<int> columns, bool print_values) {
   stream << "File statistics:\n";
-  stream << "Total rows: " << this->num_rows() << "\n";
+  stream << "Total rows: " << num_rows() << "\n";
 
-  for (int i = 0; i < num_columns(); ++i) {
+  // Verify correctness of selected columns
```
wesm: `TODO(asandryh)` here?

src/parquet/file/reader.cc
```diff
@@ -133,11 +133,23 @@ std::shared_ptr<RowGroupReader> ParquetFileReader::RowGroup(int i) {
...
+  if (columns.size() == 0) {
+    for (int i = 0; i < num_columns(); i++) {
+      columns.push_back(i);
+    }
+  } else {
+    int n_cols = num_columns();
+    auto out_of_range = [n_cols](int i)->bool{ return i < 0 || i >= n_cols; };
+    columns.remove_if(out_of_range);
```
wesm: Throw an exception for out of range `i`?

src/parquet/file/reader.cc
```diff
@@ -133,11 +133,23 @@ std::shared_ptr<RowGroupReader> ParquetFileReader::RowGroup(int i) {
 // the fixed initial size is just for an example
 #define COL_WIDTH "20"
 
-void ParquetFileReader::DebugPrint(std::ostream& stream, bool print_values) {
+void ParquetFileReader::DebugPrint(std::ostream& stream,
+    std::list<int> columns, bool print_values) {
```
wesm: Better to call this variable `column_indices` or something more descriptive

wesm:
This is OK to facilitate further debugging, but we need to be able to test this code's correctness if we continue any further down this path.

asandryh:
@wesm Thank you for the comments, I've incorporated them into the PR and update the JIRA.
You're correct, this functionality is primarily to facilitate debugging. It will also be useful after `DebugPrint` is factored out of `ParquetFileReader` in PARQUET-508.

wesm:
+1, thank you

PR #82, opened 2016-03-21
PARQUET-566: Add method to retrieve the full column path

xhochy:
This is based on the idea of `org.apache.parquet.hadoop.metadata.ColumnPath`.

### DIFF
--- a/src/parquet/column/column-reader-test.cc
+++ b/src/parquet/column/column-reader-test.cc
@@ -39,7 +39,6 @@ using std::shared_ptr;
 
 namespace parquet_cpp {
 
-using schema::ColumnPath;
 using schema::NodePtr;
 
 namespace test {
@@ -129,8 +128,7 @@ TEST_F(TestPrimitiveReader, TestInt32FlatRequired) {
   max_def_level_ = 0;
   max_rep_level_ = 0;
   NodePtr type = schema::Int32("a", Repetition::REQUIRED);
-  const ColumnDescriptor descr(type, max_def_level_, max_rep_level_,
-      ColumnPath::FromDotString("a"));
+  const ColumnDescriptor descr(type, max_def_level_, max_rep_level_);
   ExecutePlain(num_pages, levels_per_page, &descr);
   ExecuteDict(num_pages, levels_per_page, &descr);
 }
@@ -141,8 +139,7 @@ TEST_F(TestPrimitiveReader, TestInt32FlatOptional) {
   max_def_level_ = 4;
   max_rep_level_ = 0;
   NodePtr type = schema::Int32("b", Repetition::OPTIONAL);
-  const ColumnDescriptor descr(type, max_def_level_, max_rep_level_,
-      ColumnPath::FromDotString("b"));
+  const ColumnDescriptor descr(type, max_def_level_, max_rep_level_);
   ExecutePlain(num_pages, levels_per_page, &descr);
   ExecuteDict(num_pages, levels_per_page, &descr);
 }
@@ -153,8 +150,7 @@ TEST_F(TestPrimitiveReader, TestInt32FlatRepeated) {
   max_def_level_ = 4;
   max_rep_level_ = 2;
   NodePtr type = schema::Int32("c", Repetition::REPEATED);
-  const ColumnDescriptor descr(type, max_def_level_, max_rep_level_,
-      ColumnPath::FromDotString("c"));
+  const ColumnDescriptor descr(type, max_def_level_, max_rep_level_);
   ExecutePlain(num_pages, levels_per_page, &descr);
   ExecuteDict(num_pages, levels_per_page, &descr);
 }
@@ -163,8 +159,7 @@ TEST_F(TestPrimitiveReader, TestDictionaryEncodedPages) {
   max_def_level_ = 0;
   max_rep_level_ = 0;
   NodePtr type = schema::Int32("a", Repetition::REQUIRED);
-  const ColumnDescriptor descr(type, max_def_level_, max_rep_level_,
-      ColumnPath::FromDotString("a"));
+  const ColumnDescriptor descr(type, max_def_level_, max_rep_level_);
   shared_ptr<OwnedMutableBuffer> dummy = std::make_shared<OwnedMutableBuffer>();
 
   shared_ptr<DictionaryPage> dict_page = std::make_shared<DictionaryPage>(dummy,
--- a/src/parquet/column/column-reader-test.cc
+++ b/src/parquet/column/column-reader-test.cc
@@ -39,6 +39,7 @@ using std::shared_ptr;
 
 namespace parquet_cpp {
 
+using schema::ColumnPath;
 using schema::NodePtr;
 
 namespace test {
@@ -128,7 +129,8 @@ TEST_F(TestPrimitiveReader, TestInt32FlatRequired) {
   max_def_level_ = 0;
   max_rep_level_ = 0;
   NodePtr type = schema::Int32("a", Repetition::REQUIRED);
-  const ColumnDescriptor descr(type, max_def_level_, max_rep_level_);
+  const ColumnDescriptor descr(type, max_def_level_, max_rep_level_,
+      ColumnPath::FromDotString("a"));
   ExecutePlain(num_pages, levels_per_page, &descr);
   ExecuteDict(num_pages, levels_per_page, &descr);
 }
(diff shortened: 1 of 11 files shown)


xhochy:
Travis fails due to 

```/home/travis/build/apache/parquet-cpp/src/parquet/schema/types.h:85:  RValue references are an unapproved C++ feature.  [build/c++11] [3]````

In the context of parquet-cpp: Why don't we want to use this feature?

File: src/parquet/schema/types.h [hunk 1]
```diff
@@ -78,6 +78,21 @@ struct DecimalMetadata {
...
+class ColumnPath {
+ public:
+  ColumnPath() : path_() {}
+  explicit ColumnPath(const std::vector<std::string>& path) : path_(path) {}
+  explicit ColumnPath(std::vector<std::string>&& path) : path_(path) {}
+
+  static std::shared_ptr<ColumnPath> FromDotString(const std::string& dotstring);
+
+  std::shared_ptr<ColumnPath> extend(const std::string& node_name) const;
+  std::string toDotString() const;
```
wesm: Make this method `ToDotString` per the Google naming conventions

src/parquet/schema/types.cc
```diff
@@ -28,6 +29,36 @@ namespace parquet_cpp {
...
+  path[path_.size()] = node_name;
+
+  return std::shared_ptr<ColumnPath>(new ColumnPath(std::move(path)));
+}
+
+std::string ColumnPath::toDotString() const {
+  return std::accumulate(path_.cbegin(), path_.cend(), std::string(),
+      [](const std::string &str1, const std::string &str2) {
+          return str1.empty() ? str2 : str1 + "." + str2;
+      });
```
wesm: It's nice to use the fancy stuff now and again, but I think a for loop would be clearer here:

```c++
std:stringstream ss;
for (const std::string& element : path_) {
  if (!result.empty()) {
    ss << ".";
  }
  ss << element;
}
return ss.str()
```

wesm:
I think RValue references to prevent copies are fine; do you want to remove that check from cpplint? 

The one high level question here is that this does not assist with the problem of examining the intermediate nodes for purposes of record or column reassemble. I had thought that we could pass in `const Node* parent` (with default nullptr) in node construction to reflect parent/child relationships, and column paths could be reconstructed by walking the links from child to parent until hitting the root (`parent=nullptr`). What do you think?

xhochy:
With the current way to build the node tree, it is hard to inject the parent nodes. We build the tree in `src/parquet/schema/converter.cc` bottom-up. To add parent references, we would need a second pass over the node tree and a method to inject parent nodes. This method could either be public or the `FlatSchemaConverter` a `friend` of the `Node` class. Both solutions sound suboptimal for me.

xhochy:
Concerning the cpplint check: The easy way would be to disable the C++11 checks (as done in https://codereview.chromium.org/1710293002/ and https://codereview.chromium.org/1573663003/) but they would also disable more than just the RValueReference one. The alternative is to patch cpplint.py on our own. I would prefer the first solution.

wesm:
Fine with me to remove the C++11 checks. Let me look at the schema converter code more closely

wesm:
So the quickest solution I can see (which may not work, didn't try it!) would be to add a protected `set_parent` method to `Node` and calling it in `GroupNode::FromParquet` to set the parent-child relationships. Let me know what you think.

wesm:
Or better, you could do this in the `GroupNode` ctor. The downside of this is that you're mutating the nodes, so as long as we indicate clearly that node ownership is not to be shared (so that two group nodes don't accidentally step on each other's parent pointer) this is probably OK

xhochy:
Sadly you cannot modify or call protected members of a third object that is derived from the same base class but not the currently calling class. Thus you could only set parents on GroupNodes from GroupNodes.

xhochy:
From all possible alternative (like making `set_parent` public or so), I would prefer make `GroupNode` a friend class of Node. Thus it could access the protected methods of other child classes and we do not need a public method to change the parent node.

wesm:
Sorry yes you have to make Node a friend of GroupNode indeed -- that is OK with me.

wesm:
Sorry for the review delay, I was traveling this week. Looking now

src/parquet/column/column-reader-test.cc
```diff
@@ -128,7 +129,8 @@ TEST_F(TestPrimitiveReader, TestInt32FlatRequired) {
   max_def_level_ = 0;
   max_rep_level_ = 0;
   NodePtr type = schema::Int32("a", Repetition::REQUIRED);
-  const ColumnDescriptor descr(type, max_def_level_, max_rep_level_);
+  const ColumnDescriptor descr(type, max_def_level_, max_rep_level_,
+      ColumnPath::FromDotString("a"));
```
wesm: Now that we have the `parent` pointers, we don't need to construct the column path at construction time, but we can always build it by following the parent links.

src/parquet/schema/descriptor.cc
```diff
@@ -58,19 +61,23 @@ void SchemaDescriptor::BuildTree(const NodePtr& node, int16_t max_def_level,
...
+    std::shared_ptr<ColumnPath> group_path = column_path->extend(group->name());
     for (int i = 0; i < group->field_count(); ++i) {
-      BuildTree(group->field(i), max_def_level, max_rep_level);
+      BuildTree(group->field(i), max_def_level, max_rep_level, group_path);
     }
   } else {
     // Primitive node, append to leaves
-    leaves_.push_back(ColumnDescriptor(node, max_def_level, max_rep_level, this));
+    leaves_.push_back(ColumnDescriptor(node, max_def_level, max_rep_level,
+      column_path->extend(node->name()), this));
```
wesm: My preference is to add a `GetColumnPath` to `ColumnDescriptor` and build this on demand.

src/parquet/schema/descriptor.cc
```diff
@@ -58,19 +61,23 @@ void SchemaDescriptor::BuildTree(const NodePtr& node, int16_t max_def_level,
...
   }
 }
 
 ColumnDescriptor::ColumnDescriptor(const schema::NodePtr& node,
     int16_t max_definition_level, int16_t max_repetition_level,
+    const std::shared_ptr<ColumnPath>& column_path,
     const SchemaDescriptor* schema_descr) :
       node_(node),
+      path_(column_path),
```
wesm: In that case, these can be removed

src/parquet/schema/schema-converter-test.cc
```diff
@@ -62,6 +62,23 @@ class TestSchemaConverter : public ::testing::Test {
   std::unique_ptr<Node> node_;
 };
 
+bool check_for_parent_consistency(const GroupNode* node) {
```
wesm: nit: `CheckForParentConsistency` (or something like that)

src/parquet/schema/schema-descriptor-test.cc
```diff
@@ -51,7 +54,7 @@ TEST(TestColumnDescriptor, TestAttrs) {
   // Test FIXED_LEN_BYTE_ARRAY
   node = PrimitiveNode::Make("name", Repetition::OPTIONAL,
       Type::FIXED_LEN_BYTE_ARRAY, LogicalType::DECIMAL, 12, 10, 4);
-  descr = ColumnDescriptor(node, 4, 1);
+  descr = ColumnDescriptor(node, 4, 1, ColumnPath::FromDotString("name"));
```
wesm: Can remove these column paths per the above notes

src/parquet/schema/types.cc
```diff
@@ -37,6 +72,10 @@ bool Node::EqualsInternal(const Node* other) const {
     logical_type_ == other->logical_type_;
 }
 
+void Node::SetParent(const Node* p_parent) {
```
wesm: `parent` instead of `p_parent`

src/parquet/schema/types.h
```diff
@@ -78,9 +78,28 @@ struct DecimalMetadata {
...
+ protected:
+  std::vector<std::string> path_;
+};
+
+class GroupNode;
+
 // Base class for logical schema types. A type has a name, repetition level,
 // and optionally a logical type (ConvertedType in Parquet metadata parlance)
 class Node {
+  friend class GroupNode;
```
wesm: Put this friend down in the `protected:` block

wesm:
I'm not blocked on this patch for ARROW-22 so no urgency to address these comments. Sorry again for the review delay

wesm:
+1, thank you!

PR #83, opened 2016-03-21
PARQUET-559: Enable external RandomAccessSource as input to the ParquetFileReader

majetideepak:


wesm:
I'm a bit confused. Semantically, one of the essential defining characteristics of a "stream" is that you cannot seek. The Parquet format supports streaming writes (you accumulate metadata then write it at the end of the file) but only random access reads. How is `ExternalInputStream` different from `RandomAccessSource`?

src/parquet/reader-test.cc
```diff
@@ -36,6 +37,74 @@ namespace parquet_cpp {
...
+
+    auto buffer = std::make_shared<OwnedMutableBuffer>();
+    buffer->Resize(file_size);
+
+    if (!parquet_file.read(reinterpret_cast<char*>(buffer->mutable_data()), file_size)) {
+      throw ParquetException("Could not read file into buffer\n");
+    }
+    parquet_file.close();
+    stream_ = new ExternalInputStreamImpl(buffer);
+    reader_ = ParquetFileReader::OpenStream(stream_);
```
wesm: How is this different from using `BufferReader`?

wesm:
Is what you're after is an API for `RandomAccessReader` that yields an `InputStream` (so not necessarily performing a complete read, only a stream-like object which may or may not have materialized the full request)? So rather than this code:

```c++
  int64_t bytes_to_read = col.meta_data.total_compressed_size;
  std::shared_ptr<Buffer> buffer = source_->ReadAt(col_start, bytes_to_read);

  if (buffer->size() < bytes_to_read) {
    throw ParquetException("Unable to read column chunk data");
  }

  std::unique_ptr<InputStream> stream(new InMemoryInputStream(buffer));
  return std::unique_ptr<PageReader>(new SerializedPageReader(std::move(stream),
          FromThrift(col.meta_data.codec), allocator_));
```

We instead would have

```c++
  int64_t bytes_to_read = col.meta_data.total_compressed_size;
  std::unique_ptr<InputStream> stream = source_->ReadStream(col_start, bytes_to_read);
  return std::unique_ptr<PageReader>(new SerializedPageReader(std::move(stream),
          FromThrift(col.meta_data.codec), allocator_));
```

Does that sound right?

majetideepak:
`ExternalInputStream` API does not seek. The API uses `offset` to read/peek from different parts of the stream. My main goal here is to be able to read Parquet file from a stream and not from disk.  `InputStream` API will work for most part except the verification part which requires jumping to the end of the stream.

majetideepak:
`source_->Seek(filesize - FOOTER_SIZE);`
 `int64_t bytes_read = source_->Read(FOOTER_SIZE, footer_buffer);`
I cannot do this with the `InputStream`

majetideepak:
Is there a way I can see the test log output in travis-ci when a test fails ?
The test I added only fails on Linux gcc and I have no additional information on the failure.

wesm:
Per:

```c++
source_->Seek(filesize - FOOTER_SIZE);
int64_t bytes_read = source_->Read(FOOTER_SIZE, footer_buffer);
```

This is the RandomAccessSource API. I do not see why we need a `StreamSource` when creating a new subclass of `RandomAccessSource` will do just fine.

wesm:
I can see memory leaks here: https://travis-ci.org/apache/parquet-cpp/jobs/117626248#L2516 -- are you able to reproduce these locally?

src/parquet/util/input.h
```diff
@@ -140,6 +140,68 @@ class BufferReader : public RandomAccessSource {
...
+  virtual const char* GetName() {return "InMemoryExternalInputStream";}
+
+ private:
+  std::shared_ptr<Buffer> buffer_;
+  int64_t len_;
+};
+
+// ----------------------------------------------------------------------
+// A stream-like object that reads from an ExternalInputStream
+class StreamSource : public RandomAccessSource {
```
majetideepak: `StreamSource` is a subclass of `RandomAccessSource`.

src/parquet/util/input.h
```diff
@@ -140,6 +140,68 @@ class BufferReader : public RandomAccessSource {
...
+  explicit StreamSource(ExternalInputStream* stream);
+  virtual void Close() {}
+  virtual int64_t Tell() const;
+  virtual void Seek(int64_t pos);
+  virtual int64_t Read(int64_t nbytes, uint8_t* out);
+  virtual std::shared_ptr<Buffer> Read(int64_t nbytes);
+
+ private:
+  // parquet-cpp should not manage this object
+  ExternalInputStream* stream_;
```
majetideepak: However, `StreamSource`  takes in a stream instead of a file or a buffer like object.

src/parquet/file/reader.cc
```diff
@@ -74,6 +74,18 @@ ParquetFileReader::~ParquetFileReader() {
   Close();
 }
 
+std::unique_ptr<ParquetFileReader> ParquetFileReader::OpenStream(
+    ExternalInputStream* stream, MemoryAllocator* allocator) {
```
majetideepak: I want the clients of parquet-cpp to be able to pass a stream as well.

src/parquet/reader-test.cc
majetideepak: I am just emulating a client stream here. Client streams must extend  `ExternalInputStream`

src/parquet/util/input.h
```diff
@@ -140,6 +140,68 @@ class BufferReader : public RandomAccessSource {
...
+  virtual const char* GetName() = 0;
+
+  virtual ~ExternalInputStream() {}
+
+ protected:
+  ExternalInputStream() {}
+};
+
+// Implementation of an ExternalInputStream for testing
+class ExternalInputStreamImpl : public ExternalInputStream {
```
majetideepak: This is used just for testing purposes as of now. I can move this to another location if needed.

src/parquet/file/reader.cc
wesm: From the `ExternalInputStream` abstract class:

```c++
  // Read interface to read from stream at an offset into a specified buffer
   // Returns the number of bytes read.
   virtual int64_t Read(int64_t num_to_read, int64_t offset, uint8_t* buffer) = 0;
```

I may be missing something, but this looks like random access to me. I do not see why `ExternalInputStream` or `StreamSource` are needed if the user can implement their own custom subclass of `RandomAccessSource` in downstream code.

So long as `ParquetFileReader` depends on being able to seek to the end of the file and read the metadata from the footer, all input sources need to support random access. There may be a case where the metadata is read separately from the rest of the file (which is streamed, so seeking is not supported).

wesm:
Maybe it would help if you can explain the use case in some more detail.

You said:

> ExternalInputStream API does not seek. The API uses offset to read/peek from different parts of the
> stream.

Having `offset` as a parameter in these methods presumes that you have some virtual address space or file-like object available, which is not a stream.

majetideepak:
The primary use case is to provide a user API where the downstream code can create a `ParquetFileReader` by passing a `RandomAccessSource`. 
 I used the word `stream` here in the API to represent something more generic like a IO stream, and not exactly streaming data.

majetideepak:
`std::shared_ptr<Buffer> ReadAt(int64_t pos, int64_t nbytes);` in 
`RandomAccessSource` forces the downstream code to include the `Buffer` Api.
 `ExternalInputStream` will provide a cleaner API interface.

wesm:
Okay, I see what you are saying. Currently, we have the following static ctor:

```c++
  static std::unique_ptr<ParquetFileReader> OpenFile(const std::string& path,
      bool memory_map = true, MemoryAllocator* allocator = default_allocator());
```

Let's add one:

```c++
  static std::unique_ptr<ParquetFileReader> Open(
      std::unique_ptr<RandomAccessSource> source,
      MemoryAllocator* allocator = default_allocator());
```

And refactor `ParquetFileReader::OpenFile` to call this after opening the file (so code is not duplicated). I don't think adding a new abstraction is needed. Currently, the implementation of ReadAt is:

```
std::shared_ptr<Buffer> RandomAccessSource::ReadAt(int64_t pos, int64_t nbytes) {
  Seek(pos);
  return Read(nbytes);
}
```

If necessary for your needs, you can make `ReadAt` virtual and invoke the `int64_t Read(int64_t nbytes, uint8_t* out)` API. If there is some other problem I'm not seeing, please let me know.

wesm:
So I think I'm reading between the lines that you are not able to deal with `std::shared_ptr` in non-C++11 code, is that right?

majetideepak:
I agree that `ExternalInputStream` is `RandomAccessSource` like. I modified the patch according to our discussion. Non-C++11 is not a problem here.
On a separate note, do we know why parquet-cpp is an order slow on travis ci for ubuntu gcc config ?

wesm:
Thank you, +1

Travis CI seems always hit and miss, and occasionally they have issues with their GCE package mirror. Things have been much better since we limited Flink's concurrent worker usage. If you point me to any problematic builds I can raise the issue with Travis CI or the ASF if there is something systematic (vs idiosyncratic going on). A lot of the discussions I have with Travis goes like "well we were having some problems but now it's better" =)

PR #84, opened 2016-03-23
PARQUET-552: Estimate peak dynamic memory usage using FileMetadata

asandryh:
The estimate `MemoryUsage.memory` is an upper bound on dynamic memory usage. However, it does not include memory used in buffers during dictionary decoding. Flag `MemoryUsage.has_dictionary` is set to indicate whether dictionary encoding is used.

### DIFF
--- a/src/parquet/reader-test.cc
+++ b/src/parquet/reader-test.cc
@@ -45,13 +45,14 @@ class TestAllTypesPlain : public ::testing::Test {
     std::stringstream ss;
     ss << dir_string << "/" << "alltypes_plain.parquet";
 
-    reader_ = ParquetFileReader::OpenFile(ss.str());
+    reader_ = ParquetFileReader::OpenFile(ss.str(), false, &allocator_);
   }
 
   void TearDown() {}
 
  protected:
   std::unique_ptr<ParquetFileReader> reader_;
+  TrackingAllocator allocator_;
 };
 
 TEST_F(TestAllTypesPlain, NoopConstructDestruct) {
@@ -125,7 +126,7 @@ TEST_F(TestAllTypesPlain, DebugPrintWorks) {
   std::stringstream ss;
 
   std::list<int> columns;
-  reader_->DebugPrint(ss, columns);
+  reader_->DebugPrint(ss, columns, 128);
 
   std::string result = ss.str();
   ASSERT_GT(result.size(), 0);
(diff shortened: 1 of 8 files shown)


asandryh:
@wesm Can you take a look at this PR?

src/parquet/reader-test.cc
```diff
@@ -149,13 +150,43 @@ TEST_F(TestAllTypesPlain, ColumnSelectionOutOfRange) {
...
+  ASSERT_TRUE(memory_usage2.has_dictionary);
+  ASSERT_GT(memory_usage2.memory, memory_usage1.memory);
+
+  columns.clear();
+  ParquetFileReader::MemoryUsage memory_usage_all
+      = reader_->EstimateMemoryUsage(false, columns, batch_size);
+  reader_->DebugPrint(ss, columns, batch_size, true);
+  ASSERT_GT(allocator_.MaxMemory(), memory_usage_all.memory);
+  ASSERT_TRUE(memory_usage_all.has_dictionary);
+  ASSERT_GT(memory_usage_all.memory, memory_usage2.memory);
```
wesm: I believe we are at a point where creating faux metadata and a faux serialized file for testing purposes is a prudent step rather than continuing to rely on a static pre-generated binary file, otherwise it is difficult to precisely test some code path in the memory estimation. What do you think?

asandryh: Agree. I've tested memory estimation on all different combinations of (column type, encoding, compression) offline on a bunch of pre-generated Parquet file, so I'm quite confident in this functionality. But it would be nice to automate and incorporate these tests via faux structures. Do you think this should be done within this patch, or it's big enough to have a separate jira?

wesm: One way or another we need an internal API for constructing the Thrift metadata and creating the serialized file footer at least for testing purposes but also for writing Parquet files. The problem with ad hoc testing is that it doesn't protect you from bugs introduced by refactoring.

Without more comprehensive unit tests, this feature cannot be considered complete -- I am indifferent about whether it gets done in this JIRA or in a follow up JIRA, but I also don't want these tests to be thought of as "nice to have" rather than necessary.

wesm:
I just merged #85 (parquet_cpp -> parquet) -- you may not even have to rebase but letting you know just in case.

asandryh:
I have realized that a big part of the unit test for this patch is actually **confirming** that the estimate is an upper bound. It is not enough to have faux file footers - we need actual files, since we also allocate (and estimate for) memory for reading, e.g. in `LocalFileSource`. Thus, writing parquet files is a pre-req for this patch.
I will keep this PR open for now. Once we can write parquet, I'll update this PR with appropriate tests.

wesm:
Thanks and very good points. I'm going to be working on file writing here in the next week or so -- will keep you posted.

asandryh:
@majetideepak and I have lately made some changes to memory estimation that make this PR obsolete. Will open a new PR once we can write Parquet files.

PR #85, opened 2016-03-27
PARQUET-572: Rename public namespace to parquet from parquet_cpp

wesm:
cc @asandryh @majetideepak let me know if this would be problematic in some way. Thank you


majetideepak:
This shouldn't be a problem.
+1 LGTM

wesm:
Thank you

PR #86, opened 2016-04-16
PARQUET-582: Conversions functions for Parquet enums to Thrift enums

xhochy:

PR #87, opened 2016-04-16
PARQUET-583: Parquet to Thrift schema conversion

xhochy:
Depends on #86

### DIFF
--- a/src/parquet/schema/converter.cc
+++ b/src/parquet/schema/converter.cc
@@ -83,6 +83,45 @@ std::shared_ptr<SchemaDescriptor> FromParquet(const std::vector<SchemaElement>& 
   return descr;
 }
 
+void ToParquet(const GroupNode* schema, std::vector<format::SchemaElement>* out) {
+  SchemaFlattener flattener(schema, out);
+  flattener.Flatten();
+}
+
+class SchemaVisitor : public Node::ConstVisitor {
+ public:
+  explicit SchemaVisitor(std::vector<format::SchemaElement>* elements)
+     : elements_(elements) {}
+  virtual ~SchemaVisitor() {}
+
+  void Visit(const Node* node) override {
+    format::SchemaElement element;
+    node->ToParquet(&element);
+    // Override field_id here as we can get user-generated Nodes without a valid id
+    element.__set_field_id(elements_->size());
+    elements_->push_back(element);
+
+    if (node->is_group()) {
+      const GroupNode* group_node = static_cast<const GroupNode*>(node);
+      for (int i = 0; i < group_node->field_count(); ++i) {
+        group_node->field(i)->VisitConst(this);
+      }
+    }
+  }
+
+ private:
+  std::vector<format::SchemaElement>* elements_;
+};
+
+SchemaFlattener::SchemaFlattener(const GroupNode* schema,
+        std::vector<format::SchemaElement>* out)
+    : root_(schema), elements_(out) {}
+
+void SchemaFlattener::Flatten() {
+  SchemaVisitor visitor(elements_);
+  root_->VisitConst(&visitor);
+}
+
 } // namespace schema
 
 } // namespace parquet
--- a/src/parquet/schema/test-util.h
+++ b/src/parquet/schema/test-util.h
@@ -36,7 +36,7 @@ namespace parquet {
 namespace schema {
 
 static inline SchemaElement NewPrimitive(const std::string& name,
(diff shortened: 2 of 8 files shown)


src/parquet/schema/types.h
```diff
@@ -163,15 +163,26 @@ class Node {
...
+  // ToParquet returns an opaque void* to avoid exporting
+  // parquet::SchemaElement into the public API
+  virtual void ToParquet(void* opaque_element) const = 0;
+
   // Node::Visitor abstract class for walking schemas with the visitor pattern
   class Visitor {
    public:
     virtual ~Visitor() {}
 
     virtual void Visit(const Node* node) = 0;
```
wesm: I suppose this should be made `Visit(Node* node)` to differentiate with `ConstVisitor`?

src/parquet/schema/types.cc
```diff
@@ -280,6 +288,35 @@ std::unique_ptr<Node> PrimitiveNode::FromParquet(const void* opaque_element,
...
+  element->__set_num_children(0);
+  element->__set_repetition_type(ToThrift(repetition_));
+  if (logical_type_ != LogicalType::NONE) {
+    element->__set_converted_type(ToThrift(logical_type_));
+  }
+  element->__set_type(ToThrift(physical_type_));
+  // FIXME: SchemaFlattener does this for us: element->__set_field_id(id_);
+  element->__set_type_length(type_length_);
+  element->__set_precision(decimal_metadata_.precision);
+  element->__set_scale(decimal_metadata_.scale);
```
wesm: Only set these for DECIMAL annotation?

wesm: I looked any on `LogicalType::None` these have been set to -1. I'm not sure which is better; not setting them or setting them to -1 (I haven't looked to see what parquet-mr does)

src/parquet/schema/test-util.h
```diff
@@ -36,22 +36,28 @@ namespace parquet {
 namespace schema {
 
 static inline SchemaElement NewPrimitive(const std::string& name,
-    FieldRepetitionType::type repetition, format::Type::type type) {
+    FieldRepetitionType::type repetition, format::Type::type type, int id) {
```
wesm: You could set the default id to 0 and below in NewGroup, too.

src/parquet/schema/converter.cc
```diff
@@ -83,6 +83,45 @@ std::shared_ptr<SchemaDescriptor> FromParquet(const std::vector<SchemaElement>&
...
+
+  void Visit(const Node* node) override {
+    format::SchemaElement element;
+    node->ToParquet(&element);
+    // Override field_id here as we can get user-generated Nodes without a valid id
+    element.__set_field_id(elements_->size());
+    elements_->push_back(element);
+
+    if (node->is_group()) {
+      const GroupNode* group_node = static_cast<const GroupNode*>(node);
```
wesm: for future cases, I believe it's OK to use `auto` in conjunction with `static_cast<const T*>`

wesm:
This looks good to me outside of the minor comments. The way we're handling the not-always-set `SchemaElement` fields bums me out a bit, but if they can be counted on to always be -1 if not set it does make testing easier. Pretty open-minded about improving this aspect of the schema code generally

src/parquet/schema/types.cc
xhochy: `parquet-mr` does set the decimal metadata independent of the logical type. But in the case of parquet-mr, `decimal_metadata_` can be `null`. In these cases precision and scale are not set on the Thrift object. (See `parquet-hadoop/src/main/java/org/apache/parquet/format/converter/ParquetMetadataConverter.java:122` )

xhochy:
For the not-always-set `SchemaElement` fields, we probably need a way in `SchemaElement` to track if they were set (which we could also define as `value == -1`). Then we could decide consistently in the to-thrift-schema-conversion if we should write them out.

wesm:
If the struct is zeroed out, you can use the  element.__isset for this

xhochy:
Yes, we have that in Thrift. Probably I got it wrong in my last comment ;)  What I'm missing is that we haven't got something similar to `.__isset` in `PrimitiveNode` to track if we have set one of `length, precision, scale`.

But to get this PR to a close: Should I only call `element->__set_type_length` if `length != -1` or keep the code for as it is? (this is the only remaining issue here or am I missing something?)

wesm:
Keeping things as is works for me right now, we can iterate on the optional field stuff later.

wesm:
+1, merging

PR #88, opened 2016-04-20
PARQUET-587: Implement BufferReader::Read(int64_t,uint8_t*)

xhochy:
* Add unit test for BufferReader

### DIFF
--- a/src/parquet/util/input-output-test.cc
+++ b/src/parquet/util/input-output-test.cc
@@ -56,7 +56,7 @@ TEST(TestBufferedReader, Basics) {
   BufferReader reader(buffer);
 
   uint8_t out[4];
-  reader.Read(4, out);
+  ASSERT_EQ(4, reader.Read(4, out));
   ASSERT_EQ(4, reader.Tell());
   ASSERT_EQ(0, out[0]);
   ASSERT_EQ(1, out[1]);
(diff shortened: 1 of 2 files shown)


src/parquet/util/input-output-test.cc
```diff
@@ -50,6 +50,31 @@ TEST(TestInMemoryOutputStream, Basics) {
   ASSERT_TRUE(data_buf.Equals(*buffer));
 }
 
+TEST(TestBufferedReader, Basics) {
+  std::vector<uint8_t> data = {0, 1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 11, 12};
+  auto buffer = std::make_shared<Buffer>(data.data(), data.size());
+  BufferReader reader(buffer);
+
+  uint8_t out[4];
+  reader.Read(4, out);
```
wesm: Can you test reading past the end of the buffer and check that the position is incremented appropriately? Perhaps also do this with the other `Read` API, too, and verify that the returned buffer `size()` is as expected

xhochy: Done.

wesm:
I can merge this once there's a JIRA to match. Alternately you can roll this into any other patch you are working on.

xhochy:
Actually made a JIRA but just forgot to mention it in the title.

wesm:
Perfect, thank you. +1

PR #89, opened 2016-04-21
PARQUET-589: Implement BufferedInputStream for better memory usage

majetideepak:
Resolves PARQUET-576 and PARQUET-590


majetideepak:
@wesm please let me know your feedback. Thanks!

majetideepak:
@xhochy do you want to create another JIRA for #88 ? Alternatively I can add your patch to this.

majetideepak:
@wesm Will this patch also resolve PARQUET-576 ?

wesm:
I will review this soon. I looked at PARQUET-576 and this patch does not resolve it. 

Currently, we have code like:

```c++
  int64_t bytes_to_read = col.meta_data.total_compressed_size;
  std::shared_ptr<Buffer> buffer = source_->ReadAt(col_start, bytes_to_read);

  if (buffer->size() < bytes_to_read) {
    throw ParquetException("Unable to read column chunk data");
  }

  std::unique_ptr<InputStream> stream(new InMemoryInputStream(buffer));
```

It would be simpler to have a simpler option:

```c++
  int64_t bytes_to_read = col.meta_data.total_compressed_size;
  std::unique_ptr<InputStream> stream;
  RETURN_NOT_OK(source_->ReadStream(col_start, bytes_to_read, &stream));
```

If that makes sense.

src/parquet/file/reader-internal.cc
```diff
@@ -174,13 +175,19 @@ std::unique_ptr<PageReader> SerializedRowGroup::GetColumnPageReader(int i) {
   }
 
   int64_t bytes_to_read = col.meta_data.total_compressed_size;
-  std::shared_ptr<Buffer> buffer = source_->ReadAt(col_start, bytes_to_read);
+  std::unique_ptr<InputStream> stream;
 
-  if (buffer->size() < bytes_to_read) {
-    throw ParquetException("Unable to read column chunk data");
+  if (chunk_size == 0) {
```
wesm: This is a sort of ugly. Perhaps we should consider IO option configuration more generally. How does parquet-mr handle this?

src/parquet/file/reader-internal.cc
```diff
@@ -174,13 +175,19 @@ std::unique_ptr<PageReader> SerializedRowGroup::GetColumnPageReader(int i) {
...
+    std::shared_ptr<Buffer> buffer = source_->ReadAt(col_start, bytes_to_read);
+    if (buffer->size() < bytes_to_read) {
+      throw ParquetException("Unable to read column chunk data");
+    }
+    stream.reset(new InMemoryInputStream(buffer));
+  } else {
+    stream.reset(new ChunkedInMemoryInputStream(source_, allocator_, col_start,
+        col_start + bytes_to_read, chunk_size));
```
wesm: Per the PR-level discussion about simplifying the RandomAccessSource API, we could add a `source_->ReadChunkedStream(...)` API along with `ReadStream`. I'm not sure where is the best place to put the logic to switch between chunked and non-chunked streams (it would be better to have it in *exactly one* place rather than each place you might like to have both options)

src/parquet/file/reader.cc
```diff
@@ -54,11 +54,11 @@ int64_t RowGroupReader::num_rows() const {
   return contents_->num_rows();
 }
 
-std::shared_ptr<ColumnReader> RowGroupReader::Column(int i) {
+std::shared_ptr<ColumnReader> RowGroupReader::Column(int i, int64_t chunk_size) {
```
wesm: Per above having this parameter here feels like an unpleasant detail that should be handled higher up in the IO configuration for the file reader

src/parquet/util/input-output-test.cc
```diff
@@ -34,6 +34,64 @@
 
 namespace parquet {
 
+TEST(TestChunkedInMemoryInputStream, Basics) {
+  int64_t source_size = 256;
+  int64_t chunk_size = 50;
+  auto buf = std::make_shared<OwnedMutableBuffer>(source_size);
+  ASSERT_EQ(source_size, buf->size());
+  for (int i = 0; i < source_size; i++) {
+    buf->mutable_data()[i] = (uint8_t)i;
```
wesm: use `static_cast<uint8_t>(i)` if a cast is indeed necessary

wesm:
In reading this patch, I think it would make sense to have an IO interface on top of the `RandomAccessSource` that takes the file-level IO options and produces `InputStream` objects (either buffered/chunked or not based on the user's wishes). As opposed to adding methods to `RandomAccessSource`.

src/parquet/util/input.cc
```diff
@@ -235,4 +234,49 @@ void InMemoryInputStream::Advance(int64_t num_bytes) {
...
 }
 
+// ----------------------------------------------------------------------
+// ChunkedInMemoryInputStream
+ChunkedInMemoryInputStream::ChunkedInMemoryInputStream(RandomAccessSource* source,
+    MemoryAllocator* pool, int64_t start, int64_t end, int64_t chunk_size) :
+    source_(source), stream_offset_(start), stream_end_(end), buffer_offset_(0) {
+  buffer_ = std::make_shared<OwnedMutableBuffer>(chunk_size, pool);
+  chunk_size_ = buffer_->size();
+  buffer_offset_ = chunk_size_;
```
wesm: Why does this variable start out equal to `chunk_size_` (versus 0)?

src/parquet/util/input.cc
```diff
@@ -235,4 +234,49 @@ void InMemoryInputStream::Advance(int64_t num_bytes) {
...
+}
+
+const uint8_t* ChunkedInMemoryInputStream::Peek(int64_t num_to_peek, int64_t* num_bytes) {
+  *num_bytes = std::min(static_cast<int64_t>(num_to_peek), stream_end_ - stream_offset_);
```
wesm: `num_to_peek` is already `int64_t`

src/parquet/util/input.cc
```diff
@@ -235,4 +234,49 @@ void InMemoryInputStream::Advance(int64_t num_bytes) {
...
+  // increase the buffer size if needed
+  if (*num_bytes > chunk_size_) {
+    buffer_->Resize(*num_bytes);
+    chunk_size_ = buffer_->size();
+    if (chunk_size_ < *num_bytes) {
+      throw ParquetException("Unable to read column data into buffer.");
```
wesm: Should this be a DCHECK instead (seems like a "shouldn't fail" condition -- if allocation failed then `Resize` should have thrown an exception)?

src/parquet/util/input.h
```diff
@@ -183,6 +184,25 @@ class InMemoryInputStream : public InputStream {
   int64_t offset_;
 };
 
+// Implementation of an InputStream when only some of the bytes are in memory.
+class ChunkedInMemoryInputStream : public InputStream {
```
wesm: Can we call this a `BufferedInputStream`? Also, change the name of `chunk_size` and related arguments to `buffer_size`. We should document that while the passed `buffer_size` is the minimum size of the buffer, it will be grown (and, at the moment, never shrunk) as required based on the sizes of reads requested by the user.

src/parquet/util/input.cc
majetideepak: This is to force a Read lazily. Alternatively, we can read into the buffer here and set the value to 0.

majetideepak:
@wesm Thanks for the feedback! I agree with the simplification part. I will look into how parquet-mr does this and IO configuration setup.

src/parquet/util/input.cc
wesm: Or you could wait to initialize buffer (since the buffer size is only a soft limit at the moment -- how does BufferedInputStream in Java handle this?)

wesm:
see @xhochy comments in #90 re: global options for reading and writing. I'd rather we do that now versus returning later and having to refactor

xhochy:
@majetideepak See https://github.com/apache/parquet-mr/blob/master/parquet-column/src/main/java/org/apache/parquet/column/ParquetProperties.java for parquet-mr's implementation.

majetideepak:
@xhochy Thanks for that link! I will have it ready by tomorrow.

majetideepak:
@wesm @xhochy Let me know your comments. I added separate classes for Reader and Writer Properties.
I did not implement the complete list of properties.

src/parquet/file/reader-internal.cc
```diff
@@ -174,15 +174,13 @@ std::unique_ptr<PageReader> SerializedRowGroup::GetColumnPageReader(int i) {
...
 
   int64_t bytes_to_read = col.meta_data.total_compressed_size;
-  std::shared_ptr<Buffer> buffer = source_->ReadAt(col_start, bytes_to_read);
+  std::unique_ptr<InputStream> stream;
 
-  if (buffer->size() < bytes_to_read) {
-    throw ParquetException("Unable to read column chunk data");
-  }
+  stream = properties_.GetInputStream();
+  stream->Initialize(source_, col_start, bytes_to_read);
```
majetideepak: `ReaderProperties` will give us the user specified stream (chunked/ not chunked). The stream can then be `Initialized` via the `source_`.

src/parquet/file/reader.h
```diff
@@ -100,11 +101,11 @@ class ParquetFileReader {
 
   // API Convenience to open a serialized Parquet file on disk
   static std::unique_ptr<ParquetFileReader> OpenFile(const std::string& path,
-      bool memory_map = true, MemoryAllocator* allocator = default_allocator());
+      ReaderProperties opts, bool memory_map = true);
```
xhochy: Given that we often just want to use the sane defaults, wouldn't it be better user experience if we could provide here a default instantiation?  

Then we would not need to clutter the code with `ReaderProperties opts;` lines just to get the defaults.

majetideepak: Agreed! Will add a default.

src/parquet/column/properties.h
```diff
@@ -0,0 +1,153 @@
...
+        }
+        break;
+      }
+      default:
+        break;
+    }
+    return result.str();
+  }
+
+  MemoryAllocator* get_allocator() {
```
xhochy: Remove the `get_` prefix, yet we only used the name of the attribute as the name of the getter.

src/parquet/column/properties.h
```diff
@@ -0,0 +1,153 @@
...
+      stream.reset(new InMemoryInputStream());
+    }
+    return stream;
+  }
+
+  bool use_buffered_stream() {
+    return use_buffered_stream_;
+  }
+
+  int64_t get_buffer_size() {
```
xhochy: Remove the `get_` prefix, yet we only used the name of the attribute as the name of the getter.

src/parquet/column/properties.h
```diff
@@ -0,0 +1,153 @@
...
+      int dictionary_pagesize = DEFAULT_DICTIONARY_PAGE_SIZE,
+      bool enable_dictionary = DEFAULT_IS_DICTIONARY_ENABLED,
+      MemoryAllocator* allocator = default_allocator()) {
+    pagesize_ = pagesize;
+    dictionary_pagesize_ = dictionary_pagesize;
+    enable_dictionary_ = enable_dictionary;
+    allocator_ = allocator;
+  }
+
+  int get_dictionary_pagesize_threshold() {
```
xhochy: See above

src/parquet/column/properties.h
```diff
@@ -0,0 +1,153 @@
...
+    return dictionary_pagesize_;
+  }
+
+  int get_pagesize_threshold() {
```
xhochy: See above

src/parquet/column/properties.h
```diff
@@ -0,0 +1,153 @@
...
+    return pagesize_;
+  }
+
+  bool is_enable_dictionary() {
+    return enable_dictionary_;
+  }
+
+  MemoryAllocator* get_allocator() {
```
xhochy: Also here. We might differ then interface wise from Java but are more consitent with the existing ones.

src/parquet/column/properties.h
```diff
@@ -0,0 +1,157 @@
...
+class ReaderProperties {
+ public:
+  ReaderProperties(bool use_buffered_stream = false, int64_t buffer_size = 0,
+      MemoryAllocator* allocator = default_allocator()) {
+    enable_buffered_stream_ = use_buffered_stream;
+    buffer_size_ = buffer_size;
+    allocator_ = allocator;
+  }
+
+  static inline std::string type_printer(Type::type parquet_type, const char* val,
```
wesm: This is an odd place for this function. Can you move this to utility code someplace (e.g. types.h)?

src/parquet/column/properties.h
```diff
@@ -0,0 +1,157 @@
...
+  int64_t buffer_size_;
+};
+
+ReaderProperties default_reader_properties();
+
+class WriterProperties {
+ public:
+  WriterProperties(int pagesize = DEFAULT_PAGE_SIZE,
+      int dictionary_pagesize = DEFAULT_DICTIONARY_PAGE_SIZE,
+      bool enable_dictionary = DEFAULT_IS_DICTIONARY_ENABLED,
```
wesm: As the number of parameters grow, it may be better to have a ctor with minimal arguments and then add setter functions for each of the particular parameters.

src/parquet/file/reader-internal.cc
```diff
@@ -205,9 +203,9 @@ static constexpr uint32_t FOOTER_SIZE = 8;
 static constexpr uint8_t PARQUET_MAGIC[4] = {'P', 'A', 'R', '1'};
 
 std::unique_ptr<ParquetFileReader::Contents> SerializedFile::Open(
-    std::unique_ptr<RandomAccessSource> source, MemoryAllocator* allocator) {
+    std::unique_ptr<RandomAccessSource> source, ReaderProperties opts) {
   std::unique_ptr<ParquetFileReader::Contents> result(
-      new SerializedFile(std::move(source), allocator));
+      new SerializedFile(std::move(source), opts));
```
wesm: `props`?

src/parquet/file/reader-internal.cc
wesm: How about:

```c++
stream = properties_.GetStream(source_, col_start, bytes_to_read);
```

I haven't reviewed lower yet though to see if this might be a problem

src/parquet/file/reader-internal.cc
```diff
@@ -247,8 +246,8 @@ int SerializedFile::num_row_groups() const {
 
 SerializedFile::SerializedFile(
     std::unique_ptr<RandomAccessSource> source,
-    MemoryAllocator* allocator = default_allocator()) :
-        source_(std::move(source)), allocator_(allocator) {}
+    ReaderProperties opts) :
+        source_(std::move(source)), properties_(opts) {}
```
wesm: `properties`

src/parquet/file/reader.h
```diff
@@ -100,11 +101,11 @@ class ParquetFileReader {
 
   // API Convenience to open a serialized Parquet file on disk
   static std::unique_ptr<ParquetFileReader> OpenFile(const std::string& path,
-      bool memory_map = true, MemoryAllocator* allocator = default_allocator());
+      bool memory_map = true, ReaderProperties opts = default_reader_properties());
 
   static std::unique_ptr<ParquetFileReader> Open(
       std::unique_ptr<RandomAccessSource> source,
-      MemoryAllocator* allocator = default_allocator());
+      ReaderProperties opts = default_reader_properties());
```
wesm: argument name again

majetideepak:
@wesm As per your comment on lazily initializing  the buffer for BufferedInputStream class, the Java's implementation does not do it either. It initializes the buffer inside the constructor as well.
 http://grepcode.com/file/repository.grepcode.com/java/root/jdk/openjdk/8u40-b25/java/io/BufferedInputStream.java#BufferedInputStream.%3Cinit%3E%28java.io.InputStream%29

majetideepak:
@wesm let me know if you have more feedback. Thanks.

src/parquet/column/properties.h
```diff
@@ -0,0 +1,115 @@
...
+    if (buffered_stream_enabled_) {
+      stream.reset(
+          new BufferedInputStream(allocator_, buffer_size_, source, start, num_bytes));
+    } else {
+      stream.reset(new InMemoryInputStream(source, start, num_bytes));
+    }
+    return stream;
+  }
+
+  bool is_buffered_stream_enabled() { return buffered_stream_enabled_; }
```
wesm: ... const

src/parquet/column/properties.h
```diff
@@ -0,0 +1,115 @@
...
+
+  void enable_buffered_stream() { buffered_stream_enabled_ = true; }
+
+  void disable_buffered_stream() { buffered_stream_enabled_ = false; }
+
+  void set_buffer_size(int64_t buf_size) { buffer_size_ = buf_size; }
+
+  int64_t buffer_size() { return buffer_size_; }
```
wesm: const

src/parquet/column/properties.h
```diff
@@ -0,0 +1,115 @@
...
+    dictionary_enabled_ = DEFAULT_IS_DICTIONARY_ENABLED;
+  }
+
+  int64_t dictionary_pagesize() { return dictionary_pagesize_; }
+
+  void set_dictionary_pagesize(int64_t dictionary_psize) {
+    dictionary_pagesize_ = dictionary_psize;
+  }
+
+  int64_t data_pagesize() { return pagesize_; }
```
wesm: const here and above

src/parquet/types.h
```diff
@@ -266,6 +266,51 @@ inline std::string format_fwf(int width) {
   return ss.str();
 }
 
+static inline std::string type_printer(
```
wesm: Can we rename this `FormatValue`?

wesm: Also, it may not be worth inlining

wesm:
Only these minor comments, then we should be good to go

CMakeLists.txt
```diff
@@ -359,6 +359,8 @@ endif()
 # Library config
 
 set(LIBPARQUET_SRCS
+  src/parquet/types.cc
```
majetideepak: I noticed that there are other functions where `inline` might not be helpful. I moved these to a separate file.

src/parquet/column/properties.h
```diff
@@ -83,21 +83,21 @@ class WriterProperties {
...
+  const int64_t data_pagesize() { return pagesize_; }
 
   void set_data_pagesize(int64_t pg_size) { pagesize_ = pg_size; }
 
   void enable_dictionary() { dictionary_enabled_ = true; }
 
   void disable_dictionary() { dictionary_enabled_ = false; }
 
-  bool is_dictionary_enabled() { return dictionary_enabled_; }
+  const bool is_dictionary_enabled() { return dictionary_enabled_; }
```
wesm: Sorry, I should have been more clear -- put `const` after the method names

src/parquet/column/properties.h
```diff
@@ -0,0 +1,115 @@
...
+
+  const bool is_buffered_stream_enabled() { return buffered_stream_enabled_; }
+
+  void enable_buffered_stream() { buffered_stream_enabled_ = true; }
+
+  void disable_buffered_stream() { buffered_stream_enabled_ = false; }
+
+  void set_buffer_size(int64_t buf_size) { buffer_size_ = buf_size; }
+
+  const int64_t buffer_size() { return buffer_size_; }
```
wesm: commenting again because of github weirdness. `int64_t buffer_size() const` instead.

src/parquet/column/properties.h
```diff
@@ -0,0 +1,115 @@
...
+
+  const int64_t data_pagesize() { return pagesize_; }
+
+  void set_data_pagesize(int64_t pg_size) { pagesize_ = pg_size; }
+
+  void enable_dictionary() { dictionary_enabled_ = true; }
+
+  void disable_dictionary() { dictionary_enabled_ = false; }
+
+  const bool is_dictionary_enabled() { return dictionary_enabled_; }
```
wesm: same with these, `const` after the method name

wesm:
I'll merge after addressing these const issues. Thanks

majetideepak:
My fault. I didn't read your comment carefully. Fixed const issues.

wesm:
+1, thank you!

PR #90, opened 2016-04-23
PARQUET-436: Implement basic Write Support

xhochy:
Basic write support with only DataPage as page type, no compression and fixed page sizes. Also allocates and consumes more memory than the optimal solution. But I hope that it is in a state where we can review & merge, so that afterwards the work can be split up.

One outstanding design question for me is `WriteDataPage`. We could pass here a `DataPage` instance but would also need to take care of `num_rows` and `num_values` correctly for the thrift metadata, hoping for good suggestions here.

Also there are numerous TODOs in here for which we need to decide if they should be done as part of the PR or if we should open JIRAs for them.

### DIFF
--- a/src/parquet/column/column-writer-test.cc
+++ b/src/parquet/column/column-writer-test.cc
@@ -44,6 +44,11 @@ class TestPrimitiveWriter : public ::testing::Test {
     schema = std::make_shared<ColumnDescriptor>(node, 1, 0);
   }
 
+  void SetUpSchemaOptionalRepeated() {
+    node = PrimitiveNode::Make("int64", Repetition::REPEATED, Type::INT64);
+    schema = std::make_shared<ColumnDescriptor>(node, 1, 1);
+  }
+
   void SetUp() {
     SetUpSchemaRequiredNonRepeated();
   }
@@ -72,6 +77,7 @@ TEST_F(TestPrimitiveWriter, WriteReadLoopSinglePage) {
   std::fill(definition_levels.begin(), definition_levels.end(), 1);
   definition_levels[1] = 0;
   std::vector<int16_t> repetition_levels(100);
+  std::fill(repetition_levels.begin(), repetition_levels.end(), 0);
 
   // Output buffers
   std::vector<int64_t> values_out(100);
@@ -94,7 +100,6 @@ TEST_F(TestPrimitiveWriter, WriteReadLoopSinglePage) {
   SetUpSchemaOptionalNonRepeated();
   sink.reset(new InMemoryOutputStream());
   writer = BuildWriter(sink.get());
-  // TODO: Implement definition_levels
   writer->WriteBatch(values.size(), definition_levels.data(), nullptr, values.data());
   writer->Close();
   
@@ -108,7 +113,19 @@ TEST_F(TestPrimitiveWriter, WriteReadLoopSinglePage) {
   ASSERT_EQ(values_out, values_expected);
 
   // Test case 3: optional and repeated, so definition and repetition levels 
-  // TODO
+  SetUpSchemaOptionalRepeated();
+  sink.reset(new InMemoryOutputStream());
+  writer = BuildWriter(sink.get());
+  writer->WriteBatch(values.size(), definition_levels.data(), repetition_levels.data(), values.data());
+  writer->Close();
+  
+  reader = BuildReader(sink->GetBuffer());
+  values_read = 0;
+  values_out.resize(100);
+  reader->ReadBatch(values.size(), definition_levels_out.data(), repetition_levels_out.data(), values_out.data(), &values_read);
+  ASSERT_EQ(values_read, 99);
+  values_out.resize(99);
+  ASSERT_EQ(values_out, values_expected);
 }
 
 } // namespace test
--- a/src/parquet/column/column-writer-test.cc
+++ b/src/parquet/column/column-writer-test.cc
@@ -66,11 +66,12 @@ class TestPrimitiveWriter : public ::testing::Test {
         new Int64Reader(schema.get(), std::move(page_reader)));
   }
 
-  std::unique_ptr<Int64Writer> BuildWriter() {
+  std::unique_ptr<Int64Writer> BuildWriter(int64_t output_size = 100) {
     sink_.reset(new InMemoryOutputStream());
     std::unique_ptr<SerializedPageWriter> pager(
         new SerializedPageWriter(sink_.get(), Compression::UNCOMPRESSED));
-    return std::unique_ptr<Int64Writer>(new Int64Writer(schema.get(), std::move(pager)));
+    return std::unique_ptr<Int64Writer>(new Int64Writer(schema.get(), std::move(pager),
+          output_size));
   }
 
   void ReadColumn() {
@@ -81,7 +82,7 @@ class TestPrimitiveWriter : public ::testing::Test {
 
  protected:
   int64_t values_read;
-  
+
   // Output buffers
   std::vector<int64_t> values_out;
   std::vector<int16_t> definition_levels_out;
@@ -110,12 +111,12 @@ TEST_F(TestPrimitiveWriter, OptionalNonRepeated) {
   // Optional and non-repeated, with definition levels
   // but no repetition levels
   SetUpSchemaOptionalNonRepeated();
-  
+
   std::vector<int64_t> values(100, 128);
   std::vector<int16_t> definition_levels(100, 1);
   definition_levels[1] = 0;
   std::vector<int64_t> values_expected(99, 128);
-  
+
   auto writer = BuildWriter();
   writer->WriteBatch(values.size(), definition_levels.data(), nullptr, values.data());
   writer->Close();
@@ -144,6 +145,54 @@ TEST_F(TestPrimitiveWriter, OptionalRepeated) {
   ReadColumn();
   ASSERT_EQ(values_read, 99);
   values_out.resize(99);
+  ASSERT_EQ(values_out, values_expected);
+}
+
+TEST_F(TestPrimitiveWriter, RequiredTooFewRows) {
+  std::vector<int64_t> values(99, 128);
+
+  auto writer = BuildWriter();
+  writer->WriteBatch(values.size(), nullptr, nullptr, values.data());
+  ASSERT_THROW(writer->Close(), ParquetException);
+}
+
+TEST_F(TestPrimitiveWriter, RequiredTooMany) {
+  std::vector<int64_t> values(200, 128);
+
+  auto writer = BuildWriter();
+  ASSERT_THROW(writer->WriteBatch(values.size(), nullptr, nullptr, values.data()),
+      ParquetException);
+}
+
+TEST_F(TestPrimitiveWriter, OptionalRepeatedTooFewRows) {
+  // Optional and repeated, so definition and repetition levels
+  SetUpSchemaOptionalRepeated();
+
+  std::vector<int64_t> values(100, 128);
+  std::vector<int16_t> definition_levels(100, 1);
+  definition_levels[1] = 0;
+  std::vector<int16_t> repetition_levels(100, 0);
+  repetition_levels[3] = 1;
+  std::vector<int64_t> values_expected(99, 128);
+
+  auto writer = BuildWriter();
+  writer->WriteBatch(values.size(), definition_levels.data(),
+      repetition_levels.data(), values.data());
+  ASSERT_THROW(writer->Close(), ParquetException);
+}
+
+TEST_F(TestPrimitiveWriter, RequiredNonRepeatedLargeChunk) {
+  std::vector<int64_t> values(10000, 128);
+  std::vector<int64_t> values_expected(100, 128);
+
+  // Test case 1: required and non-repeated, so no definition or repetition levels
+  std::unique_ptr<Int64Writer> writer = BuildWriter(10000);
+  writer->WriteBatch(values.size(), nullptr, nullptr, values.data());
+  writer->Close();
+
+  // Just read the first 100 to ensure we could read it back in
+  ReadColumn();
+  ASSERT_EQ(values_read, 100);
   ASSERT_EQ(values_out, values_expected);
 }
 
--- a/src/parquet/column/column-writer-test.cc
+++ b/src/parquet/column/column-writer-test.cc
@@ -34,17 +34,17 @@ namespace test {
 
 class TestPrimitiveWriter : public ::testing::Test {
  public:
-  void SetUpSchemaRequiredNonRepeated() {
+  void SetUpSchemaRequired() {
     node_ = PrimitiveNode::Make("int64", Repetition::REQUIRED, Type::INT64);
     schema_ = std::make_shared<ColumnDescriptor>(node_, 0, 0);
   }
 
-  void SetUpSchemaOptionalNonRepeated() {
-    node_ = PrimitiveNode::Make("int64", Repetition::REQUIRED, Type::INT64);
+  void SetUpSchemaOptional() {
+    node_ = PrimitiveNode::Make("int64", Repetition::OPTIONAL, Type::INT64);
     schema_ = std::make_shared<ColumnDescriptor>(node_, 1, 0);
   }
 
-  void SetUpSchemaOptionalRepeated() {
+  void SetUpSchemaRepeated() {
     node_ = PrimitiveNode::Make("int64", Repetition::REPEATED, Type::INT64);
     schema_ = std::make_shared<ColumnDescriptor>(node_, 1, 1);
   }
@@ -54,7 +54,7 @@ class TestPrimitiveWriter : public ::testing::Test {
     definition_levels_out_.resize(100);
     repetition_levels_out_.resize(100);
 
-    SetUpSchemaRequiredNonRepeated();
+    SetUpSchemaRequired();
   }
 
   std::unique_ptr<Int64Reader> BuildReader() {
@@ -104,14 +104,14 @@ TEST_F(TestPrimitiveWriter, RequiredNonRepeated) {
   writer->Close();
 
   ReadColumn();
-  ASSERT_EQ(values_read_, 100);
-  ASSERT_EQ(values_out_, values);
+  ASSERT_EQ(100, values_read_);
+  ASSERT_EQ(values, values_out_);
 }
 
 TEST_F(TestPrimitiveWriter, OptionalNonRepeated) {
   // Optional and non-repeated, with definition levels
   // but no repetition levels
-  SetUpSchemaOptionalNonRepeated();
+  SetUpSchemaOptional();
 
   std::vector<int64_t> values(100, 128);
   std::vector<int16_t> definition_levels(100, 1);
@@ -122,15 +122,15 @@ TEST_F(TestPrimitiveWriter, OptionalNonRepeated) {
   writer->Close();
 
   ReadColumn();
-  ASSERT_EQ(values_read_, 99);
+  ASSERT_EQ(99, values_read_);
   values_out_.resize(99);
   values.resize(99);
-  ASSERT_EQ(values_out_, values);
+  ASSERT_EQ(values, values_out_);
 }
 
 TEST_F(TestPrimitiveWriter, OptionalRepeated) {
   // Optional and repeated, so definition and repetition levels
-  SetUpSchemaOptionalRepeated();
+  SetUpSchemaRepeated();
 
   std::vector<int64_t> values(100, 128);
   std::vector<int16_t> definition_levels(100, 1);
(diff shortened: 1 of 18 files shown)


xhochy:
Furthermore I only tested the write-read roundtrip with parquet-cpp. If someone could test it with the Java implementation (or point me to an easy way to do it), would be very nice.

src/parquet/column/column-writer-test.cc
```diff
@@ -0,0 +1,203 @@
...
+
+  // Output buffers
+  std::vector<int64_t> values_out;
+  std::vector<int16_t> definition_levels_out;
+  std::vector<int16_t> repetition_levels_out;
+
+ private:
+  NodePtr node;
+  format::ColumnChunk metadata;
+  std::shared_ptr<ColumnDescriptor> schema;
```
wesm: Give trailing underscores to all the member variables

src/parquet/column/writer.cc
```diff
@@ -0,0 +1,174 @@
...
+  InitSinks();
+}
+
+void ColumnWriter::InitSinks() {
+  definition_levels_sink_ = std::unique_ptr<InMemoryOutputStream>(
+      new InMemoryOutputStream());
+  repetition_levels_sink_ = std::unique_ptr<InMemoryOutputStream>(
+      new InMemoryOutputStream());
+  values_sink_ = std::unique_ptr<InMemoryOutputStream>(
+      new InMemoryOutputStream());
```
wesm: It's OK (and less verbose) to use `reset` on the member variables

src/parquet/column/writer.cc
```diff
@@ -0,0 +1,174 @@
...
+void ColumnWriter::WriteRepetitionLevels(int64_t num_levels, int16_t* levels) {
+  repetition_levels_sink_->Write(reinterpret_cast<uint8_t*>(levels),
+      sizeof(int16_t) * num_levels);
+}
+
+std::shared_ptr<Buffer> ColumnWriter::RleEncodeLevels(
+    const std::shared_ptr<Buffer>& buffer, int16_t max_level) {
+  // TODO: This only works with due to some RLE specifics
+  int64_t rle_size = 2 * num_buffered_values_ + sizeof(uint32_t);
+  std::shared_ptr<OwnedMutableBuffer> buffer_rle =
```
wesm: OK to use `auto` here

src/parquet/column/writer.cc
```diff
@@ -0,0 +1,174 @@
...
+    std::make_shared<OwnedMutableBuffer>(rle_size, allocator_);
+  level_encoder_.Init(Encoding::RLE, max_level,
+      num_buffered_values_, buffer_rle->mutable_data() + sizeof(uint32_t),
+      buffer_rle->size() - sizeof(uint32_t));
+  int encoded = level_encoder_.Encode(num_buffered_values_,
+      reinterpret_cast<const int16_t*>(buffer->data()));
+  DCHECK_EQ(encoded, num_buffered_values_);
+  reinterpret_cast<uint32_t*>(buffer_rle->mutable_data())[0] = level_encoder_.len();
+  buffer_rle->Resize(level_encoder_.len() + sizeof(uint32_t));
```
wesm: DCHECK that the actual size does not exceed the estimate above?

src/parquet/column/writer.cc
```diff
@@ -0,0 +1,174 @@
...
+  return std::static_pointer_cast<Buffer>(buffer_rle);
+}
+
+void ColumnWriter::WriteNewPage() {
+  // TODO: Currently we only support writing DataPages
+  std::shared_ptr<Buffer> definition_levels = definition_levels_sink_->GetBuffer();
+  std::shared_ptr<Buffer> repetition_levels = repetition_levels_sink_->GetBuffer();
+  std::shared_ptr<Buffer> values = values_sink_->GetBuffer();
```
wesm: This may cause some allocation-deallocation thrashing versus reusing the same buffers after completing each data page. Perhaps the output stream can be changed to enable buffer reuse and resetting without deallocation.

src/parquet/column/writer.h
```diff
@@ -0,0 +1,206 @@
...
+
+  std::unique_ptr<InMemoryOutputStream> definition_levels_sink_;
+  std::unique_ptr<InMemoryOutputStream> repetition_levels_sink_;
+  std::unique_ptr<InMemoryOutputStream> values_sink_;
+
+ private:
+  void InitSinks();
+};
+
+// API to read values from a single column. This is the main client facing API.
```
wesm: Stale comment

src/parquet/column/writer.h
```diff
@@ -0,0 +1,206 @@
...
+  std::unordered_map<int, std::shared_ptr<EncoderType> > encoders_;
+
+  void ConfigureDictionary(const DictionaryPage* page);
+
+  std::unique_ptr<EncoderType> current_encoder_;
+};
+
+// TODO(PARQUET-591): This is just chosen at random, we should make better estimates.
+// See also: parquet-column/../column/impl/ColumnWriteStoreV2.java:sizeCheck
+const int64_t PAGE_VALUE_COUNT = 1000;
```
wesm: Or also make it configurable

wesm:
this is a lot of work, exciting to see the progress! it will take me some time to review, i'll work some on it rest of today and try to finish tomorrow

src/parquet/column/writer.h
```diff
@@ -0,0 +1,206 @@
...
+
+template <int TYPE>
+inline void TypedColumnWriter<TYPE>::WriteBatch(int64_t num_values, int16_t* def_levels,
+    int16_t* rep_levels, T* values) {
```
wesm: On some contemplation, I'm curious whether the API should be one page at a time versus accumulate-then-write. This leaves the user to be responsible for page delimiting.

wesm: The downside of this is that some applications might have to implement their own buffering

wesm:
@xhochy see https://github.com/Parquet/parquet-compatibility for some earlier integration tests

src/parquet/column/column-writer-test.cc
```diff
@@ -0,0 +1,203 @@
...
+  std::unique_ptr<InMemoryOutputStream> sink_;
+};
+
+TEST_F(TestPrimitiveWriter, RequiredNonRepeated) {
+  std::vector<int64_t> values(100, 128);
```
wesm: as we build out the write support, we'll want to use more diverse or randomly generated test data

src/parquet/column/column-writer-test.cc
```diff
@@ -0,0 +1,203 @@
...
+
+TEST_F(TestPrimitiveWriter, OptionalRepeated) {
+  // Optional and repeated, so definition and repetition levels
+  SetUpSchemaOptionalRepeated();
+
+  std::vector<int64_t> values(100, 128);
+  std::vector<int16_t> definition_levels(100, 1);
+  definition_levels[1] = 0;
+  std::vector<int16_t> repetition_levels(100, 0);
+  std::vector<int64_t> values_expected(99, 128);
```
wesm: Any reason not to make `values` and `values_expected` the same object?

src/parquet/column/column-writer-test.cc
xhochy: For the moment, I would like to keep it in this simple way as we do not have a good tool in our chain for a better input diversification. Once we can use something like RapidCheck, property based testing will definitely unveil some buggy corner cases but I hate to see users reporting randomly failing unit tests without that they can provide the necessary input to reproduce them.

xhochy: Still: If really needed, we could add randomly generated data here but with a fixed seed.

src/parquet/column/writer.cc
xhochy: This is what I meant with "Also allocates and consumes more memory than the optimal solution." In combination with the `ChunkedInMemoryInputStream` and a "streaming" RLE-LevelEncoder we could re-write this in a separate change.

src/parquet/column/writer.h
xhochy: I'd like to have a general solution in a `WriterPropeties` class where by default paging is done on a size-based level. If we keep the interface of this part of `WriterProperties` a bit more general we could also open the ability up to the user to implement paging on a per-row basis.

src/parquet/column/writer.h
xhochy: In addition to the comment above, paging is something that I thought should normally be invisible to the user. 

From a user perspective, `ReadBatch` and `WriteBatch` feel really comfortable as the Parquet internals are all hidden.

xhochy:
Any hint on the linking problems with clang on OSX? Do we need `TypedColumnReader<>::ReadNewPage()` and `TypedColumnWriter::TypedColumnWriter` in the header files instead of the `.cc`? Or is there another problem?

wesm:
It seems we need to force-instantiate the `TypedColumnReader` types in `parquet/column/reader.cc`. Similar to the bottom of `parquet/util/buffer.cc`.

wesm:
In general this is probably a good practice so we can push down template-class functions that we don't want to be always inlined.

src/parquet/column/column-writer-test.cc
wesm: Yes -- in some of the other test cases we're generating "random" data from a set seed. Not urgent for now.

src/parquet/column/column-writer-test.cc
```diff
@@ -0,0 +1,202 @@
...
+TEST_F(TestPrimitiveWriter, RequiredNonRepeated) {
+  std::vector<int64_t> values(100, 128);
+
+  // Test case 1: required and non-repeated, so no definition or repetition levels
+  std::unique_ptr<Int64Writer> writer = BuildWriter();
+  writer->WriteBatch(values.size(), nullptr, nullptr, values.data());
+  writer->Close();
+
+  ReadColumn();
+  ASSERT_EQ(values_read_, 100);
```
wesm: The expected value goes first in googletest (no big deal but it impacts the failure message)

src/parquet/file/writer.h
```diff
@@ -0,0 +1,171 @@
...
+
+namespace parquet {
+
+class ColumnWriter;
+class PageWriter;
+class OutputStream;
+
+class RowGroupWriter {
+ public:
+  // Forward declare the PIMPL
```
wesm: Note that I had originally planned to do a proper PIMPL but found that I needed to expose elements of the `Contents` API in the public API (so mocks could be constructed), so we should probably remove all references to PIMPL unless it is truly a private implementation.

src/parquet/file/writer.h
```diff
@@ -0,0 +1,171 @@
...
+    virtual const SchemaDescriptor* schema() const = 0;
+  };
+
+  RowGroupWriter(std::unique_ptr<Contents> contents, MemoryAllocator* allocator);
+
+  /**
+   * Construct a ColumnWriter for the indicated row group-relative column.
+   *
+   * Ownership is solely within the RowGroupWriter. The ColumnWriter is only valid
+   * until the next call to NextColumn or Close. As the contents are diretly written to
```
wesm: directly

src/parquet/file/writer.h
```diff
@@ -0,0 +1,171 @@
...
+  struct Contents {
+    virtual int num_columns() const = 0;
+    virtual int64_t num_rows() const = 0;
+
+    // TODO: PARQUET-579
+    // virtual void WriteRowGroupStatitics();
+    virtual ColumnWriter* NextColumn() = 0;
+    virtual void Close() = 0;
+
+    // Return const-poitner to make it clear that this object is not to be copied
```
wesm: pointer

src/parquet/file/writer.h
```diff
@@ -0,0 +1,171 @@
...
+  /**
+   * Number of rows that shall be written as part of this RowGroup.
+   */
+  int64_t num_rows() const;
+
+  // TODO: PARQUET-579
+  // virtual void WriteRowGroupStatitics();
+
+ private:
+  // Owned by the parent ParquetFileReader
```
wesm: FileWriter

src/parquet/file/writer.h
```diff
@@ -0,0 +1,171 @@
...
+  const SchemaDescriptor* schema_;
+
+  // PIMPL idiom
+  // This is declared in the .cc file so that we can hide compiled Thrift
+  // headers from the public API and also more easily create test fixtures.
```
wesm: Aside: we should fix my comment in `reader.cc` and also here to reflect the reality (a helper implementation class, not truly a private implementation)

src/parquet/file/writer.h
```diff
@@ -0,0 +1,171 @@
...
+  std::unique_ptr<Contents> contents_;
+
+  MemoryAllocator* allocator_;
+};
+
+class ParquetFileWriter {
+ public:
+  // Forward declare the PIMPL
```
wesm: Same comment as above

src/parquet/file/writer.h
```diff
@@ -0,0 +1,171 @@
...
+  void Open(std::unique_ptr<Contents> contents);
+  void Close();
+
+  /**
+   * Construct a RowGroupWriter for the indicated number of rows.
+   *
+   * Ownership is solely within the ParquetFileWriter. The RowGroupWriter is only valid
+   * until the next call to AppendRowGroup or Close.
+   *
+   * @param num_rows The number of rows that are stored in the new RowGroup
```
wesm: Is there a reason to use C-style comments for Doxygen? Versus C++-style comments (at least in emacs c++-mode, it's easier to always use `//`)

src/parquet/column/column-writer-test.cc
```diff
@@ -0,0 +1,202 @@
...
+
+class TestPrimitiveWriter : public ::testing::Test {
+ public:
+  void SetUpSchemaRequiredNonRepeated() {
+    node_ = PrimitiveNode::Make("int64", Repetition::REQUIRED, Type::INT64);
+    schema_ = std::make_shared<ColumnDescriptor>(node_, 0, 0);
+  }
+
+  void SetUpSchemaOptionalNonRepeated() {
+    node_ = PrimitiveNode::Make("int64", Repetition::REQUIRED, Type::INT64);
```
majetideepak: This should be `Repetition::OPTIONAL` ?

src/parquet/column/column-writer-test.cc
```diff
@@ -0,0 +1,202 @@
...
+namespace parquet {
+
+using schema::NodePtr;
+using schema::PrimitiveNode;
+
+namespace test {
+
+class TestPrimitiveWriter : public ::testing::Test {
+ public:
+  void SetUpSchemaRequiredNonRepeated() {
```
majetideepak: We have been consistently using just `Required`, `Optional` and `Repeated` to name our tests.
You might want to just use `SetUpSchemaRequired()` ?

src/parquet/column/column-writer-test.cc
```diff
@@ -0,0 +1,202 @@
...
+    node_ = PrimitiveNode::Make("int64", Repetition::REQUIRED, Type::INT64);
+    schema_ = std::make_shared<ColumnDescriptor>(node_, 0, 0);
+  }
+
+  void SetUpSchemaOptionalNonRepeated() {
```
majetideepak: similarly `SetUpSchemaOptional()`

src/parquet/column/column-writer-test.cc
```diff
@@ -0,0 +1,202 @@
...
+    node_ = PrimitiveNode::Make("int64", Repetition::REQUIRED, Type::INT64);
+    schema_ = std::make_shared<ColumnDescriptor>(node_, 1, 0);
+  }
+
+  void SetUpSchemaOptionalRepeated() {
```
majetideepak: As per above `SetUpSchemaRepeated()`

src/parquet/file/file-serialize-test.cc
```diff
@@ -0,0 +1,102 @@
...
+class TestSerialize : public ::testing::Test {
+ public:
+  void SetUpSchemaRequiredNonRepeated() {
+    auto pnode = PrimitiveNode::Make("int64", Repetition::REQUIRED, Type::INT64);
+    node = GroupNode::Make("schema", Repetition::REQUIRED, std::vector<NodePtr>({pnode}));
+    schema.Init(node);
+  }
+
+  void SetUpSchemaOptionalNonRepeated() {
+    auto pnode = PrimitiveNode::Make("int64", Repetition::REQUIRED, Type::INT64);
```
majetideepak: Should be `Repetition::OPTIONAL` ?

src/parquet/file/writer.h
xhochy: No, I'm just used to them as they are also common in Java. Vim supports both ;)

src/parquet/file/file-serialize-test.cc
```diff
@@ -0,0 +1,102 @@
...
+    schema.Init(node);
+  }
+
+  void SetUp() {
+    SetUpSchemaRequiredNonRepeated();
+  }
+
+ protected:
+  NodePtr node;
+  SchemaDescriptor schema;
```
wesm: Trailing underscore in these variables

src/parquet/file/writer-internal.cc
```diff
@@ -0,0 +1,260 @@
...
+
+  // Concatenate data into a single buffer
+  std::shared_ptr<OwnedMutableBuffer> uncompressed_data =
+    std::make_shared<OwnedMutableBuffer>(uncompressed_size, allocator_);
+  uint8_t* uncompressed_ptr = uncompressed_data->mutable_data();
+  memcpy(uncompressed_ptr, repetition_levels->data(), repetition_levels->size());
+  uncompressed_ptr += repetition_levels->size();
+  memcpy(uncompressed_ptr, definition_levels->data(), definition_levels->size());
+  uncompressed_ptr += definition_levels->size();
+  memcpy(uncompressed_ptr, values->data(), values->size());
```
wesm: Add a TODO to streams this data out directly in the uncompressed case

wesm: Add another TODO to reuse the same buffer for uncompressed page data on its way to being compressed (to avoid alloc-dealloc issues)

src/parquet/file/writer-internal.cc
```diff
@@ -0,0 +1,260 @@
...
+    throw ParquetException("All columns have already been written.");
+  }
+  current_column_index_++;
+
+  if (current_column_writer_) {
+    total_bytes_written_ += current_column_writer_->Close();
+  }
+
+  auto column_descr = schema_->Column(current_column_index_);
+  auto col_meta = &metadata_->columns[current_column_index_];
```
wesm: declare these types

wesm:
OK, I'm finished reviewing. Few small things and we can merge this and move forward with building more write functionality in follow up patches

wesm:
Looks like you need to put

```c++
template class TypedColumnReader<Type::BOOLEAN>;
template class TypedColumnReader<Type::INT32>;
template class TypedColumnReader<Type::INT64>;
template class TypedColumnReader<Type::INT96>;
template class TypedColumnReader<Type::FLOAT>;
template class TypedColumnReader<Type::DOUBLE>;
template class TypedColumnReader<Type::BYTE_ARRAY>;
template class TypedColumnReader<Type::FIXED_LEN_BYTE_ARRAY>;
```

in reader.cc and correspondingly in writer.cc. I don't have a mac on me to test, though

src/parquet/file/writer-internal.cc
xhochy: Basic rule: Use auto when the type is already explicity mentioned in the same line?

src/parquet/file/writer-internal.cc
xhochy: Done.

xhochy:
Addressed as comments as far as I can see, code also links on OS X now.

src/parquet/file/writer-internal.cc
wesm: Yeah, if the type is sufficiently clear from context (e.g. if there is a `static_cast<T>`), see https://google.github.io/styleguide/cppguide.html#auto.

wesm:
OK, the build's passing so I'm calling pencils down and we'll move forward on the other JIRAs. Thank you for the patience on this, we're well on our way toward a complete implementation! =) +1

tiger-xzp:
will the interface `WriteDataPage` support compression in future, or is there any other way to implement that?

xhochy:
It will, we are just not feature-complete at the moment.

tiger-xzp:
ok, i got it. I noticed that some compression algorithm like ` GZipCodec::Compress` has be implemented in project, can i just call that to do the compression job in `WriteDataPage`?

xhochy:
Yes, you can. Probably compression is not so hard to implement. Just be careful to add some unittests and an option to the WriterProperties to configure it.

tiger-xzp:
Thanks for kindly reminder .

PR #96, opened 2016-05-05
PARQUET-599: Better size estimation for levels

xhochy:
Still not an optimal size estimation but at least we will have always the required amount.


wesm:
+1, thank you

wesm:
If you are able to add some test cases probing this code path, that would be great, but I merged in the interest of expediency as it appears to be a generally more precise estimate than what we were doing before.

PR #97, opened 2016-05-08
PARQUET-603: Implement missing information in schema descriptor

majetideepak:


src/parquet/column/CMakeLists.txt
```diff
@@ -19,6 +19,7 @@
 install(FILES
   page.h
   levels.h
+  properties.h
```
majetideepak: I missed this in the previous patch

src/parquet/schema/descriptor.h
```diff
@@ -99,14 +99,18 @@ class SchemaDescriptor {
 
   const schema::NodePtr& schema() const { return schema_; }
 
+  const schema::GroupNode* group() const { return group_; }
+
+  const schema::NodePtr& base(int i) const;
```
wesm: Add a comment explaining this method -- I'm wondering what might be a more descriptive name, like `GetLeafRoot` or something

src/parquet/schema/descriptor.cc
```diff
@@ -81,9 +82,19 @@ ColumnDescriptor::ColumnDescriptor(const schema::NodePtr& node,
...
+  if (i < 0 || i >= static_cast<int>(leaves_.size())) {
+    throw ParquetException("Invalid column id " + std::to_string(i));
+  }
   return &leaves_[i];
 }
 
+const schema::NodePtr& SchemaDescriptor::base(int i) const {
+  if (i < 0 || i >= static_cast<int>(leaves_.size())) {
+    throw ParquetException("Invalid column id " + std::to_string(i));
+  }
```
wesm: I'm a bit -0 on boundschecking / throwing an exception in these cases. We could use `DCHECK` instead since these would indicate a bug in client code rather than an exception that you would want to catch. 

@xhochy do you have any opinion on this?

majetideepak: I agree. `DCHECK` makes sense.

src/parquet/schema/descriptor.h
majetideepak: How does `GetColumnRoot` sound ?

wesm:
+1, thank you!

src/parquet/schema/descriptor.cc
xhochy: +1 for `DCHECK`

PR #98, opened 2016-05-08
PARQUET-604: Add writer headers to installation

xhochy:


wesm:
+1, thanks

wesm:
We're also going to want to create a public API for file writing https://github.com/apache/parquet-cpp/tree/master/src/parquet/api

I'm sort of lukewarm on the `api` directory; I would be fine with a `parquet/parquet.h` (or `parquet/api.h`) with everything public in it. let me know what you think

xhochy:
I'm okish with the `api` directory but would be -1 on the single `parquet/parquet.h` header. We already have some name collisions between Arrow and Parquet and although we have them nicely split up into namespaces, I would prefer to import only the really needed declarations.

PR #101, opened 2016-05-09
PARQUET-606: Move back to Ubuntu 14.04 builds, install LLVM manually

wesm:
This fixes the coverage builds.


majetideepak:
+1 LGTM. Thank you!

wesm:
@emkornfield we can do the same in Arrow to move back to trusty builds, this would also allow us to use LLVM 3.8 if so desired

PR #103, opened 2016-05-12
PARQUET-614: Remove unneeded LZ4-related code

wesm:
The LZ4 algorithm is not used in the Parquet format. We will need to make dynamic calls to liblzo in a form that does not cause GPL complications when we do implement the LZO compression type.


xhochy:
+1, thanks!

PR #104, opened 2016-05-13
PARQUET-613: Add a conda (packaging tool) recipe

wesm:
Depends on PARQUET-614. With appropriate conda-forge (https://conda-forge.github.io/) patches conda-forge/snappy-feedstock#1 and  conda-forge/zlib-feedstock#4, I'm able to build a conda package for this library. This will make it much easier for Python ecosystem packages that need to use parquet-cpp (e.g. the Python Apache Arrow library).

### DIFF
--- a/thirdparty/download_thirdparty.sh
+++ b/thirdparty/download_thirdparty.sh
@@ -7,11 +7,17 @@ TP_DIR=$(cd "$(dirname "${BASH_SOURCE:-$0}")"; pwd)
 
 source $TP_DIR/versions.sh
 
+: ${PARQUET_INSECURE_CURL=0}
+
 download_extract_and_cleanup() {
-	filename=$TP_DIR/$(basename "$1")
-	curl -L -k "$1" -o $filename
-	tar xzf $filename -C $TP_DIR
-	rm $filename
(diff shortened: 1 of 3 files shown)


wesm:
rebased

thirdparty/download_thirdparty.sh
```diff
@@ -9,7 +9,7 @@ source $TP_DIR/versions.sh
 
 download_extract_and_cleanup() {
 	filename=$TP_DIR/$(basename "$1")
-	curl -#LC - "$1" -o $filename
+	curl -L -k "$1" -o $filename
```
xhochy: ` -k, --insecure      Allow connections to SSL sites without certs (H)` not so happy about that. Is this really needed for conda? Can we e.g. do this via a flag?

xhochy:
I'm not an expert for conda, so the only part of the review I can make is the curl command.

thirdparty/download_thirdparty.sh
wesm: I hear you. Let me add an environment variable for this

wesm:
Done. will wait for a green build

wesm:
+1, thank you

PR #106, opened 2016-05-15
PARQUET-617: Export CC / CXX / LD_LIBRARY_PATH in conda build for custom C++ toolchains

wesm:
If you are building the conda package in an environment with multiple GCC toolchains, with the preferred one being one other than the default one, you must explicitly pass these environment variables down in the conda recipe. The `LD_LIBRARY_PATH` is also passed down for finding the gcc stdlib libraries when running the unit tests to verify the package.

### DIFF
--- a/conda.recipe/build.sh
+++ b/conda.recipe/build.sh
@@ -14,6 +14,7 @@ export ZLIB_HOME=$PREFIX
 
 cd ..
 
+rm -rf conda-build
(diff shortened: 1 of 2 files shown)


wesm:
This is a small bug fix.

xhochy:
+1, thanks!

wesm:
thanks!

PR #107, opened 2016-05-17
PARQUET-619: Add OutputStream for local files

xhochy:


xhochy:
Fixed all PR names I made in the last hours, seems like I forgot to prefix them with the JIRA issue :D

wesm:
Can you add a couple unit tests for this?

xhochy:
* Added unittest for normal usage
* I did not see how to add unit test for the cases where we would throw an exception. An approach could have been to do "WriteAfterClose" but that causes memory corruption. Instead I added DCHECKs at the beginning of the `Write` and `Tell` to check for `is_open_`.

wesm:
Cool, that works. Thanks for adding the tests. +1

PR #108, opened 2016-05-17
PARQUET-620: Ensure metadata is written only once

xhochy:


wesm:
lgtm, I suppose we'll eventually want to add test coverage for this

wesm:
+1, thank you

PR #109, opened 2016-05-17
PARQUET-621: Add flag to indicate if decimalmetadata is set

xhochy:


wesm:
Can you add a test case?

xhochy:
Added a test

wesm:
+1, thank you very much

PR #112, opened 2016-05-29
PARQUET-615: Allow building static and shared library in parallel

xhochy:


wesm:
Sweet, thanks for looking into this.

I haven't verified, but I believe we need position independent code also in the static library, otherwise it will cause linker problems if you try linking `libparquet.a` into another shared library build. I dealt with this recently in conda-forge builds for snappy and zlib (since we build the static libraries in our third party with `-fPIC`) -- e.g. https://github.com/conda-forge/zlib-feedstock/pull/4

per the coverage thing, it will leave some cruft in https://github.com/apache/parquet-cpp/blob/master/CMakeLists.txt#L414 -- this should probably use the `PARQUET_LINK` variable to determine whether to use static linking with the built executables. That way we can use dynamic linking by default and also default to building both the shared and static libs

wesm:
Argh: http://llvm.org/apt/  is down. Looks like we have to figure out some other more reliable way to get the LLVM toolchain (maybe switching back to precise)

xhochy:
Coverage part: Should we link the test executable preferable static and then we're fine?

wesm:
Static linking will increase build times -- for coverage of course we need to use static everywhere internally for executables

xhochy:
Static Builds for unittests are preferred and the coverage check fails if we do not build the static lib, so everything except travis/LLVM should be done.

wesm:
@xhochy if you rebase all your PRs we should get passing builds now, and I'll get things merged up

xhochy:
Successfully rebased.

wesm:
excellent. +1, thank you

xhochy:
Somehow this broke building of `parquet/thrift/parquet_types.h` for the conda packages. Did notice this as these build where accepted failures. 

This problem only seems to happen with `release` builds and `make` as build system (using `ninja` instead works).

PR #113, opened 2016-05-31
PARQUET-598: Test writing all primitive data types

xhochy:


src/parquet/column/column-writer-test.cc
```diff
@@ -95,105 +132,152 @@ class TestPrimitiveWriter : public ::testing::Test {
...
+  std::copy(values_.begin(), values_.end(), bool_buffer_);
+  values_ptr_ = bool_buffer_;
+}
+
+typedef ::testing::Types<Int32Type, Int64Type, Int96Type, FloatType, DoubleType,
+    BooleanType, ByteArrayType, FLBAType> TestTypes;
+
+TYPED_TEST_CASE(TestPrimitiveWriter, TestTypes);
+
+TYPED_TEST(TestPrimitiveWriter, RequiredNonRepeated) {
```
majetideepak: Use the current naming convention `Required` instead of `RequiredNonRepeated`?

src/parquet/column/column-writer-test.cc
```diff
@@ -95,105 +132,152 @@ class TestPrimitiveWriter : public ::testing::Test {
...
-  ReadColumn();
-  ASSERT_EQ(100, values_read_);
-  ASSERT_EQ(values, values_out_);
+  this->ReadColumn();
+  ASSERT_EQ(100, this->values_read_);
+  ASSERT_EQ(this->values_, this->values_out_);
 }
 
-TEST_F(TestPrimitiveWriter, OptionalNonRepeated) {
+TYPED_TEST(TestPrimitiveWriter, OptionalNonRepeated) {
```
majetideepak: Same as above. `Optional` instead of `OptionalNonRepeated`

src/parquet/column/column-writer-test.cc
```diff
@@ -95,105 +132,152 @@ class TestPrimitiveWriter : public ::testing::Test {
...
-  ASSERT_EQ(values, values_out_);
+  this->ReadColumn();
+  ASSERT_EQ(99, this->values_read_);
+  this->values_out_.resize(99);
+  this->values_.resize(99);
+  ASSERT_EQ(this->values_, this->values_out_);
 }
 
-TEST_F(TestPrimitiveWriter, OptionalRepeated) {
+TYPED_TEST(TestPrimitiveWriter, OptionalRepeated) {
```
majetideepak: `Repeated` instead of `OptionalRepeated`

src/parquet/column/column-writer-test.cc
```diff
@@ -95,105 +132,152 @@ class TestPrimitiveWriter : public ::testing::Test {
...
-  auto writer = BuildWriter();
-  ASSERT_THROW(writer->WriteBatch(values.size(), nullptr, nullptr, values.data()),
+  auto writer = this->BuildWriter();
+  ASSERT_THROW(
+      writer->WriteBatch(this->values_.size(), nullptr, nullptr, this->values_ptr_),
       ParquetException);
 }
 
-TEST_F(TestPrimitiveWriter, OptionalRepeatedTooFewRows) {
+TYPED_TEST(TestPrimitiveWriter, OptionalRepeatedTooFewRows) {
```
majetideepak: `Repeated` instead of `OptionalRepeated`

src/parquet/column/column-writer-test.cc
```diff
@@ -32,59 +35,93 @@ using schema::PrimitiveNode;
...
   }
 
+  void GenerateData(int64_t num_values);
+
+  void SetupValuesOut();
+
   void SetUp() {
-    values_out_.resize(100);
+    SetupValuesOut();
     definition_levels_out_.resize(100);
```
wesm: can you declare a constant for the size and use that everywhere 100 is currently hardcoded?

src/parquet/column/column-writer-test.cc
```diff
@@ -32,59 +35,93 @@ using schema::PrimitiveNode;
...
-  std::unique_ptr<Int64Reader> BuildReader() {
+  void TearDown() {
+    if (bool_buffer_) {
+      delete[] bool_buffer_;
+      bool_buffer_ = nullptr;
+    }
+    if (bool_buffer_out_) {
+      delete[] bool_buffer_out_;
+      bool_buffer_out_ = nullptr;
+    }
```
wesm: it is preferable to use `std::vector<uint8_t>` instead of `new / delete`

src/parquet/column/test-specialization.h
```diff
@@ -0,0 +1,66 @@
...
+// Parquet column chunk within a row group. It could be extended in the future
+// to iterate through all data pages in all chunks in a file.
+
+#ifndef PARQUET_COLUMN_TEST_SPECIALIZATION_H
+#define PARQUET_COLUMN_TEST_SPECIALIZATION_H
+
+#include <limits>
+#include <vector>
+
+#include "parquet/column/test-util.h"
```
wesm: Include this code in `test-util.h`? (can also rename this "test-common.h" if you like)

wesm:
Minor comments, but this is great, thanks.

src/parquet/column/test-specialization.h
xhochy: No, I actually pulled this out of `test-util.h` as it caused `unused function` errors with `-Werror`.

xhochy:
Addressed comments, Travis failed due to apt-Problems (can someone retrigge it?)

majetideepak:
You will have to push an empty commit. AFAIK there is no other way.

wesm:
@xhochy what I usually do in these cases is `git commit --amend` to bump the git hash without committing anything new

xhochy:
Successfully rebased.

wesm:
+1, thank you

PR #114, opened 2016-06-02
PARQUET-626: Disable LLVM toolchain parts of Travis CI build

wesm:
Since llvm.org apt repo was taken offline, this patch triages builds until we sort out an alternative (or remove it altogether for now).


wesm:
+1 -- I am merging this as a triage so we can get passing builds for the outstanding patch queue

PR #117, opened 2016-06-06
PARQUET-629: RowGroupSerializer should only close itself once

xhochy:


src/parquet/file/writer-internal.cc
```diff
@@ -148,16 +148,19 @@ ColumnWriter* RowGroupSerializer::NextColumn() {
 }
 
 void RowGroupSerializer::Close() {
-  if (current_column_index_ != schema_->num_columns() - 1) {
-    throw ParquetException("Not all column were written in the current rowgroup.");
+  if (!closed_) {
```
wesm: Perhaps we should have a DCHECK(!closed_), unless there's a reason we want this to be idempotent?

xhochy: I'd prefer to have close idempotent. Thus you can call it explicitly in the code or implicitly in the destructor.

wesm: Works for em

wesm:
+1, thank you

PR #119, opened 2016-06-16
PARQUET-633: Add version to WriterProperties

xhochy:
Yet only exposing the property publicly so I can make use of it in Arrow. 

Also used the interface design from parquet-mr to make the `WriterProperties` immutable so we can safely pass references of them around once they have been passed to a `ParquetFileWriter`.

### DIFF
--- a/src/parquet/column/writer.cc
+++ b/src/parquet/column/writer.cc
@@ -25,8 +25,9 @@ namespace parquet {
 // ----------------------------------------------------------------------
 // ColumnWriter
 
-WriterProperties default_writer_properties() {
-  static WriterProperties default_writer_properties;
+std::shared_ptr<WriterProperties> default_writer_properties() {
+  static std::shared_ptr<WriterProperties> default_writer_properties =
+      WriterProperties::Builder().build();
   return default_writer_properties;
 }
 
--- a/src/parquet/file/writer-internal.cc
+++ b/src/parquet/file/writer-internal.cc
@@ -168,9 +168,9 @@ void RowGroupSerializer::Close() {
 
 std::unique_ptr<ParquetFileWriter::Contents> FileSerializer::Open(
     std::shared_ptr<OutputStream> sink, std::shared_ptr<GroupNode>& schema,
-    MemoryAllocator* allocator) {
+    MemoryAllocator* allocator, std::shared_ptr<WriterProperties> properties) {
   std::unique_ptr<ParquetFileWriter::Contents> result(
-      new FileSerializer(sink, schema, allocator));
+      new FileSerializer(sink, schema, allocator, properties));
 
   return result;
 }
@@ -198,6 +198,10 @@ int FileSerializer::num_row_groups() const {
 
 int64_t FileSerializer::num_rows() const {
   return num_rows_;
+}
+
+std::shared_ptr<WriterProperties> FileSerializer::properties() const {
+  return properties_;
 }
 
 RowGroupWriter* FileSerializer::AppendRowGroup(int64_t num_rows) {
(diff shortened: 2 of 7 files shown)


src/parquet/file/writer.cc
```diff
@@ -55,7 +55,7 @@ ParquetFileWriter::~ParquetFileWriter() {
 
 std::unique_ptr<ParquetFileWriter> ParquetFileWriter::Open(
     std::shared_ptr<OutputStream> sink, std::shared_ptr<GroupNode>& schema,
-    MemoryAllocator* allocator) {
+    MemoryAllocator* allocator, std::shared_ptr<WriterProperties> properties) {
```
wesm: const reference here

src/parquet/file/writer.cc
```diff
@@ -80,4 +80,8 @@ RowGroupWriter* ParquetFileWriter::AppendRowGroup(int64_t num_rows) {
   return contents_->AppendRowGroup(num_rows);
 }
 
+std::shared_ptr<WriterProperties> ParquetFileWriter::properties() const {
```
wesm: return const reference

src/parquet/file/writer.h
```diff
@@ -93,6 +94,8 @@ class ParquetFileWriter {
     virtual int num_columns() const = 0;
     virtual int num_row_groups() const = 0;
 
+    virtual std::shared_ptr<WriterProperties> properties() const = 0;
```
wesm: const reference

src/parquet/file/writer.h
```diff
@@ -103,7 +106,8 @@ class ParquetFileWriter {
 
   static std::unique_ptr<ParquetFileWriter> Open(std::shared_ptr<OutputStream> sink,
       std::shared_ptr<schema::GroupNode>& schema,
-      MemoryAllocator* allocator = default_allocator());
+      MemoryAllocator* allocator = default_allocator(),
+      std::shared_ptr<WriterProperties> properties = default_writer_properties());
```
wesm: I guess we could benefit from some const refs in other parameters in this function

src/parquet/file/writer.h
```diff
@@ -139,6 +143,11 @@ class ParquetFileWriter {
   int num_row_groups() const;
 
   /**
+   * Configuartion passed to the writer, e.g. the used Parquet format version.
+   */
+  std::shared_ptr<WriterProperties> properties() const;
```
wesm: typo

src/parquet/file/writer-internal.h
```diff
@@ -124,7 +127,8 @@ class FileSerializer : public ParquetFileWriter::Contents {
 
  private:
   explicit FileSerializer(std::shared_ptr<OutputStream> sink,
-      std::shared_ptr<schema::GroupNode>& schema, MemoryAllocator* allocator);
+      std::shared_ptr<schema::GroupNode>& schema, MemoryAllocator* allocator,
+      std::shared_ptr<WriterProperties> properties);
```
wesm: make all these shared ptrs const refs

src/parquet/file/writer-internal.cc
```diff
@@ -200,6 +200,10 @@ int64_t FileSerializer::num_rows() const {
   return num_rows_;
 }
 
+std::shared_ptr<WriterProperties> FileSerializer::properties() const {
+  return properties_;
+}
```
wesm: return `const std::shared_ptr<T>&`

src/parquet/column/writer.cc
```diff
@@ -25,8 +25,9 @@ namespace parquet {
 // ----------------------------------------------------------------------
 // ColumnWriter
 
-WriterProperties default_writer_properties() {
-  static WriterProperties default_writer_properties;
+std::shared_ptr<WriterProperties> default_writer_properties() {
```
wesm: return `const std:shared_ptr<T>&`, saves extra copy

wesm:
Cool, I'm +1 on immutability. Left a few comments around avoiding unnecessary shared pointer copies, but otherwise this can go in

wesm:
Excellent, thank you. +1

PR #120, opened 2016-06-17
PARQUET-634: Consistent private linking of dependencies

xhochy:
Link all (static) third-party dependencies privately to not expose their API.

Also combine all static libs containing public interfaces into one. To correctly link against parquet_static from another lib, you still need all thirdparty static libs but at least the main parquet static library is only a single one. Will address the scenario "single static lib with all dependencies included" later once we have visibility macros.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -448,6 +448,10 @@ set(LIBPARQUET_SRCS
   src/parquet/column/writer.cc
   src/parquet/column/scanner.cc
 
+  src/parquet/compression/codec.cc
+  src/parquet/compression/snappy-codec.cc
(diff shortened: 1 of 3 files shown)


wesm:
Eventually we should add a comment that uses `ar` to produce a "mega static library" to not place that burden on static linkers necessarily. I wonder what are the dependency hell implications, e.g.

- project A
  - depends on project B
    - depends on project C, version 1
  - depends on project D
    - depends on project C, version 2

So conceivably, you might need two different static versions of C to produce a fully statically linked binary. I don't know enough about this problem to know the right way to handle it

wesm:
+1

PR #121, opened 2016-06-18
PARQUET-592: Support compressed writes

lomereiter:
Hello,

I'm working on compressed writes and would like some advice:

* What would be the preferred way to pass down per-column codec settings? `Builder::set_column_codec(column index (?), Compression::type)`? 
* It appears that there's some duplication in `ParquetFileWriter::Open`, namely, there is an allocator included into `WriterProperties` — I presume this should be cleaned up via a separate PR?

### DIFF
--- a/src/parquet/column/properties.h
+++ b/src/parquet/column/properties.h
@@ -160,7 +160,7 @@ class WriterProperties {
     ColumnCodecs codecs_;
   };
 
-  MemoryAllocator* allocator() { return allocator_; }
+  MemoryAllocator* allocator() const { return allocator_; }
 
   bool dictionary_enabled() const { return dictionary_enabled_; }
 
--- a/src/parquet/column/properties.h
+++ b/src/parquet/column/properties.h
@@ -147,7 +147,7 @@ class WriterProperties {
     std::shared_ptr<WriterProperties> build() {
       return std::shared_ptr<WriterProperties>(new WriterProperties(
           allocator_, dictionary_enabled_, dictionary_pagesize_,
-          pagesize_, version_, default_codec_, std::move(codecs_)));
+          pagesize_, version_, default_codec_, codecs_));
     }
 
    private:
--- a/src/parquet/column/properties.h
+++ b/src/parquet/column/properties.h
@@ -20,10 +20,12 @@
 
 #include <memory>
 #include <string>
+#include <unordered_map>
 
 #include "parquet/util/input.h"
 #include "parquet/util/mem-allocator.h"
 #include "parquet/types.h"
+#include "parquet/schema/types.h"
 
 namespace parquet {
 
@@ -79,6 +81,10 @@ static int64_t DEFAULT_DICTIONARY_PAGE_SIZE = DEFAULT_PAGE_SIZE;
 static bool DEFAULT_IS_DICTIONARY_ENABLED = true;
 static constexpr ParquetVersion::type DEFAULT_WRITER_VERSION =
     ParquetVersion::PARQUET_1_0;
+static constexpr Compression::type DEFAULT_COMPRESSION_TYPE =
+    Compression::UNCOMPRESSED;
+
+typedef std::unordered_map<std::string, Compression::type> ColumnCodecs;
 
 class WriterProperties {
  public:
@@ -89,8 +95,7 @@ class WriterProperties {
           dictionary_enabled_(DEFAULT_IS_DICTIONARY_ENABLED),
           dictionary_pagesize_(DEFAULT_DICTIONARY_PAGE_SIZE),
           pagesize_(DEFAULT_PAGE_SIZE),
-          version_(DEFAULT_WRITER_VERSION),
-          codec_(Compression::UNCOMPRESSED) {}
+          version_(DEFAULT_WRITER_VERSION) {}
     virtual ~Builder() {}
 
     Builder* allocator(MemoryAllocator* allocator) {
@@ -123,15 +128,16 @@ class WriterProperties {
       return this;
     }
 
-    Builder* compression(Compression::type codec) {
-      codec_ = codec;
+    Builder* set_compression(const std::shared_ptr<schema::ColumnPath>& path,
+                             Compression::type codec) {
+      codecs_[path->ToDotString()] = codec;
       return this;
     }
 
     std::shared_ptr<WriterProperties> build() {
       return std::shared_ptr<WriterProperties>(new WriterProperties(
           allocator_, dictionary_enabled_, dictionary_pagesize_,
-          pagesize_, version_, codec_));
+          pagesize_, version_, codecs_));
     }
 
    private:
@@ -140,7 +146,7 @@ class WriterProperties {
     int64_t dictionary_pagesize_;
     int64_t pagesize_;
     ParquetVersion::type version_;
-    Compression::type codec_;  // TODO: per-column compression settings
+    ColumnCodecs codecs_;
   };
 
   MemoryAllocator* allocator() { return allocator_; }
@@ -153,18 +159,23 @@ class WriterProperties {
 
   ParquetVersion::type version() const { return parquet_version_; }
 
-  Compression::type compression() const { return codec_; }
+  Compression::type compression(const std::shared_ptr<schema::ColumnPath>& path) const {
+    auto it = codecs_.find(path->ToDotString());
+    if (it != codecs_.end())
+      return it->second;
+    return DEFAULT_COMPRESSION_TYPE;
+  }
 
  private:
   explicit WriterProperties(MemoryAllocator* allocator, bool dictionary_enabled,
       int64_t dictionary_pagesize, int64_t pagesize, ParquetVersion::type version,
-      Compression::type codec)
+      ColumnCodecs codecs)
       : allocator_(allocator),
         dictionary_enabled_(dictionary_enabled),
         dictionary_pagesize_(dictionary_pagesize),
         pagesize_(pagesize),
         parquet_version_(version),
-        codec_(codec) {
+        codecs_(codecs) {
     pagesize_ = DEFAULT_PAGE_SIZE;
     dictionary_enabled_ = DEFAULT_IS_DICTIONARY_ENABLED;
   }
@@ -174,7 +185,7 @@ class WriterProperties {
   int64_t dictionary_pagesize_;
   int64_t pagesize_;
   ParquetVersion::type parquet_version_;
-  Compression::type codec_;
+  ColumnCodecs codecs_;
 };
 
 std::shared_ptr<WriterProperties> default_writer_properties();
--- a/src/parquet/column/properties.h
+++ b/src/parquet/column/properties.h
@@ -23,6 +23,7 @@
 
 #include "parquet/util/input.h"
 #include "parquet/util/mem-allocator.h"
+#include "parquet/types.h"
 
 namespace parquet {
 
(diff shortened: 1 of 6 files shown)


xhochy:
Going to review tomorrow more in detail but this looks quite similar to the encoding patch I'm going to push in the next 2 days. For the column-wise encoding, I'm using an `std::unordered_map<std::string,Encoding::type>` where the key is the columnpath as string.

src/parquet/file/file-serialize-test.cc
```diff
@@ -64,37 +64,45 @@ class TestSerialize : public ::testing::Test {
 };
 
 TEST_F(TestSerialize, SmallFile) {
```
xhochy: I guess for debugging reasons, it would be better if this were three separate tests that all call the same function. This should provide the codec that is failing directly in the test outputß

src/parquet/file/writer-internal.h
```diff
@@ -67,12 +67,14 @@ class SerializedPageWriter : public PageWriter {
 class RowGroupSerializer : public RowGroupWriter::Contents {
  public:
   RowGroupSerializer(int64_t num_rows, const SchemaDescriptor* schema, OutputStream* sink,
-      format::RowGroup* metadata, MemoryAllocator* allocator)
+      format::RowGroup* metadata, MemoryAllocator* allocator,
+      const std::shared_ptr<WriterProperties>& properties)
```
xhochy: Passing a reference should be enough, the ParquetFileWriter already owns the properties and this object, thus it has a longer lifetime.

src/parquet/file/writer-internal.cc
```diff
@@ -141,7 +142,7 @@ ColumnWriter* RowGroupSerializer::NextColumn() {
   col_meta->meta_data.__set_type(ToThrift(column_descr->physical_type()));
   col_meta->meta_data.__set_path_in_schema(column_descr->path()->ToDotVector());
   std::unique_ptr<PageWriter> pager(
-      new SerializedPageWriter(sink_, Compression::UNCOMPRESSED, col_meta, allocator_));
+      new SerializedPageWriter(sink_, properties_->compression(), col_meta, allocator_));
```
xhochy: When we have also support for defining compression per column, we should use a function of properties that gets the column path and then returns either the specific codec if one is defined for the column, otherwise the default.

src/parquet/file/writer-internal.cc
```diff
@@ -77,13 +77,14 @@ int64_t SerializedPageWriter::WriteDataPage(int32_t num_rows, int32_t num_values
...
   if (compressor_) {
-    // TODO(PARQUET-592): Add support for compression
-    // int64_t max_compressed_size = compressor_->MaxCompressedLen(
-    // uncompressed_data.size(), uncompressed_data.data());
-    // OwnedMutableBuffer compressed_data(compressor_->MaxCompressedLen(
-    // uncompressed_data.size(), uncompressed_data.data()));
+    uncompressed_ptr = uncompressed_data->mutable_data();
+    int64_t max_compressed_size = compressor_->MaxCompressedLen(
+        uncompressed_size, uncompressed_ptr);
+    compressed_data = std::make_shared<OwnedMutableBuffer>(max_compressed_size);
```
xhochy: Need to pass the allocator here

xhochy:
For the allocator as redundant argument: yes, we should remove that. It has been there already before we introduced the WriterProperties.

src/parquet/column/properties.h
```diff
@@ -121,9 +128,16 @@ class WriterProperties {
       return this;
     }
 
+    Builder* set_compression(const std::shared_ptr<schema::ColumnPath>& path,
```
xhochy: Can you re-add the version that sets a default for all columns for which not an explicit codec is set?
Also please remove the `set_` prefix to. Be in line with the other methods. 

We might also want to have these methods with a string as the first argument as this will be easier to use for the end-user.

src/parquet/column/properties.h
```diff
@@ -142,16 +157,25 @@ class WriterProperties {
...
+    if (it != codecs_.end())
+      return it->second;
+    return DEFAULT_COMPRESSION_TYPE;
+  }
 
  private:
   explicit WriterProperties(MemoryAllocator* allocator, bool dictionary_enabled,
-      int64_t dictionary_pagesize, int64_t pagesize, ParquetVersion::type version)
+      int64_t dictionary_pagesize, int64_t pagesize, ParquetVersion::type version,
+      ColumnCodecs codecs)
```
xhochy: Make `codecs` a const reference to save a copy.

src/parquet/file/file-serialize-test.cc
```diff
@@ -97,6 +102,18 @@ TEST_F(TestSerialize, SmallFile) {
   ASSERT_EQ(values, values_out);
 }
 
+TEST_F(TestSerialize, SmallFileUncompressed) {
+  FileSerializeTest(Compression::UNCOMPRESSED, node_);
```
xhochy: Why do you pass `node_` explicitly here?

src/parquet/file/writer.cc
```diff
@@ -55,8 +55,8 @@ ParquetFileWriter::~ParquetFileWriter() {
 
 std::unique_ptr<ParquetFileWriter> ParquetFileWriter::Open(
     std::shared_ptr<OutputStream> sink, std::shared_ptr<GroupNode>& schema,
-    MemoryAllocator* allocator, const std::shared_ptr<WriterProperties>& properties) {
-  auto contents = FileSerializer::Open(sink, schema, allocator);
+    MemoryAllocator* allocator, const std::shared_ptr<WriterProperties>& props) {
```
xhochy: Please don't abbreviate properties here.

xhochy:
Looks good, some minor nitpicks but the it is ready to merge.

src/parquet/file/file-serialize-test.cc
lomereiter: Presumably because I'm unfamiliar with GTest. How should it be passed to the function?

src/parquet/file/writer.cc
lomereiter: Ok. I've seen `props` in other places, e.g. https://github.com/apache/parquet-cpp/blob/4c7b5f2e46160969323bcdf3273b1d6b8ff32d53/src/parquet/file/reader-internal.h#L73 and https://github.com/apache/parquet-cpp/blob/4c7b5f2e46160969323bcdf3273b1d6b8ff32d53/src/parquet/file/reader-internal.h#L103 and tried to maintain consistent style.

src/parquet/file/file-serialize-test.cc
xhochy: Ah, I see the problem. You need to make FileSerialzeTest a member function of the TestSerialze class.

src/parquet/file/writer.cc
xhochy: Saving only a tiny number of characrters is not really a benefit, we rather stick to the full word for better code readability.

src/parquet/column/properties.h
lomereiter: I would need to `std::move` the codecs in `Builder::build()` as well?

src/parquet/column/properties.h
```diff
@@ -128,16 +129,25 @@ class WriterProperties {
...
+    Builder* compression(const std::shared_ptr<schema::ColumnPath>& path,
+                         Compression::type codec) {
+      return this->compression(path->ToDotString(), codec);
+    }
+
     std::shared_ptr<WriterProperties> build() {
       return std::shared_ptr<WriterProperties>(new WriterProperties(
           allocator_, dictionary_enabled_, dictionary_pagesize_,
-          pagesize_, version_, codecs_));
+          pagesize_, version_, default_codec_, std::move(codecs_)));
```
xhochy: If we move, the builder can only be used once. We probably want to have it reusable. 

@wesm @majetideepak got an opinion on that?

majetideepak: I agree with @xhochy . I don't think we are gaining much here anyway. The `std::move(unordered_map)`  will `copy` the keys and just `move` the values (compression enums here). For typical `num_columns()`, it might be okay to copy the unordered map for the advantage of reusability.

majetideepak:
+1

src/parquet/column/properties.h
lomereiter: Fair enough, replaced it with simply a copy.

src/parquet/file/writer-internal.h
```diff
@@ -67,12 +67,14 @@ class SerializedPageWriter : public PageWriter {
 class RowGroupSerializer : public RowGroupWriter::Contents {
  public:
   RowGroupSerializer(int64_t num_rows, const SchemaDescriptor* schema, OutputStream* sink,
-      format::RowGroup* metadata, MemoryAllocator* allocator)
+      format::RowGroup* metadata, MemoryAllocator* allocator,
```
majetideepak: Feel free to remove the duplicate allocator argument in this patch. I think it is cascaded right from `ParquetFileWriter::Open()`

src/parquet/column/properties.h
```diff
@@ -78,6 +81,10 @@ static int64_t DEFAULT_DICTIONARY_PAGE_SIZE = DEFAULT_PAGE_SIZE;
 static bool DEFAULT_IS_DICTIONARY_ENABLED = true;
 static constexpr ParquetVersion::type DEFAULT_WRITER_VERSION =
     ParquetVersion::PARQUET_1_0;
+static constexpr Compression::type DEFAULT_COMPRESSION_TYPE =
+    Compression::UNCOMPRESSED;
+
+typedef std::unordered_map<std::string, Compression::type> ColumnCodecs;
```
wesm: Note: it's preferable to use `using` in C++11

src/parquet/file/file-serialize-test.cc
```diff
@@ -61,40 +61,55 @@ class TestSerialize : public ::testing::Test {
  protected:
   NodePtr node_;
   SchemaDescriptor schema_;
+
+  void FileSerializeTest(Compression::type codec_type) {
+    std::shared_ptr<InMemoryOutputStream> sink(new InMemoryOutputStream());
+    auto gnode = std::static_pointer_cast<GroupNode>(node_);
+    std::shared_ptr<WriterProperties> writer_properties =
+        std::make_shared<WriterProperties::Builder>()
```
wesm: I think you can stack allocate to avoid the extra gymnastics, but no need to change

src/parquet/file/writer-internal.cc
```diff
@@ -77,13 +77,15 @@ int64_t SerializedPageWriter::WriteDataPage(int32_t num_rows, int32_t num_values
...
-    // uncompressed_data.size(), uncompressed_data.data());
-    // OwnedMutableBuffer compressed_data(compressor_->MaxCompressedLen(
-    // uncompressed_data.size(), uncompressed_data.data()));
+    uncompressed_ptr = uncompressed_data->mutable_data();
+    int64_t max_compressed_size = compressor_->MaxCompressedLen(
+        uncompressed_size, uncompressed_ptr);
+    compressed_data = std::make_shared<OwnedMutableBuffer>(max_compressed_size,
+        allocator_);
+    compressed_size = compressor_->Compress(uncompressed_size, uncompressed_ptr,
+        max_compressed_size, compressed_data->mutable_data());
```
wesm: Can you use `uncompressed_data->data()` (since const is okay -- we are only mutating compressed_data)?

wesm:
Only that one nitpick from me and this is good to go. Thanks for doing this

xhochy:
+1, LGTM

Thanks for the contribution!

wesm:
+1, thank you

lomereiter:
Thanks for all the nitpicking :)

leogouttefarde:
One issue I noticed in the file-serialize-test : building properties with the column path "schema.int64" for compression does not work (data is not getting compressed), I had to remove the "schema." prefix for it to work (even though the GroupNode is created with the "schema" prefix).

So in parquet-cpp/src/parquet/file/file-serialize-test.cc :
This line : WriterProperties::Builder().compression("schema.int64", codec_type)->build();
Needs to be : WriterProperties::Builder().compression("int64", codec_type)->build();

The test itself is not failing because it does not check that compression is actually happening, it just runs on uncompressed data every time.

wesm:
Thanks for catching this! Can you create a JIRA on issues.apache.org with this bug report?

leogouttefarde:
Ok https://issues.apache.org/jira/browse/PARQUET-673

PR #122, opened 2016-06-20
PARQUET-636: Expose selection for different encodings

xhochy:
Yet we still only support PLAIN encoding. Will implement the other encodings in separate PRs to not have huge changesets.

### DIFF
--- a/src/parquet/column/properties.h
+++ b/src/parquet/column/properties.h
@@ -22,10 +22,10 @@
 #include <string>
 #include <unordered_map>
 
-#include "parquet/util/input.h"
-#include "parquet/util/mem-allocator.h"
 #include "parquet/types.h"
 #include "parquet/schema/types.h"
+#include "parquet/util/input.h"
+#include "parquet/util/mem-allocator.h"
 
 namespace parquet {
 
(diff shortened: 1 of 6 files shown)


src/parquet/column/properties.h
```diff
@@ -139,21 +155,23 @@ class WriterProperties {
...
+          dictionary_enabled_, dictionary_pagesize_, default_encoding_, encodings_,
           pagesize_, version_, default_codec_, codecs_));
     }
 
    private:
     MemoryAllocator* allocator_;
     bool dictionary_enabled_;
     int64_t dictionary_pagesize_;
+    Encoding::type default_encoding_;
+    std::unordered_map<std::string, Encoding::type> encodings_;
```
wesm: Can you add a comment explaining this member variable?

wesm: (same probably for `codecs_`)

wesm:
Looks good outside minor comment. We can spare implementing the Parquet 2.0 encodings for now.

src/parquet/column/properties.h
lomereiter: Or perhaps refactor these pairs of variables into instances of `PerColumnProperties<Encoding::type/Compression::type>` to avoid duplicating comments.

wesm:
+1, will merge on green build

xhochy:
Build is green ;)

PR #123, opened 2016-06-20
PARQUET-635: Use conda-forge toolchain, statically link libstdc++ on Linux

wesm:
This will avoid portability issues across different Linux environments.

### DIFF
--- a/conda.recipe/build.sh
+++ b/conda.recipe/build.sh
@@ -3,6 +3,14 @@
 set -e
 set -x
 
+# FIXME: This is a hack to make sure the environment is activated.
+# The reason this is required is due to the conda-build issue
+# mentioned below.
+#
+# https://github.com/conda/conda-build/issues/910
+#
+source activate "${CONDA_DEFAULT_ENV}"
+
 cd $RECIPE_DIR
 
 # Build dependencies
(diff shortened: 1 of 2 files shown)


xhochy:
+1, LGTM

conda.recipe/build.sh
```diff
@@ -50,9 +45,16 @@ export PARQUET_INSECURE_CURL=1
 source thirdparty/versions.sh
 export GTEST_HOME=`pwd`/thirdparty/$GTEST_BASEDIR
 
+if [ `uname` == Linux ]; then
+    SHARED_LINKER_FLAGS='-static-libstdc++'
```
lomereiter: Please add `-static-libgcc` as well.

xhochy: As far as I know, this is really discouraged. e.g. from the GCC Manual:

> There are several situations in which an application should use the shared libgcc instead of the static version. The most common of these is when the application wishes to throw and catch exceptions across different shared libraries. In that case, each of the libraries as well as the application itself should use the shared libgcc.

https://gcc.gnu.org/onlinedocs/gcc/Link-Options.html

lomereiter: Thanks, I wasn't aware of this.

PR #124, opened 2016-06-23
PARQUET-638: Temporarily disable static linking libstdc++ in the conda builds

wesm:
We can re-enable this after we've addressed visibility macros.


wesm:
+1, merging as this is blocking Arrow builds

PR #125, opened 2016-06-23
PARQUET-640: Ensure that gcc 4.9 is used in conda builds

wesm:
The toolchain change in https://github.com/apache/parquet-cpp/commit/f334a8b97cd2499799cc74262a52191d04692773 saved a bunch of OS X boilerplate but caused the artifacts to get built with gcc 4.8 by accident.


wesm:
+1, merging as this was a regression before introducing the conda-forge toolchain build dependency

PR #127, opened 2016-06-24
PARQUET-639: Do not export DCHECK in public headers

wesm:
I added a test so that DCHECK does not leak in the public headers. I prefer this to renaming the macro

### DIFF
--- a/src/parquet/public-api-test.cc
+++ b/src/parquet/public-api-test.cc
@@ -20,12 +20,19 @@
 #include "parquet/api/io.h"
 #include "parquet/api/reader.h"
 #include "parquet/api/schema.h"
+#include "parquet/api/writer.h"
 
 namespace parquet {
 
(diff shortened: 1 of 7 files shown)


src/parquet/public-api-test.cc
```diff
@@ -29,4 +30,10 @@ TEST(TestPublicAPI, DoesNotIncludeThrift) {
 #endif
 }
 
+TEST(TestPublicAPI, DoesNotExportDCHECK) {
+#ifdef DCHECK
+  FAIL() << "parquet/util/logging.h should not transitively included";
```
majetideepak: extreme nitpicking: "should not be transitively included"

majetideepak:
+1 LGTM

wesm:
Thanks. +1

wesm:
Rebased. Waiting for green build

PR #129, opened 2016-06-26
PARQUET-593: Add API for writing Page statistics

lomereiter:


src/parquet/column/page.h
```diff
@@ -169,14 +170,16 @@ class PageWriter {
  public:
   virtual ~PageWriter() {}
 
-  virtual void Close() = 0;
+  // TODO think of a better way to pass column chunk statistics
+  virtual void Close(const std::shared_ptr<RowGroupStatistics>& chunk_statistics) = 0;
```
lomereiter: The API is admittedly rather ugly here, because PageWriter operates on raw bytes and not values. It could handle _column chunk_ statistics by itself, but then it would have to be initialized somewhere with a concrete type.

src/parquet/column/statistics.cc
```diff
@@ -0,0 +1,192 @@
...
+#include "parquet/util/output.h"
+#include "parquet/exception.h"
+
+namespace parquet {
+
+template <typename T>
+struct Compare {
+  explicit Compare(const ColumnDescriptor* descr)
+    : type_length_(descr->type_length()) {}
+
```
lomereiter: If not for `FLBAType`, I'd just define `operator<` for all the types.

src/parquet/column/statistics.h
```diff
@@ -0,0 +1,133 @@
...
+
+namespace parquet {
+
+class RowGroupStatistics {
+ public:
+  int64_t NullCount() const { return null_count_; }
+  int64_t DistinctCount() const { return distinct_count_; }
+  int64_t NumValues() const { return num_values_; }
+
+  virtual bool HasMinMax() const = 0;
```
lomereiter: That's a kind of crappy substitute for C++17 `std::optional` which I avoid for the time being.

src/parquet/file/reader-internal.cc
```diff
@@ -171,16 +171,37 @@ std::unique_ptr<PageReader> SerializedRowGroup::GetColumnPageReader(int i) {
...
 
-  RowGroupStatistics result;
-  result.num_values = meta_data.num_values;
-  result.null_count = meta_data.statistics.null_count;
-  result.distinct_count = meta_data.statistics.distinct_count;
-  result.max = &meta_data.statistics.max;
-  result.min = &meta_data.statistics.min;
-  return result;
+std::shared_ptr<RowGroupStatistics> SerializedRowGroup::GetColumnStats(int i) const {
+  const format::ColumnMetaData& meta_data = metadata_->columns[i].meta_data;
```
lomereiter: This really belongs to a typed `ColumnReader`, but for now I just keep the old interface.

src/parquet/column/statistics.cc
```diff
@@ -0,0 +1,185 @@
...
+template <typename DType>
+std::string TypedRowGroupStatistics<DType>::EncodedMax() {
+  std::string s;
+  if (HasMinMax()) this->PlainEncode(max_, s);
+  return s;
+}
+
+template <typename DType>
+void TypedRowGroupStatistics<DType>::PlainEncode(const T& src, std::string& dst) {
```
wesm: Use pointer for the out parameter (`std::string*`)

src/parquet/column/statistics.cc
```diff
@@ -0,0 +1,185 @@
...
+  InMemoryOutputStream output_stream(sizeof(T), allocator_);
+  PlainEncoder<DType> encoder(Schema(), allocator_);
+  encoder.Encode(&src, 1, &output_stream);
+  auto buffer = output_stream.GetBuffer();
+  auto ptr = reinterpret_cast<const char*>(buffer->data());
+  dst.assign(ptr, buffer->size());
+}
+
+template <typename DType>
+void TypedRowGroupStatistics<DType>::PlainDecode(const std::string& src, T& dst) {
```
wesm: `T* dst`

src/parquet/column/statistics.h
```diff
@@ -0,0 +1,127 @@
...
+#include "parquet/schema/descriptor.h"
+#include "parquet/util/mem-allocator.h"
+
+namespace parquet {
+
+class RowGroupStatistics {
+ public:
+  int64_t NullCount() const { return null_count_; }
+  int64_t DistinctCount() const { return distinct_count_; }
+  int64_t NumValues() const { return num_values_; }
```
wesm: use `null_count()`, `distinct_count()`, etc. for these names. see https://google.github.io/styleguide/cppguide.html#Function_Names

src/parquet/column/statistics.h
```diff
@@ -0,0 +1,127 @@
...
+
+  // Plain-encoded maximum value
+  virtual std::string EncodedMax() = 0;
+
+  virtual ~RowGroupStatistics() {}
+
+  Type::type PhysicalType() const { return descr_->physical_type(); }
+
+ protected:
+  const ColumnDescriptor* Schema() const { return descr_; }
```
wesm: either `descr()` or `schema()` for this function name

src/parquet/column/statistics.h
```diff
@@ -0,0 +1,127 @@
...
+
+  // Plain-encoded minimum value
+  virtual std::string EncodedMin() = 0;
+
+  // Plain-encoded maximum value
+  virtual std::string EncodedMax() = 0;
+
+  virtual ~RowGroupStatistics() {}
+
+  Type::type PhysicalType() const { return descr_->physical_type(); }
```
wesm: same naming thing here

src/parquet/column/statistics.h
```diff
@@ -0,0 +1,127 @@
...
+      int64_t distinct_count, MemoryAllocator* allocator = default_allocator());
+
+  bool HasMinMax() const override;
+  void Reset() override;
+  void Merge(const RowGroupStatistics& other) override;
+
+  void Update(const T* values, int64_t num_not_null, int64_t num_null);
+
+  const T& Min() const { return min_; }
+  const T& Max() const { return max_; }
```
wesm: names

src/parquet/column/writer.cc
```diff
@@ -99,13 +101,15 @@ void ColumnWriter::WriteNewPage() {
...
-      Encoding::RLE, values, Encoding::PLAIN);
+      Encoding::RLE, values, Encoding::PLAIN, page_statistics_);
   total_bytes_written_ += bytes_written;
 
   // Re-initialize the sinks as GetBuffer made them invalid.
   InitSinks();
   num_buffered_values_ = 0;
   num_buffered_encoded_values_ = 0;
+  chunk_statistics_->Merge(*page_statistics_);
+  page_statistics_->Reset();
```
wesm: I've leafed through parquet-mr and parquet-format; it suggests that page_statistics_ is allowed to be empty (as you have here) but not null, if I'm reading things right (there's many layers to the onion). Is there any reason why we might want to let the page statistics be nullptr?

wesm: Looking further down the review it seems that this will always be non-null, so maybe this is moot.

src/parquet/column/writer.h
```diff
@@ -164,11 +177,14 @@ inline void TypedColumnWriter<DType>::WriteBatch(int64_t num_values,
...
 
   if (num_rows_ > expected_rows_) {
-    throw ParquetException("More rows were written in the column chunk then expected");
+    throw ParquetException("More rows were written in the column chunk than expected");
   }
 
   WriteValues(values_to_write, values);
 
+  auto stats = static_cast<TypedRowGroupStatistics<DType>*>(page_statistics_.get());
+  stats->Update(values, values_to_write, num_values - values_to_write);
```
wesm: Does it make sense to make writing the statistics optional? Can add an option to WriterProperties. I looked at parquet-mr and I think it always tracks the statistics; I don't have a sense of whether this will materially impact serialization performance (for applications which may not care about the column statistics).

src/parquet/file/reader-internal.cc
```diff
@@ -171,16 +171,37 @@ std::unique_ptr<PageReader> SerializedRowGroup::GetColumnPageReader(int i) {
...
+    case Type::INT96:
+      return MakeColumnStats<DataType<Type::INT96>>(meta_data, descr);
+    case Type::DOUBLE:
+      return MakeColumnStats<DataType<Type::DOUBLE>>(meta_data, descr);
+    case Type::FLOAT:
+      return MakeColumnStats<DataType<Type::FLOAT>>(meta_data, descr);
+    case Type::BYTE_ARRAY:
+      return MakeColumnStats<DataType<Type::BYTE_ARRAY>>(meta_data, descr);
+    case Type::FIXED_LEN_BYTE_ARRAY:
+      return MakeColumnStats<DataType<Type::FIXED_LEN_BYTE_ARRAY>>(meta_data, descr);
```
wesm: Use the typedefs for these `DataType<T>` values? I agree it might be nicer to move this up into `TypedColumnReader` somehow and skip the switch statement, but not necessary at the moment

src/parquet/file/reader-internal.cc
```diff
@@ -171,16 +171,37 @@ std::unique_ptr<PageReader> SerializedRowGroup::GetColumnPageReader(int i) {
...
+  }
```
wesm: nit: add:

```c++
default: 
  break;
```

src/parquet/file/writer-internal.cc
```diff
@@ -93,7 +107,7 @@ int64_t SerializedPageWriter::WriteDataPage(int32_t num_rows, int32_t num_values
   data_page_header.__set_encoding(ToThrift(encoding));
   data_page_header.__set_definition_level_encoding(ToThrift(definition_level_encoding));
   data_page_header.__set_repetition_level_encoding(ToThrift(repetition_level_encoding));
-  // TODO(PARQUET-593) statistics
+  data_page_header.__set_statistics(ToThrift(page_statistics));
```
wesm: Is it possible to avoid the extra copy of min/max (there are cases where min/max can be quite large but perhaps that is too esoteric to really be concerned about)

src/parquet/column/statistics-test.cc
```diff
@@ -0,0 +1,158 @@
...
+  values_ptr_ = reinterpret_cast<bool*>(bool_buffer_.data());
+}
+
+using TestTypes = ::testing::Types<Int32Type, Int64Type, Int96Type, FloatType, DoubleType,
+    ByteArrayType, FLBAType, BooleanType>;
+
+TYPED_TEST_CASE(TestRowGroupStatistics, TestTypes);
+
+TYPED_TEST(TestRowGroupStatistics, MinMaxEncode) {
+  this->TestMinMaxEncode();
```
wesm: Can you add some tests that attempt to verify the correctness of the comparators? Especially ByteArray/FLBA

src/parquet/file/file-serialize-test.cc
```diff
@@ -96,6 +96,11 @@ class TestSerialize : public ::testing::Test {
         values_out.data(), &values_read);
     ASSERT_EQ(100, values_read);
     ASSERT_EQ(values, values_out);
+
+    auto stats = std::static_pointer_cast<Int64Statistics>(rg_reader->GetColumnStats(0));
+    ASSERT_EQ(0, stats->NullCount());
+    ASSERT_EQ(128, stats->Min());
+    ASSERT_EQ(128, stats->Max());
```
wesm: This is a fairly weak test case because there's no variety to the input data. Relates to the question around verifying the correctness of the comparators

wesm:
Great to make progress on this -- a few nits, but main concern is working to create more thorough test cases verifying correctness of reads and writes (since databases use these statistics for to know when to skip pages / row groups, etc.).

src/parquet/column/writer.h
lomereiter: It makes sense indeed, I ran the benchmarks, and the performance hit is sometimes significant.

xhochy: Can you post some results? I would really prefer to have them optional, at least for some columns.

src/parquet/file/writer-internal.cc
lomereiter: Possible through manually changing `__isset.statistics` to true and operating directly on the member variable. That said, compilers might be smart enough to do copy elision.

src/parquet/column/writer.h
lomereiter: [column-io-stats.txt](https://github.com/apache/parquet-cpp/files/340188/column-io-stats.txt)
[column-io-no-stats.txt](https://github.com/apache/parquet-cpp/files/340189/column-io-no-stats.txt)

wesm:
Thank you, I will review the updated patch here in short order.

lomereiter:
Thanks for your time. 
I haven't got around to write comparison tests yet, but here's a quick summary of changes:
* writing statistics is disabled by default (often it makes sense only for a few specific columns)
* writing and reading is covered with tests, for single- and multi-page row groups
* all column-level settings are grouped into a struct, as it's more convenient and doesn't require multiple unordered_map accesses

src/parquet/column/properties.h
```diff
@@ -85,7 +85,16 @@ static constexpr ParquetVersion::type DEFAULT_WRITER_VERSION =
...
+class PARQUET_EXPORT ColumnSettings {
+ public:
+  ColumnSettings(Encoding::type encoding = DEFAULT_ENCODING,
+      Compression::type codec = DEFAULT_COMPRESSION_TYPE, bool collect_statistics = false)
+      : encoding(encoding), codec(codec), collect_statistics(collect_statistics) {}
+
+  Encoding::type encoding;
+  Compression::type codec;
+  bool collect_statistics;
+};
```
wesm: If this is strictly POD, this should be a struct (note google style guidelines on member variable names for structs vs classes)

src/parquet/column/properties.h
```diff
@@ -161,26 +168,37 @@ class PARQUET_EXPORT WriterProperties {
       return this->compression(path->ToDotString(), codec);
     }
 
+    Builder* statistics(bool enable = true) {
+      default_column_settings_.collect_statistics = enable;
+      return this;
+    }
```
wesm: Can you make enable not an optional argument? Seeing `...->statistics()` or `statistics('foo.bar.baz')` might not make it clear what's going on

src/parquet/column/properties.h
```diff
@@ -161,26 +168,37 @@ class PARQUET_EXPORT WriterProperties {
...
+
+    Builder* statistics(const std::string& path, bool enable = true) {
+      column_settings_[path].collect_statistics = enable;
+      return this;
+    }
+
+    Builder* statistics(
+        const std::shared_ptr<schema::ColumnPath>& path, bool enable = true) {
+      return this->statistics(path->ToDotString(), enable);
+    }
```
wesm: ditty on these

src/parquet/column/properties.h
```diff
@@ -193,46 +211,48 @@ class PARQUET_EXPORT WriterProperties {
...
   }
 
   Compression::type compression(const std::shared_ptr<schema::ColumnPath>& path) const {
-    auto it = codecs_.find(path->ToDotString());
-    if (it != codecs_.end()) return it->second;
-    return default_codec_;
+    return column_settings(path).codec;
+  }
+
+  bool statistics(const std::shared_ptr<schema::ColumnPath>& path) const {
```
wesm: Better name for this method?

src/parquet/column/properties.h
wesm: Per below, it might help to rename the method `collect_statistics()`

src/parquet/column/statistics.h
```diff
@@ -0,0 +1,128 @@
...
+
+  std::string EncodedMin() override;
+  std::string EncodedMax() override;
+
+ private:
+  MemoryAllocator* allocator_;
+
+  bool has_min_max_ = false;
+  T min_;
+  T max_;
```
majetideepak: The original `RowGroupStatistics` class kept the `min_`, `max_` values as pointers pointing to the actual metadata object since these values can be arbitrarily large, especially for string data. The assumption we did not document here is that the lifetime of the `RowGroupStatistics` object values is  until the `RowGroupReader`. 
But we are `copying` the `min_`, `max_` values here. @wesm any thoughts ?

src/parquet/column/statistics-test.cc
```diff
@@ -0,0 +1,212 @@
...
+#include "parquet/schema/descriptor.h"
+#include "parquet/column/test-util.h"
+#include "parquet/column/test-specialization.h"
+#include "parquet/column/statistics.h"
+#include "parquet/column/reader.h"
+#include "parquet/column/writer.h"
+#include "parquet/file/reader.h"
+#include "parquet/file/writer.h"
+#include "parquet/util/input.h"
+#include "parquet/util/output.h"
```
wesm: Good to be in the habit of alphabetizing includes; though the whole codebase is in need of a sweep of include-what-you-use love

src/parquet/column/statistics-test.cc
```diff
@@ -0,0 +1,212 @@
...
+  void TestFullRoundtrip(int64_t num_values, int64_t null_count) {
+    auto node = PrimitiveNode::Make("column", Repetition::OPTIONAL, TestType::type_num,
+        LogicalType::NONE, FLBA_LENGTH);
+    auto gnode = std::static_pointer_cast<GroupNode>(
+        GroupNode::Make("schema", Repetition::REQUIRED, std::vector<NodePtr>({node})));
+
+    GenerateData(num_values);
+
+    // compute statistics for the whole batch
+    ColumnDescriptor descr(node, 1 /* max definition level */, 0);
```
wesm: This is a little bit unusual; it would be better to do `const int max_def_level = 1;` if you're concerned about readability

src/parquet/column/statistics-test.cc
```diff
@@ -0,0 +1,212 @@
...
+    total.Merge(statistics1);
+    total.Merge(statistics2);
+
+    ASSERT_EQ(num_null[0] + num_null[1], total.null_count());
+    ASSERT_EQ(values_.size() * 2 - num_null[0] - num_null[1], total.num_values());
+    ASSERT_EQ(total.min(), std::min(statistics1.min(), statistics2.min()));
+    ASSERT_EQ(total.max(), std::max(statistics1.max(), statistics2.max()));
+  }
+
+  using TypedStats = TypedRowGroupStatistics<TestType>;
```
wesm: Move this to the top of the fixture and use it consistently?

src/parquet/column/statistics-test.cc
```diff
@@ -0,0 +1,212 @@
...
+    auto buffer = sink->GetBuffer();
+    std::unique_ptr<RandomAccessSource> source(new BufferReader(buffer));
+    auto file_reader = ParquetFileReader::Open(std::move(source));
+    auto rg_reader = file_reader->RowGroup(0);
+    auto stats = std::static_pointer_cast<TypedStats>(rg_reader->GetColumnStats(0));
+    // check values after serialization + deserialization
+    ASSERT_EQ(null_count, stats->null_count());
+    ASSERT_EQ(num_values - null_count, stats->num_values());
+    ASSERT_EQ(expected_stats.min(), stats->min());
+    ASSERT_EQ(expected_stats.max(), stats->max());
```
wesm: Cool, as discussed on the PR the last thing that's missing here is correctness of the min and max stats

src/parquet/file/file-serialize-test.cc
```diff
@@ -96,6 +99,11 @@ class TestSerialize : public ::testing::Test {
         values_out.data(), &values_read);
     ASSERT_EQ(100, values_read);
     ASSERT_EQ(values, values_out);
+
+    auto stats = std::static_pointer_cast<Int64Statistics>(rg_reader->GetColumnStats(0));
+    ASSERT_EQ(0, stats->null_count());
+    ASSERT_EQ(128, stats->min());
+    ASSERT_EQ(128, stats->max());
```
wesm: Can you add a couple known extreme values to the test data?

src/parquet/column/statistics.h
wesm: Thanks for catching this. 

I think there's a serious, and separate issue here because `ByteArray` and `FixedLenByteArray` do not own their data buffers. Which makes the logic in https://github.com/apache/parquet-cpp/pull/129/files#diff-59138f238492befbb8ae8a07b838463fR142 problematic

For example, between data pages in a row group, the data referenced by the pointers in this objects may be freed. This doesn't happen in the test suite (the data generated is held in a `std::vector<uint8_t>` for the duration of the test). 

So what needs to happen:

- Calling Merge with ByteArray or FLBA must copy the bytes out of the minimum/maximum object into a newly allocated `Buffer`; this should be stored in the statistics object. This way the lifetime of the value's memory can be tied locally to the lifetime of the Statistics object

- When reading a file, the RowGroupStatistics may reference the deserialized Thrift metadata in a zero-copy fashion (or, if necessary, decode the value into new memory). 

@lomereiter does this make sense? 

Practically speaking to test this, the test case will need to generate multiple batches of random data, to trigger this behavior (presumably we would get UB right now)

wesm:
OK, outside of some nits, the two items to address are:

- Memory lifetime of the "pointer" types (ByteArray and FLBA)
- Correctness of comparators

src/parquet/column/statistics.h
lomereiter: Yes, thanks for pointing this subtlety out. I'll try to come up with a failing test.

src/parquet/column/statistics-test.cc
lomereiter: In fact, latest versions of clang-format do that automatically, but I'm currently sticking to the one used on Travis CI (3.7).

wesm:
@lomereiter pinging you about this, are you able to look at this in the near future? Thanks

lomereiter:
@wesm yes, I'm gonna have quite long vacation soon, no worries

wesm:
@lomereiter can you let us know a status update on this patch when you can? I don't think it's too far away from completeness from last review

lomereiter:
Hi @wasm, I've been procrastinating for a while but today I tackled the main issue with the memory lifetime. I'm not sure what's the policy regarding `util/buffer.h` so I used `std::vector<uint8_t>` for storage.

src/parquet/column/statistics-test.cc
andreweduffy: Make sure when you test ByteArrayType to include some unicode characters ;)

https://github.com/apache/parquet-mr/pull/362

wesm: that might be too ambitious for this patch, can we create a JIRA for it to track? C++11 now does have some built in library support for unicode (see the `<locale>` include) but it could end up being necessary to pull in ICU or some 3rd party C++ unicode library

src/parquet/column/statistics.h
```diff
@@ -104,14 +104,15 @@ class PARQUET_EXPORT TypedRowGroupStatistics : public RowGroupStatistics {
...
-
   bool has_min_max_ = false;
   T min_;
   T max_;
+  MemoryAllocator* allocator_;
 
   void PlainEncode(const T& src, std::string* dst);
   void PlainDecode(const std::string& src, T* dst);
+  void Copy(const T& src, T& dst, std::vector<uint8_t>& free_flag);
+  std::vector<uint8_t> min_buffer_, max_buffer_;
```
wesm: You can use the `OwnedMutableBuffer` here along with the associated `MemoryAllocator`

piyushnarang:
@lomereiter  / @wesm  - We might need to tweak stats computation for binary fields a bit based on the discussion in the PRs: https://github.com/apache/parquet-format/pull/42 and https://github.com/apache/parquet-mr/pull/362.

wesm:
Thanks for letting us know. What do you recommend we do right now?

piyushnarang:
@wesm, I think Andrew is fairly close to resolving comments on the parquet-format PR. Would be nice to ensure both the cpp & java versions are behaving in a compatible way.

wesm:
@lomereiter @piyushnarang looks like we should do an unsigned binary comparison here. @lomereiter it looks like comparators are the only thing left -- you should wait to rebase until PARQUET-573 is merged. thanks for patience!

wesm:
sorry looks like we still need to computed the signed max/min -- we can come back and add the unsigned statistics when the format version drops

https://github.com/apache/parquet-format/pull/42#discussion_r77061582

wesm:
@lomereiter let's proceed and compute only the signed max/min in this patch until the dust settles on parquet-format?

lomereiter:
@wesm ok, I'll rebase and add comparison tests this weekend so that it can be merged finally

lomereiter:
So far rebased only, will get back to tests and further cleanup next evening.

lomereiter:
Switched to using signed min/max for byte array types, added a few comparison tests. Some inefficiencies in read/write paths remain, but are topics for separate JIRAs.

wesm:
Awesome, will review when I can.

src/parquet/column/page.h
```diff
@@ -77,31 +80,24 @@ class DataPage : public Page {
...
   int32_t num_values_;
   Encoding::type encoding_;
   Encoding::type definition_level_encoding_;
   Encoding::type repetition_level_encoding_;
-
-  // So max/min can be populated privately
-  friend class SerializedPageReader;
-  std::string max_;
-  std::string min_;
+  EncodedStatistics statistics_;
```
xhochy: To avoid copying a lot of data, this would probably be better as a `shared_ptr`.

src/parquet/column/properties.h
```diff
@@ -82,28 +82,41 @@ static constexpr int64_t DEFAULT_PAGE_SIZE = 1024 * 1024;
 static constexpr bool DEFAULT_IS_DICTIONARY_ENABLED = true;
 static constexpr int64_t DEFAULT_DICTIONARY_PAGE_SIZE_LIMIT = DEFAULT_PAGE_SIZE;
 static constexpr int64_t DEFAULT_WRITE_BATCH_SIZE = 1024;
+static constexpr bool DEFAULT_COLLECT_STATISTICS = false;
```
xhochy: Are they enabled as default in `parquet-mr`? We should probably stick to the same defaults here. Collecting statistics costs a bit but will improve query times on the files a lot.

src/parquet/column/properties.h
```diff
@@ -82,28 +82,41 @@ static constexpr int64_t DEFAULT_PAGE_SIZE = 1024 * 1024;
...
 static constexpr Encoding::type DEFAULT_ENCODING = Encoding::PLAIN;
 static constexpr ParquetVersion::type DEFAULT_WRITER_VERSION =
     ParquetVersion::PARQUET_1_0;
 static std::string DEFAULT_CREATED_BY = "Apache parquet-cpp";
 static constexpr Compression::type DEFAULT_COMPRESSION_TYPE = Compression::UNCOMPRESSED;
 
-using ColumnCodecs = std::unordered_map<std::string, Compression::type>;
+class PARQUET_EXPORT ColumnSettings {
```
xhochy: I'm ok with this class (although it should be named `ColumnProperties`) but not with the outcome at the end. Currently it is possible to select a custom encoding for a column but fallback to the file-global defaults for all other settings. With this new class, once we have set one property on a column-basis, for this column we have to either set all other file-global defaults explicitly on this column or we'll get the parquet-cpp defaults.

src/parquet/column/statistics.cc
```diff
@@ -0,0 +1,217 @@
...
+#include "parquet/encodings/plain-encoding.h"
+#include "parquet/util/buffer.h"
+#include "parquet/util/comparison.h"
+#include "parquet/util/output.h"
+#include "parquet/exception.h"
+
+namespace parquet {
+
+template <typename TypedStats>
+std::shared_ptr<TypedStats> RowGroupStatistics::as() {
```
xhochy: I would prefer if this was done explicitly by the user as with all other casts.

src/parquet/column/statistics.cc
```diff
@@ -0,0 +1,217 @@
...
+template <typename DType>
+void TypedRowGroupStatistics<DType>::Copy(const T& src, T* dst, OwnedMutableBuffer&) {
+  *dst = src;
+}
+
+template <>
+void TypedRowGroupStatistics<FLBAType>::Copy(
+    const FLBA& src, FLBA* dst, OwnedMutableBuffer& buffer) {
+  if (dst->ptr == src.ptr) return;
+  auto len = descr_->type_length();
```
xhochy: Please use the explicity type here.

src/parquet/column/statistics.cc
```diff
@@ -0,0 +1,217 @@
...
+  buffer.Resize(len);
+  for (int i = 0; i < len; i++)
+    buffer[i] = src.ptr[i];
```
xhochy: Could we use `memcpy` here?

src/parquet/column/statistics.cc
```diff
@@ -0,0 +1,217 @@
...
+  *dst = FLBA(buffer.data());
+}
+
+template <>
+void TypedRowGroupStatistics<ByteArrayType>::Copy(
+    const ByteArray& src, ByteArray* dst, OwnedMutableBuffer& buffer) {
+  if (dst->ptr == src.ptr) return;
+  buffer.Resize(src.len);
+  for (uint32_t i = 0; i < src.len; i++)
```
xhochy: `memcpy`?

src/parquet/column/statistics.h
```diff
@@ -0,0 +1,180 @@
...
+  TypedRowGroupStatistics(const T& min, const T& max, int64_t num_values,
+      int64_t null_count, int64_t distinct_count);
+
+  TypedRowGroupStatistics(const ColumnDescriptor* schema, const std::string& encoded_min,
+      const std::string& encoded_max, int64_t num_values, int64_t null_count,
+      int64_t distinct_count, MemoryAllocator* allocator = default_allocator());
+
+  bool HasMinMax() const override;
+  void Reset() override;
+  void Merge(const RowGroupStatistics& other) override;
```
xhochy: This function should only take `TypedRowGroupStatistics<DType>` then you could get rid of the physical type check and check at compile time.

src/parquet/column/statistics.h
```diff
@@ -0,0 +1,180 @@
...
+
+  void Update(const T* values, int64_t num_not_null, int64_t num_null);
+
+  const T& min() const { return min_; }
+  const T& max() const { return max_; }
+
+  std::string EncodedMin() override;
```
xhochy: Return const references

src/parquet/column/writer.h
```diff
@@ -130,6 +133,18 @@ class PARQUET_EXPORT ColumnWriter {
 
   std::vector<CompressedDataPage> data_pages_;
 
+  // Initialized in subclass constructors
+  std::shared_ptr<RowGroupStatistics> page_statistics_;
+  std::shared_ptr<RowGroupStatistics> chunk_statistics_;
+
+  template <typename Type>
+  void InitStatistics() {
```
xhochy: Please make this part of the `TypedColumnWriter`

src/parquet/column/writer.cc
```diff
@@ -172,25 +183,31 @@ void ColumnWriter::FlushBufferedDataPages() {
...
       current_encoder_.reset(
-          new DictEncoder<Type>(descr, &pool_, properties->allocator()));
+          new DictEncoder<Type>(descr_, &pool_, properties->allocator()));
       break;
     default:
       ParquetException::NYI("Selected encoding is not supported");
   }
+
+  if (properties->collect_statistics(descr_->path())) {
+    InitStatistics<Type>();
```
xhochy: No need for `<Type>` if this function is part of `TypedColumnWriter`

src/parquet/file/metadata.cc
```diff
@@ -23,21 +23,49 @@
...
+    case Type::DOUBLE:
+      return MakeColumnStatsT<DoubleType>(meta_data, descr);
+    case Type::FLOAT:
+      return MakeColumnStatsT<FloatType>(meta_data, descr);
+    case Type::BYTE_ARRAY:
+      return MakeColumnStatsT<ByteArrayType>(meta_data, descr);
+    case Type::FIXED_LEN_BYTE_ARRAY:
+      return MakeColumnStatsT<FLBAType>(meta_data, descr);
+  }
+  return nullptr;
```
xhochy: Raise an exception if the type is not supported.

src/parquet/file/metadata.cc
```diff
@@ -23,21 +23,49 @@
 
 namespace parquet {
 
+template <typename DType>
+static std::shared_ptr<RowGroupStatistics> MakeColumnStatsT(
```
xhochy: `MakeColumnStatsTyped`

src/parquet/file/reader.cc
```diff
@@ -160,16 +160,16 @@ void ParquetFileReader::DebugPrint(
     // Print column metadata
     for (auto i : selected_columns) {
       auto column_chunk = group_metadata->ColumnChunk(i);
-      const ColumnStatistics stats = column_chunk->statistics();
+      const auto stats = column_chunk->statistics();
```
xhochy: Please write the type here.

xhochy: @wesm Do you have a link to the guidelines at hand?

src/parquet/column/properties.h
lomereiter: I don't understand this objection since WriterProperties::Builder interface is unchanged.

xhochy: The interface is the same, just the behaviour changed.

src/parquet/column/statistics.h
lomereiter: Thanks, moving things into TypedColumnWriter is much cleaner indeed, will do.

src/parquet/column/properties.h
lomereiter: Gotcha. AFAIU all setters should first set `column_settings_[path]` to `default_column_settings_`.

xhochy: This will still not be enough. You will also need an `isset` for each entry. I would suggest to still use the numerous maps in the builder and build the column_settings_ instances at the end.

src/parquet/file/reader.cc
wesm: https://google.github.io/styleguide/cppguide.html#auto

aside: we should survey our usages of auto to make sure we are using `const auto&` in the right places

src/parquet/column/properties.h
xhochy: Hope I have covered my issue in an understandable unit test: https://github.com/apache/parquet-cpp/pull/166

src/parquet/column/properties.h
lomereiter: I checked and they are indeed enabled by default. However, if values of min/max are large (> 4KB) they are not written (https://issues.apache.org/jira/browse/PARQUET-372).

src/parquet/column/properties.h
lomereiter: Ok, went with your suggestion, thanks for putting in the test.

src/parquet/column/statistics.h
lomereiter: Renamed to `EncodeMin`/`EncodeMax` to reflect better that these values are not stored anywhere.

wesm:
I think this is about ready to go. Needs a rebase. @xhochy ?

lomereiter:
Thanks all, I rebased the branch.

wesm:
+1

PR #131, opened 2016-06-30
PARQUET-489: Shared library symbol visibility

wesm:
This was a bit of rabbit hole, because on Linux you must instruct gcc how to hide symbols from your static thirdparty libraries (see `src/parquet/symbols.map`). I learned about this from Apache Kudu (incubating) client cmake files. 

Probably still a bit more to do here (and I may have missed some things we should export, and exporting things that should not be exported); I will make sure the Arrow-Parquet tests still pass, but wanted to get some early feedback.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -320,6 +320,7 @@ string (TOUPPER ${CMAKE_BUILD_TYPE} CMAKE_BUILD_TYPE)
 
 if ("${CMAKE_BUILD_TYPE}" STREQUAL "DEBUG")
   set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} ${CXX_FLAGS_DEBUG}")
+
 elseif ("${CMAKE_BUILD_TYPE}" STREQUAL "FASTDEBUG")
   set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} ${CXX_FLAGS_FASTDEBUG}")
 elseif ("${CMAKE_BUILD_TYPE}" STREQUAL "RELEASE")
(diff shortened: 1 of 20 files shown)


wesm:
Here is what the export table looks like now: https://gist.github.com/wesm/fa21c50c06ccf2288bc626b4c122e63e

CMakeLists.txt
```diff
@@ -508,9 +519,11 @@ if (PARQUET_BUILD_SHARED)
     set_target_properties(parquet_shared
       PROPERTIES
       LIBRARY_OUTPUT_DIRECTORY "${BUILD_OUTPUT_ROOT_DIRECTORY}"
+      LINK_FLAGS "${SHARED_LINK_FLAGS}"
       OUTPUT_NAME "parquet")
-    target_link_libraries(parquet_shared LINK_PUBLIC ${LIBPARQUET_LINK_LIBS})
-    target_link_libraries(parquet_shared LINK_PRIVATE ${LIBPARQUET_PRIVATE_LINK_LIBS})
+    target_link_libraries(parquet_shared
+      ${LIBPARQUET_LINK_LIBS}
```
xhochy: A `LINK_PUBLIC` seems to be missing here.

wesm: It's redundant, but I'll add it to be explicit

xhochy: Older CMake versions require it if you also use LINK_PRIVATE on the target.

wesm: ah, good to know, thx

wesm:
@xhochy the Arrow test suite passes now, can you let me know if this is OK?

wesm:
cc also @lomereiter @majetideepak, let me know if this impacts you in some way

lomereiter:
+1, LGTM.

xhochy:
+1, LGTM

majetideepak:
+1

wesm:
Thanks all. +1

PR #132, opened 2016-07-05
PARQUET-551:Handle compiler warnings due to disabled DCHECKs in relea…

majetideepak:
…se builds

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -331,7 +331,7 @@ endif ()
 
 message(STATUS "Build Type: ${CMAKE_BUILD_TYPE}")
 
-SET(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -std=c++11 -Wall")
+SET(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -std=c++11 -Wall -Wno-strict-aliasing")
 
 if (PARQUET_USE_SSE)
   SET(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -march=native")
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -331,7 +331,7 @@ endif ()
 
 message(STATUS "Build Type: ${CMAKE_BUILD_TYPE}")
 
-SET(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -std=c++11 -Wall -Wno-strict-aliasing")
+SET(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -std=c++11 -Wall -fno-strict-aliasing")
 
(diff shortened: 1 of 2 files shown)


CMakeLists.txt
```diff
@@ -331,7 +331,7 @@ endif ()
 
 message(STATUS "Build Type: ${CMAKE_BUILD_TYPE}")
 
-SET(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -std=c++11 -Wall")
+SET(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -std=c++11 -Wall -Wno-strict-aliasing")
```
majetideepak: We are breaking strict aliasing at 
reader-internal.cc:282: `uint32_t metadata_len = *reinterpret_cast<uint32_t*>(footer_buffer);`

xhochy:
But this now essentially means that the code passed as `condition` to DCHECK will be executed? I'm not sure that this is always wanted, e.g. if there are some costly checks.

wesm:
I believe the compiler optimizes the expression away since the result is unused.

majetideepak:
Yeah! The compiler will optimize the expression since the result is unused.

CMakeLists.txt
wesm: Looks like we should be compiling with `-fno-strict-aliasing`; I think this is considered a best practice in general to avoid odd edge cases, even though it may prevent certain compiler optimizations

src/parquet/util/logging.h
```diff
@@ -44,24 +44,31 @@ namespace parquet {
...
 #define DCHECK_LT(val1, val2) \
+  ((void)(val1));             \
   while (false)               \
   parquet::internal::NullLog()
 #define DCHECK_GE(val1, val2) \
+  ((void)(val1));             \
   while (false)               \
   parquet::internal::NullLog()
 #define DCHECK_GT(val1, val2) \
+  ((void)(val1));             \
```
wesm: Maybe put this in a `PARQUET_IGNORE_EXPR` macro or something to make it more clear?

CMakeLists.txt
majetideepak: I agree. We can use the `__restrict` attribute if needed.

xhochy:
Only thing I may be worried about is if the compiler manages to optimize the calls to `CheckIntegrity` away. Otherwise change looks fine.

majetideepak:
This is a good point. I just debugged the `mem-pool-test` in release build and debug build with a break point inside `CheckIntegrity`. The debug build hit the break point and the release build did not.
So I guess it works for calls to `CheckIntegrity` as well.

xhochy:
Nice, then a clear +1 from my side.

wesm:
+1, thank you

PR #133, opened 2016-07-06
PARQUET-653: Build conda artifacts with -static-libstdc++ again

wesm:
The final stage of this work will be to move dev/testing packaging builds to centos6 images on Circle CI (this may be possible in Travis -- which allows you to use docker -- but easier with Circle CI)


xhochy:
+1, LGTM

wesm:
+1, thanks

wesm:
So, this unfortunately has broken the Arrow builds. I think we're going to have to set up a separate repo with a Dockerfile that uses the RHEL devtoolset to build these artifacts. static linking libstdc++ is far too finicky it turns out

wesm:
We could also just use conda forge and keep bumping the build number when we're ready to upgrade the artifact

PR #134, opened 2016-07-09
PARQUET-656: Revert attempts to make conda artifacts portable

wesm:
The toolchain issue is not tractable on an ad hoc basis. I suggest we use a consistent gcc toolchain across the CI environments for Parquet and Arrow, and we can post dev builds for human consumption / testing via conda-forge (where the RHEL devtoolset is there and already working well). Sound good @xhochy?

see also https://github.com/conda-forge/staged-recipes/pull/992

### DIFF
--- a/conda.recipe/build.sh
+++ b/conda.recipe/build.sh
@@ -3,20 +3,18 @@
 set -e
 set -x
 
-# FIXME: This is a hack to make sure the environment is activated.
-# The reason this is required is due to the conda-build issue
-# mentioned below.
-#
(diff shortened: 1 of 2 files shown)


xhochy:
Sounds fine.

One of the things I really like about the manylinux1 PEP is that it provides a clear spec of a common base system and also a tool to check that you actually adhere to this system.

wesm:
Agreed. FWIW, I'm definitely supportive of having manylinux1 wheels as a primary means of obtaining a working Python stack when things are more "production ready"

wesm:
+1

PR #135, opened 2016-07-09
PARQUET-657: Do not define DISALLOW_COPY_AND_ASSIGN if already defined

wesm:
This can cause conflicts with other libraries that also use this common macro (from Google's gutil library originally)


xhochy:
+1, LGTM

PR #136, opened 2016-07-10
PARQUET-658: Add virtual destructor to ColumnReader

xhochy:


wesm:
Thanks for catching this, +1. Did you run into UB when a pointer to the base was being deleted?

xhochy:
Actually discovered this while chasing for PARQUET-659

wesm:
Oh good, well good timing then. I'll merge all these things and hopefully we can merge ARROW-233

PR #137, opened 2016-07-10
PARQUET-659: Export extern templates for typed column reader/writer classes

wesm:
While we've already force-instantiated these template classes, they were not visible across the board. gcc on Linux has no problem with it, but found that the Arrow OS X builds were core dumping (with a misleading error message) because of it. The test suites pass locally for me now.


xhochy:
+1, LGTM

wesm:
+1

wesm:
the merged build errored due to an idiosyncratic homebrew problem

PR #138, opened 2016-07-11
PARQUET-661: Fix perl hashbang to use perl in environment

wesm:
As noted in https://github.com/conda-forge/staged-recipes/pull/992, if `perl` is in the path but in a non-standard location, then `ctest` fails.


xhochy:
+1, LGTM

wesm:
+1

PR #139, opened 2016-07-11
PARQUET-662: Compile ParquetException implementation and explicitly export

wesm:
This was causing downstream failure in ARROW-237. Now thirdparty libraries can safely throw `ParquetException` and be properly recognized on both ends.


wesm:
I verified locally that this resolves the downstream failure on OS X.

wesm:
I added a simple option to `ADD_PARQUET_TEST` so we can write unit tests against the shared library (defaults to the static library).

nongli:
+1

wesm:
Thanks. I cleaned up the cmake function a bit and will merge on green build.

majetideepak:
+1 LGTM

PR #140, opened 2016-07-29
PARQUET-671: performance improvements for rle/bit-packed decoding

edani:
Testing on my own data shows an order-of-magnitude improvement.

I separated the commits for clarity, each one gives an imcremental improvement.

The motivation for the last commit (allowing NULL for def_levels/rep_level) is a workaround for Spark  which doesn't seem to be able to generate columns without def_level, even when a column is specified as "not nullable".


wesm:
This is great, thank you! I will review as soon as I can — optimizing the bit decoding was high on my list of performance improvements. 

Can you open a JIRA on issues.apache.org for this and prefix your PR title with "PARQUET-XXX:" (where XXX is the JIRA number?)

src/parquet/column/reader.h
```diff
@@ -184,7 +188,7 @@ inline int64_t TypedColumnReader<DType>::ReadBatch(int batch_size, int16_t* def_
   }
 
   // Not present for non-repeated fields
-  if (descr_->max_repetition_level() > 0) {
+  if (descr_->max_repetition_level() > 0 && rep_levels) {
     num_rep_levels = ReadRepetitionLevels(batch_size, rep_levels);
     if (num_def_levels != num_rep_levels) {
       throw ParquetException("Number of decoded rep / def levels did not match");
```
majetideepak: This will throw an exception if we set only  `def_levels` to NULL. Either both should be NULL or none.

src/parquet/util/rle-encoding.h
```diff
@@ -26,6 +26,7 @@
 #include "parquet/util/compiler-util.h"
```
majetideepak: I am using this space for comments on "src/parquet/util/bpacking.h". Some of the unpack*_32 methods can be put in a loop. Putting them in a loop allows us to use vectorization pragmas as well.

src/parquet/encodings/dictionary-encoding.h
```diff
@@ -64,22 +64,17 @@ class DictionaryDecoder : public Decoder<Type> {
 
   virtual int Decode(T* buffer, int max_values) {
     max_values = std::min(max_values, num_values_);
-    for (int i = 0; i < max_values; ++i) {
-      buffer[i] = dictionary_[index()];
+    int decoded_values = idx_decoder_.GetBatchWithDict(buffer, max_values, dictionary_);
+    if (decoded_values != max_values) {
+      ParquetException::EofException();
     }
```
majetideepak: Can we remove the exception here and subtract, return `decoded_values` below ?
The semantics of `Decode` like methods should not worry about the mismatch in the values. The layers using these methods should throw an exception if required.

src/parquet/encodings/plain-encoding.h
```diff
@@ -142,10 +142,8 @@ class PlainDecoder<BooleanType> : public Decoder<BooleanType> {
 
   virtual int Decode(bool* buffer, int max_values) {
     max_values = std::min(max_values, num_values_);
-    bool val;
-    for (int i = 0; i < max_values; ++i) {
-      if (!bit_reader_.GetValue(1, &val)) { ParquetException::EofException(); }
-      buffer[i] = val;
+    if (bit_reader_.GetBatch(1, buffer, max_values) != max_values) {
+      ParquetException::EofException();
```
majetideepak: Same as above

src/parquet/util/rle-encoding.h
wesm: Since this is thirdparty code benchmarking / perf enhancements would be best tackled in a separate patch -- can you open a JIRA about this?

src/parquet/column/reader.h
wesm: I believe this would only occur if there a bug in parquet-cpp -- at most a DCHECK is needed upon initializing the column reader

src/parquet/encodings/dictionary-encoding.h
wesm: Below (in `index()`, which was removed), `num_values_` was being decremented with each requested dictionary index, so this only preserves the existing logic

src/parquet/encodings/plain-encoding.h
wesm: This is also preserving the existing logic -- removing it seems out of scope for the patch unless test coverage was specifically added to explore this code path

src/parquet/util/bit-stream-utils.inline.h
```diff
@@ -86,34 +89,106 @@ inline bool BitWriter::PutVlqInt(uint32_t v) {
 }
 
 template <typename T>
+inline void GetValue_(
+  int num_bits, T* v, const uint8_t* buffer, int& bit_offset,
+  int& byte_offset, uint64_t& buffered_values, int max_bytes) {
```
wesm: The indentation here is a bit off, can you run `make format` if you have LLVM / clang-tools installed?

wesm: The Google C++ guide counterindicates mutable reference arguments: https://google.github.io/styleguide/cppguide.html#Reference_Arguments. Can you pass the mutable arguments as pointers (`uint64_t* buffered_values` would be more clear, anyway). You'll have to set the out values of `bit_offset` and `byte_offset` explicitly, but that is better than implicit mutation

src/parquet/util/rle-encoding.h
```diff
@@ -114,6 +115,10 @@ class RleDecoder {
   template <typename T>
   int GetBatch(T* values, int batch_size);
 
+  /// Like GetBatch but the values are then decoded using the provided dictionary
+  template <typename T>
+  int GetBatchWithDict(T* values, int batch_size, Vector<T>& dictionary);
```
wesm: `const Vector<T>&`?

wesm: It may be better to put the dictionary as the first argument (https://google.github.io/styleguide/cppguide.html#Function_Parameter_Ordering)

wesm:
Left some minor comments, and running clang-format would be good, but otherwise this is great. Any chance you would want to add a gbenchmark case? Doesn't need to be in this patch if you want to create a separate JIRA as a reminder to add a benchmark

src/parquet/column/reader.h
majetideepak: If we invoke `ReadBatch(batch_size, NULL, rep_levels, values, &values_read)`, this modification will throw mismatch exception correct ?

src/parquet/encodings/dictionary-encoding.h
majetideepak: Makes sense.

src/parquet/column/reader.h
wesm: Yes, but this would be a misuse of the API

src/parquet/util/rle-encoding.h
majetideepak: Will do.

src/parquet/column/reader.h
wesm: We could certainly insert a DCHECK that verifies that if `rep_levels` is not nullptr then neither is `def_levels`. 

@edani could you explain how this issue manifested itself? how I'm reading what you said about Spark is that there are always definition levels, even when `max_definition_level` is 0. parquet-cpp can just skip decoding the data in this case.

src/parquet/util/rle-encoding.h
edani: Note, I doubt that improving bpacking.h is the quickest/most efficient course of action. @lemire has done more work on bit-packing code, and a SIMD version can be found at https://github.com/lemire/simdcomp

wesm: I'm willing to trust him on this one given that this is one of his areas of research

lemire: We are available to help, if help is needed. Indeed, https://github.com/lemire/simdcomp is solid C library that can deliver multiple time the performance of a scalar implementation by making explicit use of SIMD instructions. It is possible to achieve good gains over a scalar Parquet decoder using deliberate SIMD code. I am personally available to look at this if there is interest...

wesm: We are happy to accept the help. Currently we are not taking care even to allocate aligned / padded memory for SIMD

lemire: @wesm  @edani  SSE/AVX does not require aligned access... nor is it likely to be a performance concern. Can you guys email me? (lemire@gmail.com) Though I am interested, I have not been following this project. I'd be interested in having a **quick** email exchange with people who are knowledgeable in ``parquet-cpp``. Then we can probably put something together...

edani:
Thanks for the review!  I pushed a commit to fix the style. I'm working on my dev environment so I can run clang-format and look into benchmarks.

src/parquet/column/reader.h
edani: My problem is is that I couldn't get Spark to write columns with no definition levels. Instead I get columns with `max_definition_level=1`, and `def_devels` set to all ones. And the JSON schema says `"nullable":true`, which is the opposite of what I asked.

The fix should really be made against Spark.  But it was trivial to apply a workaround to parquet-cpp.

edani:
I think I addressed every comment. I can squash commits once everyone is OK with the current state.

majetideepak:
+1

wesm:
No need to squash; we have a patch tool for that

src/parquet/column/reader.h
```diff
@@ -184,9 +188,9 @@ inline int64_t TypedColumnReader<DType>::ReadBatch(int batch_size, int16_t* def_
   }
 
   // Not present for non-repeated fields
-  if (descr_->max_repetition_level() > 0) {
+  if (descr_->max_repetition_level() > 0 && rep_levels) {
     num_rep_levels = ReadRepetitionLevels(batch_size, rep_levels);
-    if (num_def_levels != num_rep_levels) {
+    if (def_levels && num_def_levels != num_rep_levels) {
```
wesm: This is still a little bit pedantic. If `rep_levels` is not null and `def_levels` is null, then a DCHECK should fail. Performing this check in release builds doesn't make total sense to me (indicates an improper use of the API).

wesm:
+1, merging this despite the small nit

src/parquet/column/reader.h
wesm: I'm still not totally grasping this. Basically it sounds like Spark does not let users write non-nullable columns. So with the data being the data, the correct answer is to always pass a def levels buffer, even if it's all 1's. Ordinarily of `max_definition_level() > 0` then def_levels should never be done. It sounds like you are passing nullptr as a performance optimizing, is that right?

src/parquet/column/reader.h
majetideepak: The comment says `Set def_levels or rep_levels to nullptr if you want to skip reading them.` If the API user just wants to read `rep_levels` and not the `def_levels`, then this additional check is required. Alternatively, I feel we should throw a meaningful error message for improper use of API as well.

majetideepak: If we decide that the API user should either read both `def_levels`, `rep_levels` or none, then I agree, a DCHECK indicating an API violation should be used.

wesm: The correct behavior would be to DCHECK that `def_levels` is not null if max_definition_level is > 0 and likewise for max_repetition_level. The only case where the first check will pass and the 2nd does not if if the user calls the API incorrectly. We can add a DCHECK for that, too, if we want (ref_levels passed but not def_levels -- bad API usage)

majetideepak: I believe edani's purpose of the `def_levels`/ `rep_levels` check in the `if condition` is to provide users a way to skip reading these values altogether. A user can choose not to read these values inspite of `max_definition_level > 0` and `max_repetition_level > 0` by passing NULL.

src/parquet/column/reader.h
edani: Yes, the only purpose is performance. Decoding def_levels has about the same cost as decoding values, so the difference is visible.

wesm: @edani in this case you need to know for sure that your data does not have any null values. An arbitrary Parquet reader can't assume this, of course, so I'm find leaving in the "short circuit" to save the computational work as long as it's well documented.

src/parquet/util/rle-encoding.h
julienledem: @lemire @wesm @edani did this exchange happen? could someone send a summary to dev@parquet.apache.org ? if not I can start a thread with @nongli as well

lemire: I'll be happy to participate in a thread on this issue, but yes, we did exchange emails.

I want to (soon) build a prototype to show what I think is possible... I'd be happy to do so publicly.

julienledem: @lemire sounds great! Looking forward to it.

majetideepak: @lemire This is indeed very useful and can lead to techniques to use not just SSE/AVX, but heterogeneous hardware as well. We will be happy to assist you in building your prototype too.

lemire: Following a brief but instructive email exchange with @edani, I have come up with this prototype to illustrate what I think is possible... https://github.com/lemire/dictionary (all the mistakes are my own)

lemire: https://issues.apache.org/jira/browse/PARQUET-688

PR #141, opened 2016-08-03
PARQUET-672: Build dev binary artifacts in debug mode

wesm:
I will move these builds to conda-forge as soon as possible so we aren't overly reliant on the Travis master builds for development of Apache Arrow


wesm:
+1

PR #142, opened 2016-08-06
PARQUET-666: Add support for writing dictionaries

xhochy:
Working version but still some minor stuff to clean up (but I'd be happy for comments).

Had to change the Encoder interface to make it work with dictionary encoding. Also dictionary encoded RowGroups are currently limited to a single DataPage.

### DIFF
--- a/src/parquet/column/properties.h
+++ b/src/parquet/column/properties.h
@@ -77,9 +77,8 @@ class PARQUET_EXPORT ReaderProperties {
 
 ReaderProperties PARQUET_EXPORT default_reader_properties();
 
-static int64_t DEFAULT_PAGE_SIZE = 1024 * 1024;
+static int64_t DEFAULT_PAGE_SIZE = 64 * 1024 * 1024;
 static int64_t DEFAULT_DICTIONARY_PAGE_SIZE = DEFAULT_PAGE_SIZE;
-static bool DEFAULT_IS_DICTIONARY_ENABLED = true;
 static Encoding::type DEFAULT_ENCODING = Encoding::PLAIN;
 static constexpr ParquetVersion::type DEFAULT_WRITER_VERSION =
     ParquetVersion::PARQUET_1_0;
@@ -93,7 +92,6 @@ class PARQUET_EXPORT WriterProperties {
    public:
     Builder()
         : allocator_(default_allocator()),
-          dictionary_enabled_(DEFAULT_IS_DICTIONARY_ENABLED),
           dictionary_pagesize_(DEFAULT_DICTIONARY_PAGE_SIZE),
           default_encoding_(DEFAULT_ENCODING),
           pagesize_(DEFAULT_PAGE_SIZE),
@@ -113,16 +111,6 @@ class PARQUET_EXPORT WriterProperties {
 
     Builder* data_pagesize(int64_t pg_size) {
       pagesize_ = pg_size;
-      return this;
-    }
-
-    Builder* enable_dictionary() {
-      dictionary_enabled_ = true;
-      return this;
-    }
-
-    Builder* disable_dictionary() {
-      dictionary_enabled_ = false;
       return this;
     }
 
@@ -162,14 +150,13 @@ class PARQUET_EXPORT WriterProperties {
     }
 
     std::shared_ptr<WriterProperties> build() {
-      return std::shared_ptr<WriterProperties>(new WriterProperties(allocator_,
-          dictionary_enabled_, dictionary_pagesize_, default_encoding_, encodings_,
-          pagesize_, version_, default_codec_, codecs_));
+      return std::shared_ptr<WriterProperties>(
+          new WriterProperties(allocator_, dictionary_pagesize_, default_encoding_,
+              encodings_, pagesize_, version_, default_codec_, codecs_));
     }
 
    private:
     MemoryAllocator* allocator_;
-    bool dictionary_enabled_;
     int64_t dictionary_pagesize_;
     // Encoding used for each column if not a specialized one is defined as
     // part of encodings_
@@ -185,8 +172,6 @@ class PARQUET_EXPORT WriterProperties {
 
   MemoryAllocator* allocator() const { return allocator_; }
 
-  bool dictionary_enabled() const { return dictionary_enabled_; }
-
   int64_t dictionary_pagesize() const { return dictionary_pagesize_; }
 
   int64_t data_pagesize() const { return pagesize_; }
@@ -194,9 +179,19 @@ class PARQUET_EXPORT WriterProperties {
   ParquetVersion::type version() const { return parquet_version_; }
 
   Encoding::type encoding(const std::shared_ptr<schema::ColumnPath>& path) const {
+    Encoding::type coding = default_encoding_;
     auto it = encodings_.find(path->ToDotString());
-    if (it != encodings_.end()) { return it->second; }
-    return default_encoding_;
+    if (it != encodings_.end()) { coding = it->second; }
+
+    // Use the correct enum value for dictionary coding based on the used Parquet version
+    if (coding == Encoding::PLAIN_DICTIONARY || coding == Encoding::RLE_DICTIONARY) {
+      if (parquet_version_ == ParquetVersion::PARQUET_1_0) {
+        return Encoding::PLAIN_DICTIONARY;
+      } else {
+        return Encoding::RLE_DICTIONARY;
+      }
+    }
+    return coding;
   }
 
   Compression::type compression(const std::shared_ptr<schema::ColumnPath>& path) const {
@@ -206,26 +201,21 @@ class PARQUET_EXPORT WriterProperties {
   }
 
  private:
-  explicit WriterProperties(MemoryAllocator* allocator, bool dictionary_enabled,
-      int64_t dictionary_pagesize, Encoding::type default_encoding,
+  explicit WriterProperties(MemoryAllocator* allocator, int64_t dictionary_pagesize,
+      Encoding::type default_encoding,
       const std::unordered_map<std::string, Encoding::type>& encodings, int64_t pagesize,
       ParquetVersion::type version, Compression::type default_codec,
       const ColumnCodecs& codecs)
       : allocator_(allocator),
-        dictionary_enabled_(dictionary_enabled),
         dictionary_pagesize_(dictionary_pagesize),
         default_encoding_(default_encoding),
         encodings_(encodings),
         pagesize_(pagesize),
         parquet_version_(version),
         default_codec_(default_codec),
-        codecs_(codecs) {
-    pagesize_ = DEFAULT_PAGE_SIZE;
-    dictionary_enabled_ = DEFAULT_IS_DICTIONARY_ENABLED;
-  }
+        codecs_(codecs) {}
 
   MemoryAllocator* allocator_;
-  bool dictionary_enabled_;
   int64_t dictionary_pagesize_;
   Encoding::type default_encoding_;
   std::unordered_map<std::string, Encoding::type> encodings_;
--- a/src/parquet/column/writer.cc
+++ b/src/parquet/column/writer.cc
@@ -18,6 +18,7 @@
 #include "parquet/column/writer.h"
 
 #include "parquet/column/properties.h"
+#include "parquet/encodings/dictionary-encoding.h"
 #include "parquet/encodings/plain-encoding.h"
 
 namespace parquet {
@@ -32,11 +33,14 @@ std::shared_ptr<WriterProperties> default_writer_properties() {
 }
 
 ColumnWriter::ColumnWriter(const ColumnDescriptor* descr,
-    std::unique_ptr<PageWriter> pager, int64_t expected_rows, MemoryAllocator* allocator)
+    std::unique_ptr<PageWriter> pager, int64_t expected_rows, bool has_dictionary,
+    MemoryAllocator* allocator)
     : descr_(descr),
       pager_(std::move(pager)),
       expected_rows_(expected_rows),
+      has_dictionary_(has_dictionary),
       allocator_(allocator),
+      pool_(allocator),
       num_buffered_values_(0),
       num_buffered_encoded_values_(0),
       num_rows_(0),
@@ -47,7 +51,6 @@ ColumnWriter::ColumnWriter(const ColumnDescriptor* descr,
 void ColumnWriter::InitSinks() {
   definition_levels_sink_.reset(new InMemoryOutputStream());
   repetition_levels_sink_.reset(new InMemoryOutputStream());
-  values_sink_.reset(new InMemoryOutputStream());
 }
 
 void ColumnWriter::WriteDefinitionLevels(int64_t num_levels, const int16_t* levels) {
@@ -84,7 +87,7 @@ void ColumnWriter::WriteNewPage() {
   // TODO: Currently we only support writing DataPages
   std::shared_ptr<Buffer> definition_levels = definition_levels_sink_->GetBuffer();
   std::shared_ptr<Buffer> repetition_levels = repetition_levels_sink_->GetBuffer();
-  std::shared_ptr<Buffer> values = values_sink_->GetBuffer();
+  std::shared_ptr<Buffer> values = GetValuesBuffer();
 
   if (descr_->max_definition_level() > 0) {
     definition_levels =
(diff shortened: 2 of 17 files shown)


wesm:
Will review this as soon as I can

src/parquet/column/writer.cc
```diff
@@ -130,17 +136,42 @@ template <typename Type>
...
           std::unique_ptr<EncoderType>(new PlainEncoder<Type>(schema, allocator));
       break;
+    case Encoding::PLAIN_DICTIONARY:
+      current_encoder_ =
+          std::unique_ptr<EncoderType>(new DictEncoder<Type>(schema, &pool_, allocator));
+      break;
+    case Encoding::RLE_DICTIONARY:
+      current_encoder_ =
+          std::unique_ptr<EncoderType>(new DictEncoder<Type>(schema, &pool_, allocator));
+      break;
```
majetideepak: can we combine the above two cases `PLAIN_DICTIONARY` and `RLE_DICTIONARY`?

src/parquet/encodings/plain-encoding.h
```diff
@@ -163,19 +163,41 @@ class PlainEncoder : public Encoder<DType> {
...
+      : Encoder<DType>(descr, Encoding::PLAIN, allocator),
+        values_sink_(new InMemoryOutputStream(IN_MEMORY_DEFAULT_CAPACITY, allocator)) {}
 
-  void Encode(const T* src, int num_values, OutputStream* dst) override;
+  std::shared_ptr<Buffer> FlushValues() override;
+  void Put(const T* src, int num_values) override;
+  void Encode(const T* src, int num_values, OutputStream* dst);
+
+ protected:
+  std::shared_ptr<InMemoryOutputStream> values_sink_;
```
majetideepak: What is the motivation to make the `values_sink_` a part of the Encoder and remove it from the ColumnWriter ? 
A downside I see is that we cannot share a single encoder instance among multiple ColumnWriters.

majetideepak: I guess in the current implementation this is not an issue since we are creating encoders per each ColumnWriter. But I am not sure if coupling the value buffer pointer with the encoders is a correct approach.

src/parquet/encodings/dictionary-encoding.h
```diff
@@ -194,19 +207,43 @@ class DictEncoderBase {
...
-        allocator_(allocator),
-        pool_(pool),
-        dict_encoded_size_(0) {
-    hash_slots_.Assign(hash_table_size_, HASH_SLOT_EMPTY);
-    if (!CpuInfo::initialized()) { CpuInfo::Init(); }
+  /// Encode value. Note that this does not actually write any data, just
+  /// buffers the value's index to be written later.
+  void Put(const T& value);
+
+  std::shared_ptr<Buffer> FlushValues() override {
```
majetideepak: Can we combine these to form an `Encode(const T* src, int num_values, OutputStream* dst)` interface to retain the old encoding API ?

src/parquet/column/writer.cc
```diff
@@ -97,9 +100,11 @@ void ColumnWriter::WriteNewPage() {
   }
 
   // TODO(PARQUET-590): Encodings are hard-coded
+  Encoding::type encoding = Encoding::PLAIN;
+  if (has_dictionary_) { encoding = Encoding::PLAIN_DICTIONARY; }
```
wesm: Add a TODO that this is different in Parquet V2 (where the data pages have `RLE_DICTIONARY` encoding). If we are encoding for V1 writers, then we must use `PLAIN_DICTIONARY` (and the dictionary itself must be written out in plain encoding)

src/parquet/column/writer.h
```diff
@@ -174,12 +183,12 @@ inline void TypedColumnWriter<DType>::WriteBatch(int64_t num_values,
   num_buffered_encoded_values_ += values_to_write;
 
   // TODO(PARQUET-591): Instead of rows as a boundary, do a size check
-  if (num_buffered_values_ >= PAGE_VALUE_COUNT) { WriteNewPage(); }
+  if ((num_buffered_values_ >= PAGE_VALUE_COUNT) && !has_dictionary_) { WriteNewPage(); }
```
wesm: I believe other Parquet implementations accumulate all of the encoded data pages in memory prior to writing them all (possibly prefixed with a dictionary page) -- (see the "column page write store" concept in parquet-mr or the implementation of `BaseColumnWriter::Flush()` in Impala https://github.com/apache/incubator-impala/blob/master/be/src/exec/hdfs-parquet-table-writer.cc#L444).

We may scope this work in additional JIRAs but we'll need to do something similar, and also setting some kind of an optional dictionary size threshold so that when the dictionary grows to a certain size that we can switch over to plain encoding. You can imagine a worst-case scenario where you select dictionary encoding for strings in a large table and all of the observed strings are unique (so that the dictionary page could grow arbitrarily large)

wesm: If/when we do this, we can allocate memory for the encoded/compressed data pages from the common `MemPool` which you added already to this class

src/parquet/encodings/dictionary-encoding.h
wesm: Per the discussion above -- we will need to accumulate data pages containing dictionary indices in memory prior to writing to the `OutputStream*` (because the dictionary continues to grow as more data is observed, and when it reaches a certain size we will want to have the option to switch over to PLAIN / some other encoding to prevent the dictionary page from becoming too big)

src/parquet/encodings/plain-encoding.h
wesm: Per the above discussion about accumulating data pages in memory, we may consider nixing OutputStream from the encoder API. May need to do some more analysis of how other implementations have had to deal with this. For example, parquet-mr has

https://github.com/apache/parquet-mr/blob/master/parquet-hadoop/src/main/java/org/apache/parquet/hadoop/ColumnChunkPageWriteStore.java#L81

and then the concatenated data pages are written out to the output stream after the dictionary page (if there is one)

https://github.com/apache/parquet-mr/blob/master/parquet-hadoop/src/main/java/org/apache/parquet/hadoop/ColumnChunkPageWriteStore.java#L187

src/parquet/encodings/plain-encoding.h
```diff
@@ -207,9 +229,25 @@ class PlainEncoder<BooleanType> : public Encoder<BooleanType> {
...
 template <typename DType>
+inline std::shared_ptr<Buffer> PlainEncoder<DType>::FlushValues() {
+  std::shared_ptr<Buffer> buffer = values_sink_->GetBuffer();
+  values_sink_.reset(
+      new InMemoryOutputStream(IN_MEMORY_DEFAULT_CAPACITY, this->allocator_));
+  return buffer;
+}
+
+template <typename DType>
+inline void PlainEncoder<DType>::Put(const T* buffer, int num_values) {
```
wesm: Per above, we may make this the only "public" encoding API

src/parquet/file/writer-internal.cc
```diff
@@ -114,6 +114,48 @@ int64_t SerializedPageWriter::WriteDataPage(int32_t num_rows, int32_t num_values
   return sink_->Tell() - start_pos;
 }
 
+int64_t SerializedPageWriter::WriteDictionaryPage(
+    int32_t num_values, const std::shared_ptr<Buffer>& values, Encoding::type encoding) {
+  int64_t uncompressed_size = values->size();
+
+  // Compress the data
```
wesm: Possible to share code between here and `WriteDataPage`?

wesm:
Let me know what you think on the comments and we can create some follow up JIRAs (e.g. switching the encoding of data pages mid-stream when the dictionary gets too big)

src/parquet/column/writer.cc
xhochy: yes, we should.

src/parquet/encodings/plain-encoding.h
xhochy: Also we cannot share the encoder as with DictionaryEncoding, the encoder itself carry state (i.e. the dictionary).

xhochy:
@wesm @majetideepak as in-source comments are a bit cluttered in github, I'm continuing the discussion about the pages here: 

I would: 

- [ ] Create a JIRA for switching from DICTIONARY to PLAIN encoding if the dict gets too large
- [ ] Implement the "keep all pages in RAM" in this PR. For the ones where we can write them directly, I would have a look at adding a fast path directly in this PR, if it's too complicated: Going to make a JIRA

wesm:
Sounds good to me

xhochy:
For the "keep all pages" in RAM we have to make a design decision first:

1. Store the triplet of (repetition_levels, definition_levels, values) for each data page and at the end pass them to the SerializedPageWriter
2. Move some logic out of the PageWriter and store an array of the yet produced pages. At the end pass these pages as blobs to the SerializedPageWriter (that then only appends them to the output stream).
3. Buffer all values of a column and only produce the pages at the very end.

I would prefer option 1 but would love to hear feedback on this.

majetideepak:
We have to trade memory (option 3) vs compute (option 1). Option 3 is actually fast since we do not have to regenerate the original values (if at all due to PARQUET-687). 
It definitely makes sense to go with option 1. Do we know which approach parquet-mr takes ?

wesm:
The other Parquet implementations I'm aware of (Impala, parquet-mr) produce serialized (compressed, encoded) data pages as they go -- we should probably do the same

wesm:
See https://github.com/apache/parquet-mr/blob/master/parquet-column/src/main/java/org/apache/parquet/column/impl/ColumnWriterV1.java#L100

When each page is completed, the serialized byte buffer is computed:

https://github.com/apache/parquet-mr/blob/master/parquet-hadoop/src/main/java/org/apache/parquet/hadoop/ColumnChunkPageWriteStore.java#L86

Impala has a default data page size of 64K values and delimits pages whenever either the definition levels buffers or the values buffer (according to the max data page size -- 1GB at present) fills up:

https://github.com/apache/incubator-impala/blob/master/be/src/exec/hdfs-parquet-table-writer.cc#L100

https://github.com/apache/incubator-impala/blob/master/be/src/exec/hdfs-parquet-table-writer.cc#L377

wesm:
Note we have a somewhat tangentially related bug around the data page write path: https://issues.apache.org/jira/browse/PARQUET-676

xhochy:
This then is more in the direction of Option 2, I would then implement this.

majetideepak:
If we are going to keep pages in memory till the end, is there any advantage to serialize them eagerly ?
I don't think there will be much benefit from compression as well since we are dealing with in memory data.

xhochy:
I don't see a benefit in serialising them eagerly but compressing them will reduce the memory footprint.

xhochy:
The main benefit of serialising them directly is that we only need to append them at the end of the outputstream when we close the ColumnWriter, otherwise a call to `ColumnWriter::Close` will be much more time consuming. Although this is just shifting work around, it always feels better to have a fast close method.

majetideepak:
My only concern is that if we have to switch from DICTIONARY to PLAIN (PARQUET-687) midway, then the `serialization -- de-serialization` cycles are wasted in the eager approach.

wesm:
@majetideepak that's actually not true IIUC. There's no problem (at least per the Parquet format) having both dictionary-encoded data pages and plain encoded data pages in the same column chunk (the page encoding switches from PLAIN_DICTIONARY or RLE_DICTIONARY to PLAIN, for example). We aren't testing this yet in parquet-cpp AFAIK

At minimum, after reaching the end of a particular data page, the data is already partially serialized in the encoders, so you have to decide whether to:

- Keep around the RLE-encoded level buffers and value encoder buffers, and page header metadata, and write these out into the output stream later

- Concatenate the serialized metadata and encoded buffers right away into a single contiguous in-memory buffer, which later gets written to the output stream

The latter is easier implementation wise. The format is probably more work since you must take ownership over the internal buffers in the encoder classes and have them allocate new memory (and then do the concatenation later).

majetideepak:
@wesm thank you for the clarification. I didn't know that different encodings are possible in the same column chunk. This makes sense now.

wesm:
We also will need to mind the `encodings` field in the `ColumnMetaData`

https://github.com/apache/parquet-format/blob/master/src/main/thrift/parquet.thrift#L470

This means we also need to keep track of what encodings have been used (since there may be multiple) and set those correctly in the output metadata.

xhochy:
data pages are now stored as DataPage objects and are then written at the end to the output sink on a call to `Close()`. Compression is still done as part of `Close()` but I created a JIRA to change that: https://issues.apache.org/jira/browse/PARQUET-689 Didn't want to implement this yet, so we can get this PR and the other ones merged and fix the merge conflicts before we make even more changes.

xhochy:
This PR also fixes https://issues.apache.org/jira/browse/PARQUET-591

src/parquet/column/properties.h
```diff
@@ -77,9 +77,8 @@ class PARQUET_EXPORT ReaderProperties {
 
 ReaderProperties PARQUET_EXPORT default_reader_properties();
 
-static int64_t DEFAULT_PAGE_SIZE = 1024 * 1024;
+static int64_t DEFAULT_PAGE_SIZE = 64 * 1024 * 1024;
```
wesm: The default used elsewhere is much smaller. Impala uses 64K, parquet-mr uses 1M. In the absence of benchmarks against different IO sources (e.g. file system vs HDFS) we can probably stick with the 1M default

src/parquet/column/properties.h
```diff
@@ -162,14 +150,13 @@ class PARQUET_EXPORT WriterProperties {
...
-          dictionary_enabled_, dictionary_pagesize_, default_encoding_, encodings_,
-          pagesize_, version_, default_codec_, codecs_));
+      return std::shared_ptr<WriterProperties>(
+          new WriterProperties(allocator_, dictionary_pagesize_, default_encoding_,
+              encodings_, pagesize_, version_, default_codec_, codecs_));
     }
 
    private:
     MemoryAllocator* allocator_;
-    bool dictionary_enabled_;
```
wesm: Based also on the discussions in PARQUET-573 this will need to be restored either now or later to a per-column on/off state (and if the the writer "exits" dictionary encoding due to the dictionary growing too large). I'm ok with punting on this here and tackling in a follow up PR or in #143)

src/parquet/column/writer.cc
```diff
@@ -96,21 +100,42 @@ void ColumnWriter::WriteNewPage() {
...
+  uint8_t* uncompressed_ptr = uncompressed_data->mutable_data();
+  memcpy(uncompressed_ptr, repetition_levels->data(), repetition_levels->size());
+  uncompressed_ptr += repetition_levels->size();
+  memcpy(uncompressed_ptr, definition_levels->data(), definition_levels->size());
+  uncompressed_ptr += definition_levels->size();
+  memcpy(uncompressed_ptr, values->data(), values->size());
+  DataPage page(
+      uncompressed_data, num_buffered_values_, encoding_, Encoding::RLE, Encoding::RLE);
+
+  data_pages_.push_back(std::move(page));
```
wesm: You could use `emplace_back` here

wesm:
+1. Can you create a JIRA about a default data page size (whether we use the one in parquet-mr or some other)? Going to merge this so @majetideepak can resolve rebase conflicts in #143 and we can get the dictionary encoding configuration details settled (there probably needs to be the option to toggle it on/off for any given column, with the "fallback encoding" also configurable -- defaulting to PLAIN of course)

src/parquet/column/properties.h
xhochy: Misread the double 1024 here. Thought it was only 1k, so I upped to mentally 64k. We should stay with 1M probably. Will revert this.

xhochy: JIRA for the size selection: https://issues.apache.org/jira/browse/PARQUET-693

src/parquet/column/properties.h
xhochy: We currently only need the `encoding` parameter to determine how we encode the data. This was currently totally redundant information and I don't like to keep variables that are possibly needed (or not) for future implementation.

xhochy:
on/off toggling is fully supported at the moment by setting the encoding variable. Thus I removed the has_dictionary switch. You can select on a per-column basis if you want to have dictionary encoding. Only the fallback mode incl. the fallback codec is missing.

wesm:
it seems most Parquet implementations have dictionary encoding turned on by default; so we can let users turn it off by default, or on/off for a particular column

xhochy:
Yes, that's just a matter of the DEFAULT_ENCODING constant.

wesm:
We've been going in circles about this! I think the simplest way forward and most consistent with other implementations is:

- boolean flag for dictionary encoding on/off
- value encoding indicating what encoding should be used if we are not dictionary encoding (either dictionary encoding is off, or the dictionary grew too large)

the problem with using `DEFAULT_ENCODING` is that it leaks the V1/V2 encoding types (e.g. PLAIN_DICTIONARY) to the user, which is not desirable IMHO.

xhochy:
The PLAIN_DICTIONARY vs RLE_DICTIONARY encoding is already handled by the merged coded transparently, i.e. the implementation does write the right(TM) metadata.

Summed up the current state/discussion in https://issues.apache.org/jira/browse/PARQUET-695, hope this captures everything.

PR #144, opened 2016-08-22
PARQUET-681: Add tool to scan a parquet file

majetideepak:
Added a ReadBatchValues() API to the Column class.
Added a parquet-scan tool
Separated examples into benchmarks/tools
added clang tidy and clang format to benchmarks and tools


src/parquet/column/reader.h
```diff
@@ -44,6 +44,9 @@ class PARQUET_EXPORT ColumnReader {
   static std::shared_ptr<ColumnReader> Make(const ColumnDescriptor*,
       std::unique_ptr<PageReader>, MemoryAllocator* allocator = default_allocator());
 
+  virtual int64_t ReadBatchValues(int32_t batch_size, int16_t* def_levels,
+      int16_t* rep_levels, uint8_t* values, int64_t* values_read) = 0;
```
majetideepak: This API enables the downstream applications to use their own buffers to read Parquet values.

majetideepak: @wesm  and @xhochy, do you have any comments on this API ?

wesm: I think it would be preferable to have `reinterpret_cast<T*>(values_buffer)` on the user end, to be more explicit. There is the `T` member of the typed column readers that you can use.

benchmarks/.clang-format
```diff
@@ -0,0 +1,65 @@
+---
```
wesm: Is it possible to put these files in the root directory? We can also nested them under `src/` if there is some complication

src/parquet/types.cc
```diff
@@ -213,4 +213,29 @@ std::string logical_type_to_string(LogicalType::type t) {
       break;
   }
 }
+
+size_t get_value_byte_size(Type::type parquet_type) {
```
wesm: return `int` here

src/parquet/types.h
```diff
@@ -276,6 +276,8 @@ std::string logical_type_to_string(LogicalType::type t);
 std::string type_to_string(Type::type t);
 
 std::string FormatStatValue(Type::type parquet_type, const char* val);
+
+size_t get_value_byte_size(Type::type t);
```
wesm: Think we need to fix the naming for the other methods in this file to be in line with Google SG (e.g. `TypeToString` and `GetTypeByteSize`)

tools/parquet-scan.cc
```diff
@@ -0,0 +1,109 @@
...
+      for (auto i : columns) {
+        total_rows[col] = 0;
+        std::shared_ptr<parquet::ColumnReader> col_reader = group_reader->Column(i);
+        size_t value_byte_size =
+            get_value_byte_size(col_reader->descr()->physical_type());
+        std::vector<uint8_t> values(batch_size * value_byte_size);
+
+        int64_t values_read = 0;
+        while (col_reader->HasNext()) {
+          total_rows[col] += col_reader->ReadBatchValues(batch_size, def_levels.data(),
```
wesm: I see. The problem solved by the method above is not knowing the column types at runtime. I am not confident enough whether this is a good idea or bad idea (my gut is that it isn't the best because we also write out the ByteArray / FixedLenByteArray structs), so for now I would suggest a `ScanAllValues` method that internally invokes a `ScanAll<ColumnReaderType>(col_reader)`.

src/parquet/column/reader.h
xhochy: I would also prefer to have the cast on the user side and use the `ReadBatch` we currently export.

wesm:
Want to rebase this?

majetideepak:
I plan to finish this patch by evening today. I have some changes I need to test. I added the `ScanAllValues` API.

src/parquet/column/scanner.h
```diff
@@ -48,7 +48,6 @@ class PARQUET_EXPORT Scanner {
         value_offset_(0),
         values_buffered_(0),
         reader_(reader) {
-    // TODO: don't allocate for required fields
```
majetideepak: This has been implemented below

src/parquet/column/reader.h
```diff
@@ -58,6 +58,11 @@ class PARQUET_EXPORT ColumnReader {
 
   const ColumnDescriptor* descr() const { return descr_; }
 
+  int64_t GetRemainingInPage() {
```
majetideepak: In our current implementation, for a given column `batch size`, values are not returned across DataPages.
Since, different columns (based on the type) can have different number of values in a `DataPage`, it becomes difficult to specify a `batch size` such that all columns return the same number of values.  i.e. a batch of rows is not returned.
The purpose of this is API is to query the number of remaining values in a Page for a given column. The client can now take the minimum of `GetRemainingInPage` in all the columns to specify a `batch size`.

majetideepak:
I verified `parquet-scan` locally. I am not sure how to test the tools.

wesm:
What do you think about making this scan function into a public API that you invoke from parquet-scan? I suspect this can help with microbenchmarking / performance tuning

majetideepak:
I made the scan function into a public API.

src/parquet/column/reader.h
wesm: Is this used in the patch? If it's not needed let's take this out

wesm:
Minor comment above, then can merge this

src/parquet/column/reader.h
majetideepak: Removed it.

wesm:
Cool. +1, thank you!

PR #143, opened 2016-08-18
PARQUET-573: Create a public API for reading and writing file metadata

majetideepak:
This patch adds API to read and write metadata as well as improves the writer properties class.
I am planning to add some comments to the code next. Meanwhile, any feedback will be helpful.

PR #145, opened 2016-08-25
Incorrect count written to "num_values" in column metadata

mitra:
The the num_values in meta_data associated with ColumnChunk is incorrectly
set to the number of actual values present excluding the null values. It
should be the total number of values including the Null values.

If we create a parquet file with a schema with optional fields the
"num_values" field in the meta data gets set incorrectly. Such a parquet
file cannot be read using the parquet-tools in parquet-mr.

Signed-off-by: Anand Mitra <anand.mitra@gmail.com>

### DIFF
--- a/src/parquet/file/writer-internal.cc
+++ b/src/parquet/file/writer-internal.cc
@@ -109,7 +109,7 @@ int64_t SerializedPageWriter::WriteDataPage(int32_t num_rows, int32_t num_values
 
   metadata_->meta_data.total_uncompressed_size += uncompressed_size + header_size;
   metadata_->meta_data.total_compressed_size += compressed_size + header_size;
-  metadata_->meta_data.num_values += num_values;
+  metadata_->meta_data.num_values += num_rows;
 
   return sink_->Tell() - start_pos;
 }


majetideepak:
Mitra, can you create a JIRA and use that JIRA number to describe your commit here ?

wesm:
Can you reproduce this issue with a unit test?

wesm:
I believe this was fixed in PARQUET-666 / #142 -- can you confirm? Thanks

wesm:
Of course, having an explicit unit test (perhaps after #143) is merged would be ideal

xhochy:
Yes, I've incorporated this PR in #143 but I don't feel confident that we fixed this issue until we have a unit test. There are several counts in a column chunk and it's very easy to mix them up thus I fear that we will break it again by accident in the future.

mitra:
I have implemented the dremel algorithm over libparquet-cpp and was checking for interoperability on the generated parquet files. I had found that for some nested structures the generated files could not be read using parquet-mr. I then generated the exact same structures using parquet-mr and did a diff on the metadata from both the sides and could easily identify which value we were getting wrong. From there identifying the part of the code modifying the value and verifying it against my existing test case was simple.

I could create a unit test for a hand coded schema and values for repetition_levels and definition_levels but I am not sure if that would improve your confidence. Other suggestions are welcome as well.

wesm:
> I could create a unit test for a hand coded schema and values for repetition_levels and definition_levels

This is exactly what we need to be able to verify the bug fix

wesm:
Should be easy to add a unit test for this now that PARQUET-573 is merged, I will look into it if no one beats me to it

PR #147, opened 2016-09-01
PARQUET-699: Update parquet.thrift from https://github.com/apache/parquet-format

flode:
Support logical types TIME_MICROS and TIMESTAMP_MICROS

Also I think the current code was incorrect. Parquet reserved the LogicalTypes 8 and 10, but those were completely omitted types.h. So types with greater indices were mapped incorrectly.

### DIFF
--- a/src/parquet/compression/codec.cc
+++ b/src/parquet/compression/codec.cc
@@ -37,6 +37,9 @@ std::unique_ptr<Codec> Codec::Create(Compression::type codec_type) {
     case Compression::LZO:
       ParquetException::NYI("LZO codec not implemented");
       break;
+    case Compression::BROTLI:
+      ParquetException::NYI("BROTLI codec not implemented");
+      break;
     default:
       ParquetException::NYI("Unrecognized codec");
       break;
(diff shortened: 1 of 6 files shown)


wesm:
Thanks -- can you open a JIRA on issues.apache.org and prefix the PR title with "PARQUET-XXX:" where XXX is the JIRA number? 

You're right that the enum mapping was wrong because of the reserved slots, good catch =)

flode:
I created the JIRA and updated the pull request and commit message.

xhochy:
@wesm conda packages fail with `ERROR: placeholder '/opt/conda/envs/_build_placehold_placehold_placehold_placehold_placehold_placeho' too short in: conda-forge::ncurses-5.9-8`. This is certainly not due to this PR as master is also broken.

xhochy:
+1, LGTM

wesm:
@xhochy sounds like that's a conda problem, I will look when I can.

Want to try merging this patch with the PR merge tool to make sure your's ASF bits are properly set up? 

`python dev/merge_parquet_pr.py`

(must be run with Python 2.7 for now ... maybe we can fix that)

xhochy:
@wesm my ASF bits are not yet set up, so I'm not sure when I'm able to do my first merge.

wesm:
Got it. Let me know when they are and I'll let you merge your first PR

PR #148, opened 2016-09-01
PARQUET-700: Disable dictionary encoding for boolean columns

xhochy:


src/parquet/column/writer.cc
```diff
@@ -194,7 +194,8 @@ std::shared_ptr<ColumnWriter> ColumnWriter::Make(const ColumnDescriptor* descr,
     std::unique_ptr<PageWriter> pager, int64_t expected_rows,
     const WriterProperties* properties) {
   Encoding::type encoding = properties->encoding(descr->path());
-  if (properties->dictionary_enabled(descr->path())) {
+  if (properties->dictionary_enabled(descr->path()) &&
+      descr->physical_type() != Type::BOOLEAN) {
```
wesm: Can you test this?

xhochy: Yes and it even made the unit test class structure simpler.

wesm:
+1, thank you

PR #150, opened 2016-09-02
PARQUET-676: Fix incorrect MaxBufferSize for small bit widths

wesm:
This function was implemented incorrectly in the original source code from Impala. The bug never presented itself because Impala allocates a much larger buffer than is required for the data page rather than using the `MaxBufferSize` value.

In a worst case scenario, the RleEncoder can result in a concatenation of short RLE runs of 8 values (the minimum for RLE). For example:

0 0 0 0 0 0 0 0 1 1 1 1 1 1 1 1 0 0 0 0 0 0 0 0 ...

In encoded form, each RLE run occupies 2 bytes (for bit width 1). Thus 1024 values with this structure and bit width 1, at least 256 bytes are required to fully encode. If these were encoded as all literal runs, you can encode up to 504 values in a literal run (plus the indicator byte). Thus, in the same case (bit width 1) only

(504 + 504 + 16)
64 + 64 + 3 = 131

bytes are required. This should also fix PARQUET-698.

### DIFF
--- a/src/parquet/column/levels-test.cc
+++ b/src/parquet/column/levels-test.cc
@@ -183,4 +183,27 @@ TEST(TestLevels, TestLevelsDecodeMultipleSetData) {
   }
 }
 
+TEST(TestLevelEncoder, MinimumBufferSize) {
+  // PARQUET-676, PARQUET-698
+  const int kNumToEncode = 1024;
+
+  std::vector<int16_t> levels;
+  for (int i = 0; i < kNumToEncode; ++i) {
+    if (i % 9 == 0) {
+      levels.push_back(0);
+    } else {
+      levels.push_back(1);
+    }
+  }
+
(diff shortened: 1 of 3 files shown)


majetideepak:
+1 LGTM. Good catch! The math looks correct to me.
As noted in the comments of `rle_encoding.h`, the break-even point for rle is at 8 values for 1 width values taking 2 bytes for each group. The worst case scenario is the example you posted above.

majetideepak:
However, in practice, shouldn't the switching algorithm use literal run in your example ? The existing upper bounds is calculated based on just max literal run, and should work if the switching is correct. Do we need to revisit the switching logic as well ?

wesm:
Neither is right or wrong necessarily. The RLE encoding stores an extra byte but may be simpler to decode (not having to unpack each bit), but I don't think it's worth stressing over.

wesm:
+1

PR #154, opened 2016-09-06
PARQUET-710: Remove unneeded private member variables from RowGroupReader ABI

majetideepak:
The injection capability is retained.
 The private members of the `RowGroupReader` class are removed.

### DIFF
--- a/src/parquet/file/reader.cc
+++ b/src/parquet/file/reader.cc
@@ -45,11 +45,14 @@ RowGroupReader::RowGroupReader(std::unique_ptr<Contents> contents)
     : contents_(std::move(contents)) {}
 
 std::shared_ptr<ColumnReader> RowGroupReader::Column(int i) {
-  DCHECK(i < metadata()->num_columns()) << "The RowGroup only has " << metadata()->num_columns() << "columns, requested column: " << i;
+  DCHECK(i < metadata()->num_columns()) << "The RowGroup only has "
+                                        << metadata()->num_columns()
+                                        << "columns, requested column: " << i;
   const ColumnDescriptor* descr = metadata()->schema()->Column(i);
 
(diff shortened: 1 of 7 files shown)


src/parquet/file/reader.cc
```diff
@@ -41,17 +41,18 @@ namespace parquet {
...
+    : contents_(std::move(contents)) {}
 
 std::shared_ptr<ColumnReader> RowGroupReader::Column(int i) {
-  DCHECK(i < schema_->num_columns()) << "The RowGroup only has " << schema_->num_columns()
-                                     << "columns, requested column: " << i;
-  const ColumnDescriptor* descr = schema_->Column(i);
+  DCHECK(i < metadata()->num_columns()) << "The RowGroup only has "
+                                        << metadata()->num_columns()
+                                        << "columns, requested column: " << i;
+  const ColumnDescriptor* descr = metadata()->schema()->Column(i);
```
majetideepak: The `RowGroupMetadata` is extended to include `SchemaDescriptor` to support this. Alternatively, we can modify the `contents` struct to include `std::unique_ptr<PageReader> Column(int i)` instead of  `std::unique_ptr<PageReader>  GetColumnPageReader(int i)`

wesm:
Can we change the title of the patch? It is not accurate anymore. How about "Remove unneeded private member variables from RowGroupReader ABI"

src/parquet/file/reader.cc
wesm: Since this API is not intended as public I'm mostly neutral on this.

wesm:
Can you also remove the references to "PIMPL" in the comments. My mistake (the pattern is dependency injection).

xhochy:
+1, LGTM

wesm:
+1

PR #155, opened 2016-09-06
PARQUET-708: account for "worst case scenario" in MaxBufferSize for bit_width > 1

xhochy:


src/parquet/util/rle-encoding.h
```diff
@@ -171,8 +171,12 @@ class RleEncoder {
 
   /// Returns the maximum byte size it could take to encode 'num_values'.
   static int MaxBufferSize(int bit_width, int num_values) {
-    int bytes_per_run = BitUtil::Ceil(bit_width * MAX_VALUES_PER_LITERAL_RUN, 8.0);
```
majetideepak: Should we add another variable `MIN_VALUES_PER_LITERAL_RUN` to 8 and use the same previous logic here ?

src/parquet/column/levels-test.cc
```diff
@@ -206,4 +206,32 @@ TEST(TestLevelEncoder, MinimumBufferSize) {
...
+  // Test the worst case for bit_width=2 consisting of
+  // LiteralRun(size=8)
+  // RepeatedRun(size=8)
+  // LiteralRun(size=8)
+  // ...
+  const int kNumToEncode = 1024;
+
+  std::vector<int16_t> levels;
+  for (int i = 0; i < kNumToEncode; ++i) {
+    if ((i % 16) < 7) {
```
majetideepak: should  `< 8` or  ` <= 7` ?

src/parquet/column/levels-test.cc
```diff
@@ -206,4 +206,32 @@ TEST(TestLevelEncoder, MinimumBufferSize) {
...
+      levels.push_back(0);
+    } else {
+      levels.push_back(1);
+    }
+  }
+
+  std::vector<uint8_t> output(
+      LevelEncoder::MaxBufferSize(Encoding::RLE, 2, kNumToEncode));
```
majetideepak: Add tests for other bit widths may be 1, 4, 8 ?

src/parquet/util/rle-encoding.h
xhochy: This is not the minimal number of values for a literal run but rather the threshold for checking if we do a literal or a repeated run.

src/parquet/column/levels-test.cc
xhochy: No, this is `< 7` to force a literal run. This loop generates the sequence of literal and repeated runs as mentioned in the comment above.

src/parquet/column/levels-test.cc
xhochy: Will do.

majetideepak:
+1 LGTM

wesm:
+1, thank you

PR #156, opened 2016-09-07
PARQUET-711: Use metadata builders in parquet writer

majetideepak:
I wrote a sample file and the metadata seems to be correct.
@xhochy I fixed some missing metadata like `dictionary_page_offset`. You might want to check if this fixes the Drill problem.

### DIFF
--- a/src/parquet/column/column-io-benchmark.cc
+++ b/src/parquet/column/column-io-benchmark.cc
@@ -25,13 +25,12 @@
 
 namespace parquet {
 
-using format::ColumnChunk;
 using schema::PrimitiveNode;
 
 namespace benchmark {
 
 std::unique_ptr<Int64Writer> BuildWriter(int64_t output_size, OutputStream* dst,
-    ColumnChunk* metadata, ColumnDescriptor* schema, const WriterProperties* properties) {
+    ColumnChunkMetaDataBuilder* metadata, ColumnDescriptor* schema, const WriterProperties* properties) {
   std::unique_ptr<SerializedPageWriter> pager(
       new SerializedPageWriter(dst, Compression::UNCOMPRESSED, metadata));
   return std::unique_ptr<Int64Writer>(new Int64Writer(
@@ -57,17 +56,18 @@ void SetBytesProcessed(::benchmark::State& state, Repetition::type repetition) {
 
 template <Repetition::type repetition>
 static void BM_WriteInt64Column(::benchmark::State& state) {
-  format::ColumnChunk metadata;
+  format::ColumnChunk thrift_metadata;
   std::vector<int64_t> values(state.range_x(), 128);
   std::vector<int16_t> definition_levels(state.range_x(), 1);
   std::vector<int16_t> repetition_levels(state.range_x(), 0);
   std::shared_ptr<ColumnDescriptor> schema = Int64Schema(repetition);
   std::shared_ptr<parquet::WriterProperties> properties = default_writer_properties();
+  auto metadata = parquet::ColumnChunkMetaDataBuilder::Make(properties, schema.get(), reinterpret_cast<uint8_t*>(&thrift_metadata)); 
 
   while (state.KeepRunning()) {
     InMemoryOutputStream dst;
     std::unique_ptr<Int64Writer> writer =
-        BuildWriter(state.range_x(), &dst, &metadata, schema.get(), properties.get());
+        BuildWriter(state.range_x(), &dst, metadata.get(), schema.get(), properties.get());
     writer->WriteBatch(
         values.size(), definition_levels.data(), repetition_levels.data(), values.data());
     writer->Close();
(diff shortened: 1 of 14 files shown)


src/parquet/file/writer-internal.cc
```diff
@@ -133,50 +133,38 @@ int64_t SerializedPageWriter::WriteDictionaryPage(const DictionaryPage& page) {
...
 
     if (current_column_writer_) {
-      total_bytes_written_ += current_column_writer_->Close();
+      total_bytes_written_ = current_column_writer_->Close();
       current_column_writer_.reset();
     }
-
-    metadata_->__set_total_byte_size(total_bytes_written_);
+    // Ensures all columns have been written
+    metadata_->Finish(total_bytes_written_);
```
majetideepak: `total_bytes_written_` value updated by the `ColumnWriter` is actually redundant. `total_bytes_written_` is computed in the metadata builder by adding the `compressed_size` of all columns. 
I am still passing the value here for checking correctness. There is a DCHECK inside `RowGroupMetaDataBuilder::Finish()`.

src/parquet/column/column-io-benchmark.cc
```diff
@@ -57,17 +57,19 @@ void SetBytesProcessed(::benchmark::State& state, Repetition::type repetition) {
...
 template <Repetition::type repetition>
 static void BM_WriteInt64Column(::benchmark::State& state) {
-  format::ColumnChunk metadata;
+  format::ColumnChunk thrift_metadata;
   std::vector<int64_t> values(state.range_x(), 128);
   std::vector<int16_t> definition_levels(state.range_x(), 1);
   std::vector<int16_t> repetition_levels(state.range_x(), 0);
   std::shared_ptr<ColumnDescriptor> schema = Int64Schema(repetition);
   std::shared_ptr<parquet::WriterProperties> properties = default_writer_properties();
+  auto metadata = parquet::ColumnChunkMetaDataBuilder::Make(
```
xhochy: No need for the `parquet::` prefix here, the benchmarks are also in the same namespace.

src/parquet/file/writer.h
```diff
@@ -152,19 +143,9 @@ class PARQUET_EXPORT ParquetFileWriter {
    */
   const std::shared_ptr<WriterProperties>& properties() const;
 
-  /**
-   * Returns the file schema descriptor
-   */
-  const SchemaDescriptor* descr() { return schema_; }
```
xhochy: These two function are quite comfortable, I'd like to preserve them. It's ok for me if they call a bit deeper but they are just convenient from an API user level.

xhochy:
@majetideepak Perfect! This fixes the Drill issue.

xhochy:
Except for my wish to keep the two helper functions: LGTM

majetideepak:
I will add those two helper functions back

src/parquet/file/writer-internal.h
```diff
@@ -67,24 +71,17 @@ class SerializedPageWriter : public PageWriter {
...
-  RowGroupSerializer(int64_t num_rows, const SchemaDescriptor* schema, OutputStream* sink,
-      format::RowGroup* metadata, const WriterProperties* properties)
+  RowGroupSerializer(int64_t num_rows, OutputStream* sink,
+      RowGroupMetaDataBuilder* metadata, const WriterProperties* properties)
       : num_rows_(num_rows),
-        schema_(schema),
         sink_(sink),
-        metadata_(metadata),
-        allocator_(properties->allocator()),
+        metadata_(std::move(metadata)),
```
wesm: This move not needed?

src/parquet/file/writer.cc
```diff
@@ -64,9 +61,16 @@ std::unique_ptr<ParquetFileWriter> ParquetFileWriter::Open(
   return result;
 }
 
+const SchemaDescriptor* ParquetFileWriter::descr() const {
```
wesm: Maybe call this `schema()`?

wesm:
I admit when to use `descr` or `schema` names is unclear. We should not use `schema` to refer to the `GroupNode` instance that initializes the `SchemaDescriptor`, and perhaps then `schema` should only refer to the `SchemaDescriptor` object (and use `descr` when referring to `ColumnDescriptor`). Doesn't need to be all handled in this patch.

majetideepak:
Let's use `schema` for `SchemaDescriptor`, `descr` for `ColumnDescriptor` and `schema_node` for the `GroupNode`. I will implement PARQUET-705 and PARQUET-707 in this patch

src/parquet/schema/descriptor.h
```diff
@@ -100,18 +100,20 @@ class PARQUET_EXPORT SchemaDescriptor {
...
   const schema::NodePtr& GetColumnRoot(int i) const;
 
+  const std::string& name() const { return group_node_->name(); }
+
  private:
   friend class ColumnDescriptor;
 
   schema::NodePtr schema_;
-  const schema::GroupNode* group_;
+  const schema::GroupNode* group_node_;
```
majetideepak: This is consistent with `primitive_node_` used in the `ColumnDescriptor`

src/parquet/schema/descriptor.h
```diff
@@ -100,18 +100,20 @@ class PARQUET_EXPORT SchemaDescriptor {
...
-  const schema::NodePtr& schema() const { return schema_; }
+  const schema::NodePtr& schema_root() const { return schema_; }
 
-  const schema::GroupNode* group() const { return group_; }
+  const schema::GroupNode* group_node() const { return group_node_; }
 
   // Returns the root (child of the schema root) node of the leaf(column) node
   const schema::NodePtr& GetColumnRoot(int i) const;
 
+  const std::string& name() const { return group_node_->name(); }
```
majetideepak: PARQUET-707

wesm:
+1, thank you!

PR #157, opened 2016-09-12
PARQUET-687: C++: Switch to PLAIN encoding if dictionary grows too large

majetideepak:
Implemented dictionary fallback encoding
Added tests
Added a fast path to serialize data pages

### DIFF
--- a/src/parquet/column/column-writer-test.cc
+++ b/src/parquet/column/column-writer-test.cc
@@ -133,15 +133,21 @@ class TestPrimitiveWriter : public ::testing::Test {
   }
 
   int64_t metadata_num_values() {
+    // Metadata accessor must be created lazily.
+    // This is because the ColumnChunkMetaData semantics dictate the metadata object is
+    // complete (no changes to the metadata buffer can be made after instantiation)
     auto metadata_accessor =
-        ColumnChunkMetaData::Make(reinterpret_cast<uint8_t*>(&thrift_metadata_));
+        ColumnChunkMetaData::Make(reinterpret_cast<const uint8_t*>(&thrift_metadata_));
     return metadata_accessor->num_values();
   }
 
-  int64_t metadata_num_encodings() {
+  std::vector<Encoding::type> metadata_encodings() {
+    // Metadata accessor must be created lazily.
+    // This is because the ColumnChunkMetaData semantics dictate the metadata object is
+    // complete (no changes to the metadata buffer can be made after instantiation)
     auto metadata_accessor =
-        ColumnChunkMetaData::Make(reinterpret_cast<uint8_t*>(&thrift_metadata_));
-    return metadata_accessor->encodings().size();
+        ColumnChunkMetaData::Make(reinterpret_cast<const uint8_t*>(&thrift_metadata_));
+    return metadata_accessor->encodings();
   }
 
  protected:
@@ -358,13 +364,16 @@ TYPED_TEST(TestPrimitiveWriter, RequiredVeryLargeChunk) {
   ASSERT_EQ(SMALL_SIZE, this->values_read_);
   this->values_.resize(SMALL_SIZE);
   ASSERT_EQ(this->values_, this->values_out_);
+  std::vector<Encoding::type> encodings = this->metadata_encodings();
+  // There are 3 encodings (RLE, PLAIN_DICTIONARY, PLAIN) in a fallback case
+  // Dictionary encoding is not allowed for boolean type
+  // There are 2 encodings (RLE, PLAIN) in a non dictionary encoding case
+  ASSERT_EQ(Encoding::RLE, encodings[0]);
   if (this->type_num() != Type::BOOLEAN) {
-    // There are 3 encodings (RLE, PLAIN_DICTIONARY, PLAIN) in a fallback case
-    ASSERT_EQ(3, this->metadata_num_encodings());
+    ASSERT_EQ(Encoding::PLAIN_DICTIONARY, encodings[1]);
+    ASSERT_EQ(Encoding::PLAIN, encodings[2]);
   } else {
-    // Dictionary encoding is not allowed for boolean type
-    // There are 2 encodings (RLE, PLAIN) in a non dictionary encoding case
-    ASSERT_EQ(2, this->metadata_num_encodings());
+    ASSERT_EQ(Encoding::PLAIN, encodings[1]);
   }
 }
 
--- a/src/parquet/column/column-writer-test.cc
+++ b/src/parquet/column/column-writer-test.cc
@@ -351,7 +351,7 @@ TYPED_TEST(TestPrimitiveWriter, RequiredLargeChunk) {
   ASSERT_EQ(this->values_, this->values_out_);
 }
 
-// Test case for dictionary fallback encoding skip BooleanType
+// Test case for dictionary fallback encoding
 TYPED_TEST(TestPrimitiveWriter, RequiredVeryLargeChunk) {
   this->GenerateData(VERY_LARGE_SIZE);
 
--- a/src/parquet/column/column-writer-test.cc
+++ b/src/parquet/column/column-writer-test.cc
@@ -40,7 +40,7 @@ const int SMALL_SIZE = 100;
 // Larger size to test some corner cases, only used in some specific cases.
 const int LARGE_SIZE = 10000;
 // Very large size to test dictionary fallback.
-const int VERY_LARGE_SIZE = 1000000;
+const int VERY_LARGE_SIZE = 400000;
 
 template <typename TestType>
 class TestPrimitiveWriter : public ::testing::Test {
@@ -77,6 +77,8 @@ class TestPrimitiveWriter : public ::testing::Test {
 
     SetUpSchemaRequired();
   }
+
+  Type::type type_num() { return TestType::type_num; }
 
   void BuildReader() {
     auto buffer = sink_->GetBuffer();
@@ -214,8 +216,6 @@ void TestPrimitiveWriter<BooleanType>::GenerateData(int64_t num_values) {
 typedef ::testing::Types<Int32Type, Int64Type, Int96Type, FloatType, DoubleType,
     BooleanType, ByteArrayType, FLBAType> TestTypes;
 
-typedef TestPrimitiveWriter<Int32Type> TestInt32PrimitiveWriter;
-
 TYPED_TEST_CASE(TestPrimitiveWriter, TestTypes);
 
 TYPED_TEST(TestPrimitiveWriter, RequiredPlain) {
@@ -345,8 +345,8 @@ TYPED_TEST(TestPrimitiveWriter, RequiredLargeChunk) {
   ASSERT_EQ(this->values_, this->values_out_);
 }
 
-// Test case for dictionary fallback encoding
-TEST_F(TestInt32PrimitiveWriter, RequiredVeryLargeChunk) {
+// Test case for dictionary fallback encoding skip BooleanType
+TYPED_TEST(TestPrimitiveWriter, RequiredVeryLargeChunk) {
   this->GenerateData(VERY_LARGE_SIZE);
 
   auto writer = this->BuildWriter(VERY_LARGE_SIZE, Encoding::PLAIN_DICTIONARY);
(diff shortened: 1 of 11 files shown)


src/parquet/column/column-writer-test.cc
```diff
@@ -130,7 +132,17 @@ class TestPrimitiveWriter : public ::testing::Test {
     ASSERT_EQ(this->values_, this->values_out_);
   }
 
-  int64_t metadata_num_values() const { return metadata_accessor_->num_values(); }
+  int64_t metadata_num_values() {
+    auto metadata_accessor =
+        ColumnChunkMetaData::Make(reinterpret_cast<uint8_t*>(&thrift_metadata_));
```
majetideepak: Metadata accessor needs to be built lazily.

src/parquet/column/writer.cc
```diff
@@ -118,7 +119,13 @@ void ColumnWriter::AddDataPage() {
...
       uncompressed_data, num_buffered_values_, encoding_, Encoding::RLE, Encoding::RLE);
 
-  data_pages_.push_back(std::move(page));
+  // Write the page to OutputStream eagerly if there is no dictionary or
+  // if dictionary encoding has fallen back to PLAIN
+  if (has_dictionary_ && !fallback_) {  // Save pages until end of dictionary encoding
+    data_pages_.push_back(std::move(page));
+  } else {  // Eagerly write pages
+    WriteDataPage(page);
+  }
```
majetideepak: Fast path to serialize data pages.

src/parquet/file/metadata.cc
```diff
@@ -374,10 +374,12 @@ class ColumnChunkMetaDataBuilder::ColumnChunkMetaDataBuilderImpl {
       if (properties_->version() == ParquetVersion::PARQUET_2_0) {
         thrift_encodings.push_back(ToThrift(properties_->dictionary_index_encoding()));
       }
-    }
-    if (!properties_->dictionary_enabled(column_->path()) || dictionary_fallback) {
+    } else {  // Dictionary not enabled
       thrift_encodings.push_back(ToThrift(properties_->encoding(column_->path())));
     }
+    if (dictionary_fallback) {  // Only PLAIN encoding is supported for fallback in V1
+      thrift_encodings.push_back(ToThrift(Encoding::PLAIN));
```
majetideepak: hard code fallback encoding to `PLAIN` for now. For V2, we can use user specified encoding.

src/parquet/column/writer.h
```diff
@@ -189,6 +190,22 @@ inline void TypedColumnWriter<DType>::WriteBatch(int64_t num_values,
   if (current_encoder_->EstimatedDataEncodedSize() >= properties_->data_pagesize()) {
     AddDataPage();
   }
+  if (has_dictionary_ && !fallback_) { VerifyDictionaryFallback(); }
+}
+
+template <typename DType>
+inline void TypedColumnWriter<DType>::WriteBatch(int64_t num_values,
```
majetideepak: We check for `DataPage` limits only after we have inserted the values. If a user writes a large number of values, the `DataPage` size can become much above the limit. The purpose of this chunking is to bound this. Even if a user writes large number of values, the chunking will ensure the `AddDataPage()` is called at a reasonable pagesize limit (above the default limit).

src/parquet/column/writer.h
```diff
@@ -189,6 +190,22 @@ inline void TypedColumnWriter<DType>::WriteBatch(int64_t num_values,
   if (current_encoder_->EstimatedDataEncodedSize() >= properties_->data_pagesize()) {
```
majetideepak: The values are already encoded  `WriteValues()` at this point. We could be going way above the limit `properties_->data_pagesize()` if `num_values` is high.

src/parquet/column/writer.h
wesm: Add this as a code comment rather than on GitHub?

src/parquet/file/metadata.cc
wesm: Add a TODO / put this in a comment?

src/parquet/column/writer.h
```diff
@@ -103,13 +105,15 @@ class PARQUET_EXPORT ColumnWriter {
   int total_bytes_written_;
   bool closed_;
 
+  bool fallback_;
```
wesm: Add a comment here?

src/parquet/column/writer.h
```diff
@@ -57,6 +58,7 @@ class PARQUET_EXPORT ColumnWriter {
  protected:
   virtual std::shared_ptr<Buffer> GetValuesBuffer() = 0;
   virtual void WriteDictionaryPage() = 0;
+  virtual void VerifyDictionaryFallback() = 0;
```
wesm: Comment describing this method?

src/parquet/column/column-writer-test.cc
wesm: Put this in a comment?

src/parquet/column/column-writer-test.cc
```diff
@@ -334,5 +345,28 @@ TYPED_TEST(TestPrimitiveWriter, RequiredLargeChunk) {
...
+  writer->Close();
+
+  // Just read the first SMALL_SIZE rows to ensure we could read it back in
+  this->ReadColumn();
+  ASSERT_EQ(SMALL_SIZE, this->values_read_);
+  this->values_.resize(SMALL_SIZE);
+  ASSERT_EQ(this->values_, this->values_out_);
+  if (this->type_num() != Type::BOOLEAN) {
+    // There are 3 encodings (RLE, PLAIN_DICTIONARY, PLAIN) in a fallback case
+    ASSERT_EQ(3, this->metadata_num_encodings());
```
wesm: At this risk of being pedantic, I wonder if we should explicitly check that the encodings vector contains either `PLAIN_DICTIONARY` or `RLE_DICTIONARY`

src/parquet/column/page.h
```diff
@@ -171,7 +171,7 @@ class PageWriter {
  public:
   virtual ~PageWriter() {}
 
-  virtual void Close() = 0;
+  virtual void Close(bool fallback) = 0;
```
wesm: Comment for this fallback parameter

src/parquet/column/writer.cc
```diff
@@ -134,52 +141,72 @@ void ColumnWriter::WriteDataPage(const DataPage& page) {
 int64_t ColumnWriter::Close() {
   if (!closed_) {
     closed_ = true;
-    if (has_dictionary_) { WriteDictionaryPage(); }
+    if (has_dictionary_ && !fallback_) { WriteDictionaryPage(); }
```
wesm: Is this correct? You write dictionary index pages until the dictionary exceeds a particular size. You still need the dictionary

src/parquet/column/writer.cc
```diff
@@ -118,7 +119,13 @@ void ColumnWriter::AddDataPage() {
   DataPage page(
       uncompressed_data, num_buffered_values_, encoding_, Encoding::RLE, Encoding::RLE);
 
-  data_pages_.push_back(std::move(page));
+  // Write the page to OutputStream eagerly if there is no dictionary or
+  // if dictionary encoding has fallen back to PLAIN
```
wesm: See below. I don't think this is correct

wesm: Sorry, my mistake, `data_pages_` is cleared after the fallback code path. Gerrit draft comments would be useful here

src/parquet/column/writer.cc
wesm: Hadn't read the `VerifyDictionaryFallback` yet

src/parquet/column/writer.cc
```diff
@@ -134,52 +141,72 @@ void ColumnWriter::WriteDataPage(const DataPage& page) {
...
+    WriteDictionaryPage();
+    // Write the buffered Dictionary Indicies
+    AddDataPage();
+    for (size_t i = 0; i < data_pages_.size(); i++) {
+      WriteDataPage(data_pages_[i]);
+    }
+    data_pages_.clear();
+    fallback_ = true;
+    // Only PLAIN encoding is supported for fallback in V1
+    current_encoder_ = std::unique_ptr<EncoderType>(
```
wesm: `reset` might be more concise (same in the other usages above)

src/parquet/column/writer.h
wesm: Maybe call this `CheckDictionarySizeLimit`?

src/parquet/column/writer.cc
```diff
@@ -134,52 +141,72 @@ void ColumnWriter::WriteDataPage(const DataPage& page) {
...
       ParquetException::NYI("Selected encoding is not supported");
   }
 }
 
+// Only one Dictionary Page is written.
+// Fallback to PLAIN if dictionary page limit is reached.
+template <typename Type>
+void TypedColumnWriter<Type>::VerifyDictionaryFallback() {
+  auto dict_encoder = static_cast<DictEncoder<Type>*>(current_encoder_.get());
+  if (dict_encoder->dict_encoded_size() >= properties_->dictionary_pagesize()) {
```
wesm: this property could be named more clearly. `dictionary_pagesize_limit()`?

wesm:
This looks good. I think for even integer / float data with a lot of repeated values that dictionary encoding can have a lot of perf benefits. Will be interesting to do some benchmarking to see how much

src/parquet/column/writer.h
xhochy: +1 for the  `Check` naming

src/parquet/column/writer.h
xhochy: Please make the batch size configurable via the WriterProperties.

src/parquet/column/writer.cc
```diff
@@ -134,52 +141,72 @@ void ColumnWriter::WriteDataPage(const DataPage& page) {
...
+    WriteDictionaryPage();
```
xhochy: Could we extract the following lines into a function `FlushBufferedPages` ?

src/parquet/column/column-writer-test.cc
```diff
@@ -358,13 +364,16 @@ TYPED_TEST(TestPrimitiveWriter, RequiredVeryLargeChunk) {
...
   ASSERT_EQ(this->values_, this->values_out_);
+  std::vector<Encoding::type> encodings = this->metadata_encodings();
+  // There are 3 encodings (RLE, PLAIN_DICTIONARY, PLAIN) in a fallback case
+  // Dictionary encoding is not allowed for boolean type
+  // There are 2 encodings (RLE, PLAIN) in a non dictionary encoding case
+  ASSERT_EQ(Encoding::RLE, encodings[0]);
   if (this->type_num() != Type::BOOLEAN) {
-    // There are 3 encodings (RLE, PLAIN_DICTIONARY, PLAIN) in a fallback case
-    ASSERT_EQ(3, this->metadata_num_encodings());
+    ASSERT_EQ(Encoding::PLAIN_DICTIONARY, encodings[1]);
```
majetideepak: This comment to explicitly verify the encodings did uncover a bug in the metadata writer

src/parquet/file/metadata.cc
```diff
@@ -368,7 +368,7 @@ class ColumnChunkMetaDataBuilder::ColumnChunkMetaDataBuilderImpl {
     column_chunk_->meta_data.__set_total_compressed_size(compressed_size);
     std::vector<format::Encoding::type> thrift_encodings;
     thrift_encodings.push_back(ToThrift(Encoding::RLE));
-    if (properties_->dictionary_enabled(column_->path())) {
+    if (has_dictionary) {
```
majetideepak: The bug is that the column writer will not use dictionary encoding for `Boolean` type. We should depend on the column writer value and not on the default properties value.

src/parquet/column/writer.h
```diff
@@ -56,9 +57,26 @@ class PARQUET_EXPORT ColumnWriter {
 
  protected:
   virtual std::shared_ptr<Buffer> GetValuesBuffer() = 0;
+  /**
+   * Serializes Dictionary Page if enabled
+   */
```
wesm: Comment style is inconsistent, since we have been using `//` we should probably stick to that for now (https://google.github.io/styleguide/cppguide.html#Comment_Style).

wesm:
+1, thank you!

src/parquet/column/writer.h
```diff
@@ -189,6 +208,29 @@ inline void TypedColumnWriter<DType>::WriteBatch(int64_t num_values,
...
+  // The purpose of this chunking is to bound this. Even if a user writes large number
+  // of values, the chunking will ensure the AddDataPage() is called at a reasonable
+  // pagesize limit
+  int64_t write_batch_size = properties_->write_batch_size();
+  int num_batches = num_values / write_batch_size;
+  int64_t num_remaining = num_values % write_batch_size;
+  for (int round = 0; round < num_batches; round++) {
+    int64_t offset = round * write_batch_size;
+    WriteMiniBatch(
+        write_batch_size, &def_levels[offset], &rep_levels[offset], &values[offset]);
```
lomereiter: Too late but: as far as I understand `values` don't include nulls, so this chunking can easily lead to a segfault in case nulls are present in the full batch. (Tried to rebase my PR after this one was merged, got failing tests)

wesm: You're right, sorry I missed that. `WriteMiniBatch` needs to return the value offset. @majetideepak can you open a JIRA? Thanks

majetideepak: I will fix this and add a test case immediately. Sorry for missing this.

majetideepak: JIRA: https://issues.apache.org/jira/browse/PARQUET-719
and 
PR: https://github.com/apache/parquet-cpp/pull/160

PR #159, opened 2016-09-13
PARQUET-718: Fix I/O of non-dictionary encoded pages

xhochy:
We have set dictionary_page_offset if though we did not have a dictionary page present, this caused too much data to be loaded into the SerializedPageReader.


xhochy:
Rebased & fixed the bug.

wesm:
+1. Is it safe to rely on whether the `dictionary_page_offset` is set?

PR #160, opened 2016-09-15
PARQUET-719: Fix WriterBatch API to handle NULL values 

majetideepak:
Fixed bug and added a test case


src/parquet/column/column-writer-test.cc
```diff
@@ -150,6 +150,8 @@ typedef ::testing::Types<Int32Type, Int64Type, Int96Type, FloatType, DoubleType,
 
 TYPED_TEST_CASE(TestPrimitiveWriter, TestTypes);
 
+typedef TestPrimitiveWriter<Int32Type> TestNullValuesWriter;
```
xhochy: Please use using.

majetideepak: Done.

wesm:
+1

PR #161, opened 2016-09-15
PARQUET-720: Mark ScanAllValues as inline to prevent link error

flode:
When reader.h is included in multiple cpp files, it causes a linker error
because ScanAllValues is defined in multiple TUs

An alternative solution would be to move ScanAllValues to a separate cpp file.

Should I open a JIRA for this?


wesm:
Yes, please move the function to a `.cc` file. Can you create a JIRA? Thanks!

xhochy:
@flode Can you fix the line lengths reported in Travis? Then this is ready to merge.

PR #162, opened 2016-09-15
PARQUET-689: C++: Compress DataPages eagerly

majetideepak:


src/parquet/column/page.h
```diff
@@ -95,6 +95,21 @@ class DataPage : public Page {
   std::string min_;
 };
 
+class CompressedDataPage : public DataPage {
+ public:
+  CompressedDataPage(const std::shared_ptr<Buffer>& buffer, int32_t num_values,
```
xhochy: I was a bit confused that we do not have to have the compression codec here. For the moment it is ok, as it will not be used. Once we use the CompressedDataPage for reading, it might be worth it (but it seems that there is not that much freedom in having different compression codec inside a single column chunk.)

majetideepak: I had the exact line of thought. May be we can add a comment indicating that the codec information is stored in the parent `ColumnReader/ ColumnWriter`

PR #163, opened 2016-09-18
PARQUET-720: Mark ScanAllValues as inline to prevent link error

xhochy:
Closes #161

PR #164, opened 2016-09-18
PARQUET-720: Compile scan-all.cc to prevent linker errors

wesm:
Closes #161. There was a missing namespace from `scan-all.h` and a code formatting issue.


wesm:
Closing in favor of #163

PR #167, opened 2016-09-23
PARQUET-728: Incorporate upstream Arrow API changes

wesm:
I bumped the thirdparty git hash to Arrow trunk. Pardon the diff noise from clang format. Also resolves PARQUET-729.


wesm:
By the way, it seems the LLVM apt repo is back: http://apt.llvm.org/

src/parquet/file/writer-internal.cc
```diff
@@ -142,7 +142,7 @@ ColumnWriter* RowGroupSerializer::NextColumn() {
   // Throws an error if more columns are being written
   auto col_meta = metadata_->NextColumnChunk();
 
-  if (current_column_writer_) { total_bytes_written_ = current_column_writer_->Close(); }
+  if (current_column_writer_) { total_bytes_written_ += current_column_writer_->Close(); }
```
majetideepak: Did this below check fail for multiple columns ? 
https://github.com/apache/parquet-cpp/commit/55604b297f444e95e132658b2ef384870ae1f701#diff-553ab87bc327b48a8dcd0b499b521737R470

IIRC we are not using this value to set the rowgroup size.

majetideepak:
+1 LGTM

src/parquet/file/writer-internal.cc
wesm: The `total_bytes_written_` is what is being used as the `total_byte_size` in the row group metadata. If you revert the `+=` to `=` the DCHECK (since the test suite now writes multi-column tables) fails

majetideepak: The fix you made here is correct. However, I don't think we are using `total_bytes_written_` to set the `total_byte_size` value in the row group metadata. We are re-computing it in the metadata writer over all the columns written.
https://github.com/apache/parquet-cpp/blob/master/src/parquet/file/metadata.cc#L471
 `total_byte_size += row_group_->columns[i].meta_data.total_compressed_size;`
 `................. `
`row_group_->__set_total_byte_size(total_byte_size);`

wesm: The fact that we are computing the total bytes written twice is a bad code smell to me -- that this bug was only caught because of the DCHECK in `RowGroupMetaDataBuilder::Finish` tells me that the unit tests elsewhere are inadequate. I would prefer to take out the the double-accounting (save these checks for for the unit test suite) and use the `total_bytes_written` passed as an argument to `Finish` but not until the tests are improved.

majetideepak: The tests you updated with multiple columns should be sufficient. However, we should keep this check until we feel confident. JIRA to remove the redundancy PARQUET-730 later.

PR #169, opened 2016-09-26
PARQUET-731: API to return metadata size and Skip reading values

majetideepak:


src/parquet/column/reader.h
```diff
@@ -124,8 +125,12 @@ class PARQUET_EXPORT TypedColumnReader : public ColumnReader {
   // This API is the same for both V1 and V2 of the DataPage
   //
   // @returns: actual number of levels read (see values_read for number of values read)
-  int64_t ReadBatch(int32_t batch_size, int16_t* def_levels, int16_t* rep_levels,
-      T* values, int64_t* values_read);
+  int64_t ReadBatch(int batch_size, int16_t* def_levels, int16_t* rep_levels, T* values,
+      int64_t* values_read);
```
majetideepak: Change to match the definition

src/parquet/column/reader.h
```diff
@@ -166,7 +171,7 @@ inline int64_t TypedColumnReader<DType>::ReadBatch(int batch_size, int16_t* def_
 
   // TODO(wesm): keep reading data pages until batch_size is reached, or the
   // row group is finished
-  batch_size = std::min(batch_size, num_buffered_values_);
+  batch_size = std::min(batch_size, num_buffered_values_ - num_decoded_values_);
```
majetideepak: This is currently not a bug since the underlying layers  `ReadDef/RepLevels` and `ReadValues` handle cases where the requesting `batch_size` > remaining values in the current page. However, this fix corrects the logic here.

src/parquet/column/column-reader-test.cc
```diff
@@ -151,6 +151,64 @@ TEST_F(TestPrimitiveReader, TestInt32FlatRepeated) {
...
+  vector<int32_t> sub_values(
+      values_.begin() + 2 * levels_per_page, values_.begin() + 2.5 * levels_per_page);
+  ASSERT_TRUE(vector_equal(sub_values, vresult));
+
+  // 2) skip_size == page_size (skip across two pages)
+  levels_skipped = reader->Skip(levels_per_page);
+  ASSERT_EQ(levels_per_page, levels_skipped);
+  // Read half a page
+  reader->ReadBatch(
+      levels_per_page / 2, &dresult[0], &rresult[0], &vresult[0], &values_read);
```
xhochy: I would prefer if you used `.data()` for clarity on the `vector` pointers.

src/parquet/column/reader.h
```diff
@@ -201,6 +206,33 @@ inline int64_t TypedColumnReader<DType>::ReadBatch(int batch_size, int16_t* def_
...
+    // If the number of rows to skip is more than the number of undecoded values, skip the
+    // Page.
+    if (rows_to_skip > (num_buffered_values_ - num_decoded_values_)) {
+      rows_to_skip -= num_buffered_values_ - num_decoded_values_;
+      num_decoded_values_ = num_buffered_values_;
+    } else {
+      // We need to read this Page
+      // Jump to the right offset in the Page
+      int64_t values_read = 0;
+      std::vector<uint8_t> vals(
```
xhochy: Not so happy that we allocate here memory, especially as it's memory that cannot be tracked by the allocator. We should rather have a variant of `ReadBatch` that reads to "/dev/null".

src/parquet/column/column-reader-test.cc
majetideepak: Will modify this.

src/parquet/column/reader.h
majetideepak: Will try to implement this

majetideepak: The main difficulty here is that we need to count the number of `values` in a  `DATA_PAGE` using the `definition levels`. 
I modified the patch by using memory via the allocator and reduced the footprint via chunking.
Another optimization which is probably not worth the effort is to implement a `Skip` inside the decoders. This will help avoid reading the repetition levels and values after reading the definition levels.

majetideepak:
@xhochy do you have more feedback on this ? Thanks.

wesm:
+1

PR #172, opened 2016-10-05
PARQUET-738: Update arrow version that also supports newer Xcode

xhochy:


xhochy:
@wesm I'm asthonished that this doesn't break. Shouldn't your recent I/O changes in Arrow change some interfaces?

wesm:
It looks like not, because I did not remove any public APIs from `ReadableFileInterface` -- I had thought we were *implementing* the interfaces in Parquet but it's the other way around, my mistake.

wesm:
The change I made was to add an abstract `Read(int64_t nbytes, shared_ptr<Buffer>* out)` method and add default implementations of `ReadAt` (rather than having them as pure virtual)

wesm:
+1

PR #173, opened 2016-10-06
PARQUET-741: Always allocate fresh buffers while compressing

xhochy:
Still fails for FLBAType, need to adjust the unit tests for that.


src/parquet/column/column-writer-test.cc
```diff
@@ -154,8 +166,54 @@ class TestPrimitiveWriter : public PrimitiveTypedTest<TestType> {
...
+    auto data = std::make_shared<std::vector<uint8_t>>(
+        values_read_recently * this->descr_->type_length());
+    uint8_t* data_ptr = data->data();
+    for (int64_t i = 0; i < values_read_recently; i++) {
+      memcpy(data_ptr + this->descr_->type_length() * i,
+          this->values_out_[i + values_read_].ptr, this->descr_->type_length());
+      this->values_out_[i + values_read_].ptr =
+          data_ptr + this->descr_->type_length() * i;
+    }
+    data_buffer_.push_back(data);
```
wesm: Use emplace_back to move instead of copying, then you can avoid the `shared_ptr`

xhochy: Ok....forgot a call to `std::move`, sorry.

PR #174, opened 2016-10-06
PARQUET-742: Add missing license headers

xhochy:
- Removed `TODO` as it is totally outdated


wesm:
@xhochy it's fine to put a normal license header on it. it needs to be referenced in the NOTICE file

wesm:
If you put a note in NOTICE (which we need to release anyway), I will merge this

wesm:
+1

PR #175, opened 2016-10-07
PARQUET-739: Don't use a static buffer for data accessed by multiple threads

flode:
This buffer is used in multiple threads, so it cannot be static. It's small enough to just have it on the stack. I could add a vector to the RleDecoder class as well if you prefer


xhochy:
Can you change the title to include `:` after PARQUET-739?

PR #176, opened 2016-10-07
PARQUET-745: TypedRowGroupStatistics fails to PlainDecode min and max in ByteArrayType

flode:
I'm not sure how this code is supposed to work, this seems to solve the issue.

The code was added in 176b08c305919551ecfd5fc2f741f7bff4deefdd by @lomereiter I think.

### DIFF
--- a/src/parquet/column/statistics.cc
+++ b/src/parquet/column/statistics.cc
@@ -162,13 +162,13 @@ void TypedRowGroupStatistics<DType>::PlainDecode(const std::string& src, T* dst)
 
 template <>
 void TypedRowGroupStatistics<ByteArrayType>::PlainEncode(
-   const T& src, std::string* dst) {
+     const T& src, std::string* dst) {
   dst->assign(reinterpret_cast<const char*>(src.ptr), src.len);
 }
 
 template <>
 void TypedRowGroupStatistics<ByteArrayType>::PlainDecode(
-   const std::string& src, T* dst) {
+     const std::string& src, T* dst) {
   dst->len = src.size();
   dst->ptr = reinterpret_cast<const uint8_t*>(src.c_str());
 }
--- a/src/parquet/column/statistics.cc
+++ b/src/parquet/column/statistics.cc
@@ -64,27 +64,7 @@ TypedRowGroupStatistics<DType>::TypedRowGroupStatistics(const ColumnDescriptor* 
 
   if (!encoded_min.empty()) { PlainDecode(encoded_min, &min_); }
   if (!encoded_max.empty()) { PlainDecode(encoded_max, &max_); }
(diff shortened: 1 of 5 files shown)


lomereiter:
Thanks for pointing this out, I checked `parquet-mr` code and it's indeed the case that binary min/max are [expected to be non-encoded](https://github.com/apache/parquet-mr/blob/e54ca615f213f5db6d34d9163c97eec98920d7a7/parquet-column/src/test/java/org/apache/parquet/column/statistics/TestStatistics.java#L376-L380). This is not touched in `parquet-format` anyhow so perhaps file a bug there as well?

Tests are going to be failing until the write path is changed as well. A cleaner way would be to put logic into `PlainEncode`/`PlainDecode` methods instead of the constructor.

flode:
I don't think we need to specialize for fixed length array and int96. The problem with the variable byte array is that there is no length prefix in min/max. For fixed and into96 just copying the bytes should work.

src/parquet/column/statistics.cc
```diff
@@ -64,27 +64,7 @@ TypedRowGroupStatistics<DType>::TypedRowGroupStatistics(const ColumnDescriptor*
...
-
-  SetDescr(schema);
-
-  min_.len = encoded_min.size();
-  min_.ptr = reinterpret_cast<const uint8_t*>(encoded_min.c_str());
-  max_.len = encoded_max.size();
-  max_.ptr = reinterpret_cast<const uint8_t*>(encoded_max.c_str());
-
-  has_min_max_ = !encoded_min.empty() && !encoded_max.empty();
+  has_min_max_ = !encoded_min.empty() || !encoded_max.empty();
```
lomereiter: Good point. Feel free to change this constructor to take `const EncodedStatistics&` instead, as it's used only in `MakeTypedColumnStats` in `metadata.cc`.

flode: I found it in two places, reader-internal.cc and metadata.cc. I also found a difference in the two calls:
In metadata.cc `metadata.num_values - metadata.statistics.null_count`
In reader-internal.cc `metadata.num_values`
Is that different on purpose?

As `metadata.statistics` has type `Statistics`, it's better to use this than converting to EncodedStatistics, right?

lomereiter: The call in `reader-internal.cc`is actually unused, I'd put it there before Metadata class appeared, it can be safely removed now. Regarding `num_values`, see #145.

`Statistics` should be converted to `EncodedStatistics` because there's a policy not to expose Thrift-generated classes to the public interface.

wesm:
How hard would it be to add a unit test for this?

wesm:
Looks like this may be failing only from cpplint

flode:
Checks are all green now

wesm:
+1, thank you

PR #177, opened 2016-10-10
PARQUET-739: Don't use a static buffer for data accessed by multiple threads

majetideepak:
I am hitting this problem at another location too.


wesm:
+1

wesm:
Would be nice to test this kind of stuff at some point, but may be difficult to make test failures reproducible

majetideepak:
Yeah! Reproducibility is an issue. This bug surfaced when loading lots of data.
We could use static code analyzers that might help us uncover such bugs.

flode:
Running helgrind in some automated way could help, it detects when there's a data race problem. That makes reproducing easier as it doesn't need to crash to be detected

PR #178, opened 2016-10-11
PARQUET-747: Better hide TypedRowGroupStatistics in public API

wesm:
EDIT: It appears the linker issue is coming from the use of `TypedRowGroupStatistics<T>::Update` in `WriteBatch`, and Update hasn't been exported. There's some gcc quirks around this (see https://codereview.chromium.org/643703003/), so it's easier to hide the WriteBatch implementation so that `libparquet_arrow` doesn't look for non-exported symbols during linking (since `WriteBatch` was being inlined). 

Template subclasses don't behave in the same way with visibility attributes. See also PARQUET-659. Observed in build failure https://circleci.com/gh/conda-forge/staged-recipes/10100


wesm:
Note: apparently the linking issue only showed up in conda-forge because it's using the Redhat devtoolset, which I guess has stricter linker behavior, since this hadn't been showing up in local gcc 4.9 builds

xhochy:
+1, also showed up for my while doing `manylinux1` builds, probably the same toolset.

wesm:
Argh, now I'm running into this issue: https://llvm.org/bugs/show_bug.cgi?id=24815

wesm:
This provides some clue about how to fix, but I'm a little bit stumped for the moment: https://codereview.chromium.org/643703003/

wesm:
Looking more closely -- I don't think we should be exporting `TypedRowGroupStatistics` in the public API at all, so we need to refactor the headers a bit

wesm:
I'm going to merge this and see if that resolves the issue, since it's very difficult for me to reproduce the conda-forge link failure

PR #179, opened 2016-10-11
PARQUET-744: Corrected build instructions for dependency requirements

chrisaycock:


README.md
```diff
@@ -73,13 +80,15 @@ export GTEST_HOME=`pwd`/thirdparty/$GTEST_BASEDIR
...
     - THRIFT_HOME customizes the thrift installed location.
     - SNAPPY_HOME customizes the snappy installed location.
-    - LZ4_HOME customizes the lz4 installed location.
 
 - `make`
 
 The binaries will be built to ./debug which contains the libraries to link against as
 well as a few example executables.
 
+For release-level builds (enable optimizations and diable debugging), pass
```
wesm: disable

README.md
```diff
@@ -20,17 +20,18 @@
...
 - zlib
 - thrift 0.7+ [install instructions](https://thrift.apache.org/docs/install/)
 - googletest 1.7.0 (cannot be installed with package managers)
 - Google Benchmark (only required if building benchmarks)
 
 You can install these dependencies using a package manager or using the
-`thirdparty/` scripts in this repository. On Homebrew, you can run:
+`thirdparty/` scripts in this repository.
+
+macOS already has zlib; on Homebrew, you can run:
```
wesm: Did this change between 10.11 and 10.12?

README.md
```diff
@@ -73,13 +80,15 @@ export GTEST_HOME=`pwd`/thirdparty/$GTEST_BASEDIR
   - You can customize dependent library locations through various environment variables:
     - THRIFT_HOME customizes the thrift installed location.
     - SNAPPY_HOME customizes the snappy installed location.
-    - LZ4_HOME customizes the lz4 installed location.
```
wesm: Add `ZLIB_HOME` here

README.md
```diff
@@ -42,10 +43,16 @@ source thirdparty/set_thirdparty_env.sh
...
-setup_build_env.sh`.  By default, it will create a build directory `build/`.
-You can override the build directory by setting the BUILD_DIR env variable to
-another location.
+with third party dependencies.  You use it by running
+
+```shell
+source setup_build_env.sh
+```
+
+(You can't build Thrift on macOS with this script! Use Homebrew.)
```
wesm: Maybe this merits 2 sections (, 1 for Linux and 1 for OS X) with a single set of shell commands to copy and paste for each? May slightly improve the bounce rate on following the instructions

chrisaycock: I would also prefer to move the single-line Linux version above the three-line sample of what to run. I agree that it's a lot easier to read, "paste this for Mac and paste this for Linux".

README.md
wesm: Fine as is, with the split sections per below I think it will be easy enough to follow

README.md
```diff
@@ -20,40 +20,43 @@
 ## Third Party Dependencies
 
 - snappy
-- lz4
 - zlib
 - thrift 0.7+ [install instructions](https://thrift.apache.org/docs/install/)
 - googletest 1.7.0 (cannot be installed with package managers)
```
majetideepak: Should we provide a link to SO: http://stackoverflow.com/questions/13513905/how-to-setup-googletest-as-a-shared-library-on-linux ?

README.md
```diff
@@ -20,40 +20,43 @@
...
 ```shell
- brew install snappy lz4 thrift zlib
+source setup_build_env.sh
 ```
 
-To build the thirdparty libraries in-tree, run:
+By default, it will create a build directory `build/`. You can override the
+build directory by setting the BUILD_DIR env variable to another location.
+
+To manually build the thirdparty libraries in-tree, run:
```
majetideepak: Do you want to mention that we can choose specific dependencies to be installed ?
`./thirdparty/build_thirdparty.sh zlib`
options: arrow, zlib, gbenchmark, gtest, snappy, thrift

chrisaycock:
Should we just rewrite the README? I've taken an incremental approach to just fix the incorrect items in there. Full rewrite is a bigger undertaking and should be coordinated with the existing developers.

majetideepak:
I am okay with minimal changes in this patch.
Let's open a separate JIRA for README rewrite. 
I feel the README can be made better with a table of contents and including other details like in my comments above.

wesm:
I'm going to merge this as is -- we should do the same thing as ARROW-189 and get rid of the download_thirdparty / build_thirdparty scripts then re-evaluate the README details

wesm:
+1

PR #180, opened 2016-10-17
PARQUET-752: Account for upstream Arrow API changes

wesm:
As soon as ARROW-261/317 are merged, I'll update the thirdparty git SHA so we can get a green build


src/parquet/arrow/writer.cc
```diff
@@ -45,6 +45,8 @@ using parquet::schema::GroupNode;
 namespace parquet {
 namespace arrow {
 
+namespace BitUtil = ::arrow::BitUtil;
```
xhochy: Didn't know this would work. Why not use `using` here?

wesm: Using doesn't work with namespaces AFAICT

wesm:
Updated the thirdparty SHA. Guess I may have to wait til morning for the OS X builds

wesm:
The externalproject changes in Arrow broke the build, looking into it

wesm:
+1

PR #183, opened 2016-11-03
PARQUET-762: C++: Use optimistic allocation instead of Arrow Builders

xhochy:
About 2x as fast on non-string data.


wesm:
+1. We may also want to factor out the bitmap allocation (and memset to all zeros) into a utility method in libarrow

PR #184, opened 2016-11-03
PARQUET-763: C++: Expose ParquetFileReader through Arrow reader

xhochy:


src/parquet/arrow/reader.h
```diff
@@ -100,6 +100,8 @@ class PARQUET_EXPORT FileReader {
   // Read a table of flat columns into a Table.
   ::arrow::Status ReadFlatTable(std::shared_ptr<::arrow::Table>* out);
 
+  const ParquetFileReader* parquet_reader();
```
wesm: This can be `parquet_reader() const`

wesm:
+1 -- note that Parquet is R-T-C (http://apache.org/foundation/glossary.html#ReviewThenCommit, Arrow is still commit-then-review for the time being)

PR #185, opened 2016-11-06
PARQUET-764: Support batches for PLAIN boolean writes that aren't a multiple of 8

xhochy:
cc @majetideepak

### DIFF
--- a/src/parquet/encodings/plain-encoding.h
+++ b/src/parquet/encodings/plain-encoding.h
@@ -228,8 +228,7 @@ class PlainEncoder<BooleanType> : public Encoder<BooleanType> {
       bit_writer_.reset(                                                          \
           new BitWriter(bits_buffer_->mutable_data(), bits_buffer_->size()));     \
                                                                                   \
-      /* As we have allocated a large enough buffer, we can write                 \
-         all values directly. */                                                  \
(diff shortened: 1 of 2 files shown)


src/parquet/encodings/plain-encoding.h
```diff
@@ -181,59 +181,73 @@ class PlainEncoder<BooleanType> : public Encoder<BooleanType> {
...
+        values_sink_->Write(bit_writer_->buffer(), bit_writer_->bytes_written()); \
+      }                                                                           \
+    }                                                                             \
+                                                                                  \
+    if (bit_offset < num_values) {                                                \
+      int bits_to_write = num_values - bit_offset;                                \
+      int bytes_required = BitUtil::Ceil(bits_to_write, 8);                       \
+      if (bytes_required > bits_buffer_->size()) {                                \
+        bits_buffer_->Resize(bytes_required);                                     \
+      }                                                                           \
```
majetideepak: instead of `resizing`, could we have a loop here that Flushes values at every `IN_MEMORY_DEFAULT_CAPACITY` ?

xhochy: Removed the resizing and also removed the unique_ptr around the `bits_buffer_`.

wesm:
+1

PR #187, opened 2016-11-07
PARQUET-702: Add a writer + reader example with detailed comments

majetideepak:
1. Added a reader writer example.
    -  Covered all the physical types supported by Parquet.


examples/reader-writer.cc
```diff
@@ -0,0 +1,426 @@
+// Licensed to the Apache Software Foundation (ASF) under one
```
xhochy: I would prefer to have this file clang-formatted. I would find it much better readable with consistent linebreaks and indentations as the other code.

examples/reader-writer.cc
```diff
@@ -0,0 +1,426 @@
...
+constexpr int NUM_ROW_GROUPS = 10;
+constexpr int NUM_ROWS_PER_ROW_GROUP = 512;
+constexpr int NUM_COLUMNS = 8;
+constexpr int FIXED_LENGTH = 10;
+const std::string PARQUET_FILENAME = "parquet_cpp_example.parquet";
+
+class ParquetWriter {
+ public:
+  ParquetWriter() {
+      ostream_.reset(new parquet::InMemoryOutputStream());
```
xhochy: Why do we first write into memory and not directly to a file? If there is some reason, please state it here.

examples/reader-writer.cc
```diff
@@ -0,0 +1,426 @@
...
+  }
+
+  void SetupSchema() {
+    parquet::schema::NodeVector fields;
+    // Create a primitive node named 'boolean_field' with type:BOOLEAN, repetition:REQUIRED
+    fields.push_back(parquet::schema::PrimitiveNode::Make("boolean_field", parquet::Repetition::REQUIRED, parquet::Type::BOOLEAN, parquet::LogicalType::NONE));
```
xhochy: At first glance, I'm ok with with not doing `using parquet::*` but this line gets really long because of that. I would prefer to have less `parquet::` prefixes here.

examples/reader-writer.cc
```diff
@@ -0,0 +1,426 @@
...
+    // Create a primitive node named 'int32_field' with type:INT32, repetition:REQUIRED, logical type:TIME_MILLIS
```
xhochy: The comments are very repetitive, either leave them away after the first or add explanations to those that introduce different repetition types.

examples/reader-writer.cc
```diff
@@ -0,0 +1,426 @@
...
+  parquet::ParquetFileWriter* GetWriter() {
+    if (!writer_) {
+      // Create a parquet file writer using the schema and the output stream.
+      writer_ = parquet::ParquetFileWriter::Open(ostream_, schema_);
+    }
+    return writer_.get();
+  }
+
+  void WriteToFile(std::string filename) {
+    std::ofstream parquetFile;
```
xhochy: With writing the file at the end, it builds up the impression that you need to serialise everything in-memory. As mentioned earlier either describe why you do it or add a comment here stating it a bit more or just use the file implementation in Parquet.

examples/reader-writer.cc
```diff
@@ -0,0 +1,426 @@
...
+    parquet_writer->SetupSchema();
+
+    // Get a handle to the ParquetFileWriter
+    parquet::ParquetFileWriter* file_writer = parquet_writer->GetWriter();
+
+    // Append a RowGroup with a specific number of rows.
+    parquet::RowGroupWriter* rgWriter = file_writer->AppendRowGroup(NUM_ROWS_PER_ROW_GROUP);
+
+    // Write the Bool column 
+    parquet::ColumnWriter* colWriter = rgWriter->NextColumn();
```
xhochy: I would prefer if we describe the `TypedColumnWriter` API instead of introducing a new API just for the examples.

examples/reader-writer.cc
```diff
@@ -0,0 +1,426 @@
...
+/* Parquet is a structured columnar file format
+ * Parquet File = "Parquet data" + "Parquet Metadata"
+ * "Parquet data" is simply a vector of RowGroups. Each RowGroup is a batch of rows in a columnar layout
+ * "Parquet Metadata" contains the "file schema" and attributes of the RowGroups and their Columns
+ * "file schema" is a tree where each node is either a primitive type (leaf nodes) or a complex (nested) type (internal nodes)
+ * For specific details, please refer the format here:
+ * https://github.com/apache/parquet-format/blob/master/LogicalTypes.md
+**/
+
+constexpr int NUM_ROW_GROUPS = 10;
```
xhochy: Not used.

examples/reader-writer.cc
```diff
@@ -0,0 +1,426 @@
...
+
+        int64_t values_read = 0;
+        int64_t rows_read = 0;
+        int16_t definition_level;
+        int16_t repetition_level;
+        int i;
+        std::shared_ptr<parquet::ColumnReader> column_reader;
+
+        // Get the Column Reader for the boolean column
+        column_reader = row_group_reader->Column(0);
```
xhochy: I would also use the `TypedColumnReader` here.

examples/reader-writer.cc
majetideepak: Will write to file directly.

examples/reader-writer.cc
xhochy: Yes. I would like the user to cast to them to actually have typed `Write` calls. In the end you either have to cast your data pointer with `reinterpret_cast<uint8_t*>` or the `ColumnWriter` to a `TypedColumnWriter`.

Using the `WriteAllValues` will also prohibit some future features like "I want to write an uint8_t to an int32_t physical column`". We can support this via differently typed `WriteBatch` call, e.g. we do this already for `const bool*` and `const std::vector<bool>&` for Boolean columns.

majetideepak: Will make this change

examples/reader-writer.cc
```diff
@@ -0,0 +1,481 @@
...
+  // parquet::REPEATED fields require both definition and repetition level values
+  try {
+    // Create a ParquetWriter instance
+    auto parquet_writer = std::unique_ptr<ParquetWriter>(new ParquetWriter());
+
+    // Setup the schema
+    parquet_writer->SetupSchema();
+
+    // Get a handle to the ParquetFileWriter
+    parquet::ParquetFileWriter* file_writer = parquet_writer->GetWriter();
```
wesm: It may be slightly pedantic, but I don't think there's a need for the ParquetWriter class here. Instead:

```c++
std::shared_ptr<OutputStream> out_file = std::make_shared<LocalFileOutputStream>(PARQUET_FILENAME);
std::shared_ptr<schema::GroupNode>> schema = SetupSchema();  // static method
std::shared_ptr<ParquetFileWriter> file_writer = ParquetFileWriter::Open(out_file, schema);
```

examples/reader-writer.cc
```diff
@@ -0,0 +1,481 @@
...
+        file_writer->AppendRowGroup(NUM_ROWS_PER_ROW_GROUP);
+
+    // Write the Bool column
+    parquet::BoolWriter* bool_writer =
+        static_cast<parquet::BoolWriter*>(rg_writer->NextColumn());
+    for (int i = 0; i < NUM_ROWS_PER_ROW_GROUP; i++) {
+      bool value = ((i % 2) == 0) ? true : false;
+      bool_writer->WriteBatch(1, nullptr, nullptr, &value);
+    }
+    bool_writer->Close();
```
wesm: Is it necessary to call `Close`?

examples/reader-writer.cc
majetideepak: Both valid points. Fixing them.

wesm:
+1 -- will merge on green build

PR #190, opened 2016-11-09
PARQUET-775: Make TrackingAllocator thread-safe

xhochy:
Mutex is only locked around the statistics as the other parts are
thread-safe and should run in parallel if needed/possible.


majetideepak:
+1 LGTM

wesm:
+1

PR #191, opened 2016-11-11
Standardize the schema output to match the parquet-mr format

trink:
- root node name is preceded by 'message'
- byte_array type is named 'binary'
- column entries end with a semicolon
- add logical type output


trink:
An update to fix the tests will be posted shortly

PR #193, opened 2016-11-17
PARQUET-779: Export TypedRowGroupStatistics in libparquet

lomereiter:
Workaround is necessary because of GCC bug 40068/50044.
`-Wattributes` diagnostic is temporarily disabled, as it seems the easiest solution.

### DIFF
--- a/src/parquet/column/statistics.h
+++ b/src/parquet/column/statistics.h
@@ -211,7 +211,7 @@ typedef TypedRowGroupStatistics<DoubleType> DoubleStatistics;
 typedef TypedRowGroupStatistics<ByteArrayType> ByteArrayStatistics;
 typedef TypedRowGroupStatistics<FLBAType> FLBAStatistics;
 
-#ifdef __GNUC__
+#if defined(__GNUC__) && !defined(__clang__)
 #pragma GCC diagnostic push
 #pragma GCC diagnostic ignored "-Wattributes"
 #endif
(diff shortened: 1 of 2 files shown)


src/parquet/column/statistics.h
```diff
@@ -202,23 +202,32 @@ void TypedRowGroupStatistics<ByteArrayType>::PlainEncode(const T& src, std::stri
...
-extern template class TypedRowGroupStatistics<ByteArrayType>;
-extern template class TypedRowGroupStatistics<FLBAType>;
+typedef TypedRowGroupStatistics<BooleanType> BoolStatistics;
+typedef TypedRowGroupStatistics<Int32Type> Int32Statistics;
+typedef TypedRowGroupStatistics<Int64Type> Int64Statistics;
+typedef TypedRowGroupStatistics<Int96Type> Int96Statistics;
+typedef TypedRowGroupStatistics<FloatType> FloatStatistics;
+typedef TypedRowGroupStatistics<DoubleType> DoubleStatistics;
+typedef TypedRowGroupStatistics<ByteArrayType> ByteArrayStatistics;
+typedef TypedRowGroupStatistics<FLBAType> FLBAStatistics;
```
wesm: Is there a reason to use typedef over using?

lomereiter: I was driven simply by wanting consistency with the rest of the codebase (column/reader.h, column/writer.h). I see however that the tradition is different in tests and `arrow-cpp`

lomereiter: I guess we could add this to `clang-tidy` config: http://clang.llvm.org/extra/clang-tidy/checks/modernize-use-using.html

wesm:
+1

PR #195, opened 2016-11-23
PARQUET-780: WriterBatch API does not properly handle NULL values for byte array types

majetideepak:
CC: @xhochy @trink


src/parquet/column/writer.cc
```diff
@@ -336,6 +336,9 @@ inline int64_t TypedColumnWriter<DType>::WriteMiniBatch(int64_t num_values,
     throw ParquetException("More rows were written in the column chunk than expected");
   }
 
+  // PARQUET-780
+  if (values_to_write > 0) { DCHECK(NULL != values) << "Values ptr cannot be NULL"; }
```
majetideepak: The `if conditions` will also be optimized out in a Release mode.

PR #196, opened 2016-11-25
PARQUET-782: Support writing to Arrow sinks

xhochy:


xhochy:
Depends on https://github.com/apache/arrow/pull/215

src/parquet/arrow/reader.cc
```diff
@@ -376,6 +376,79 @@ Status FlatColumnReader::Impl::TypedReadBatch(
 }
 
 template <>
+Status FlatColumnReader::Impl::TypedReadBatch<::arrow::BooleanType, BooleanType>(
+    int batch_size, std::shared_ptr<Array>* out) {
```
wesm: This is all because the `c_type` trait disappeared, or something else?

xhochy: Yes, that are all needed changes.

wesm:
+1, thank you

PR #198, opened 2016-12-01
PARQUET-785: LIST schema conversion for Arrow lists

xhochy:


src/parquet/arrow/schema.cc
```diff
@@ -220,6 +277,19 @@ Status FromParquetSchema(
   return Status::OK();
 }
 
+Status ListToNode(const std::shared_ptr<::arrow::ListType>& type, const std::string& name,
+    bool nullable, const WriterProperties& properties, NodePtr* out) {
+  Repetition::type repetition = Repetition::REQUIRED;
+  if (nullable) { repetition = Repetition::OPTIONAL; }
```
wesm: Use a ternary here? (`nullable ? Repetition::OPTIONAL : Repetition::REQUIRED`)

src/parquet/arrow/arrow-schema-test.cc
```diff
@@ -157,6 +157,178 @@ TEST_F(TestConvertParquetSchema, ParquetFlatDecimals) {
...
+        GroupNode::Make("my_list", Repetition::OPTIONAL, {list}, LogicalType::LIST));
+    auto arrow_element = std::make_shared<Field>("str", UTF8, false);
+    auto arrow_list = std::make_shared<::arrow::ListType>(arrow_element);
+    arrow_fields.push_back(std::make_shared<Field>("my_list", arrow_list, true));
+  }
+
+  // // List<Integer> (nullable list, non-null elements)
+  // optional group my_list (LIST) {
+  //   repeated int32 element;
+  // }
```
wesm: Here's the two-level encoding. What about one-level? https://github.com/apache/parquet-cpp/blob/master/src/parquet/schema/types.h#L43

To be fully rigorous you might want to test a required group.

wesm: My understanding that everything but the 3-level encoding is deprecated, so this may not be terribly important

PR #199, opened 2016-12-03
PARQUET-767: Add release scripts

xhochy:
Based on the release scripts from Apache Aurora (as we don't have the luxury of the Maven release plugin, we need a bit more code). Kudos to @StephanErb for pointing me to them.

- [x] Rename JIRA release tag from `cpp-0.1` to `cpp-0.1.0` ( @julienledem can you take care of this?) This is then in line with all other Apache Parquet artifacts.
- [x] Test the scripts with dry-run mode (needs the above point)
- [x] Reference Apache Aurora in LICENSE/NOTICE
- [x] Add Guides
- [x] Add script to verify the release


wesm:
@xhochy I just renamed the JIRA release tag. We still need to reference Impala in the LICENSE

xhochy:
As a sidenote: The fixed Travis image will be released somewhere in the next ~12h, then we are hopefully back to green Parquet builds.

StephanErb:
I would recommend to also include a short documentation how to use the scripts. Here are the docs we have been using so far: https://github.com/apache/aurora/blob/master/docs/development/committers-guide.md#creating-a-release

xhochy:
That's the TODO `Add Guides` above ;)

wesm:
+1. It's kind of an unbelievable amount of code

PR #201, opened 2016-12-05
PARQUET-789: Catch/translate ParquetExceptions in parquet::arrow::FileReader

wesm:
I ran into this issue when debugging ARROW-406. I also changed the translated status type to `IOError` if that makes sense.


wesm:
+1

PR #203, opened 2016-12-15
PARQUET-799: Fix bug in MemoryMapSource::CloseFile

wdforson:
The key change here is to ensure that munmap is called at most once for
a given memory-mapped file.

Previously, MemoryMapSource::CloseFile was calling munmap on every
invocation (provided that a file had ever been successfully mapped into
memory for that instance). This is problematic in a multi-threaded
environment, even if each MemoryMapSource instance is being used in only
one thread, as illustrated by the following hypothetical sequence of
operations:

thread 1 @ time 1: munmap(0xf000, 4096)
thread 2 @ time 2: void *addr = mmap(NULL, 4096, ...) // addr <- 0xf000
thread 1 @ time 3: munmap(0xf000, 4096)

After time 3, the mapping for the memory segment beginning at 0xf000 has
been invalidated, so the next attempt by thread 2 to access memory
within that segment will likely cause a segfault (unless yet another
thread has mmap'd that segment in the meantime, in which case the
results could be even more interesting, but certainly no better).

Also, I'm adding/modifying a couple comments in header files to mark
"sample" implementations accordingly. This is intended to give API
consumers a heads up as to the intent and level of maturity of those
sections of the codebase.

### DIFF
--- a/src/parquet/file/reader.h
+++ b/src/parquet/file/reader.h
@@ -79,7 +79,8 @@ class PARQUET_EXPORT ParquetFileReader {
   ParquetFileReader();
   ~ParquetFileReader();
 
-  // API Convenience to open a serialized Parquet file on disk
+  // API Convenience to open a serialized Parquet file on disk, using SAMPLE
+  // implementations of RandomAccessSource
   static std::unique_ptr<ParquetFileReader> OpenFile(const std::string& path,
       bool memory_map = true, ReaderProperties props = default_reader_properties());
 
--- a/src/parquet/file/reader.h
+++ b/src/parquet/file/reader.h
@@ -79,8 +79,9 @@ class PARQUET_EXPORT ParquetFileReader {
   ParquetFileReader();
   ~ParquetFileReader();
 
-  // API Convenience to open a serialized Parquet file on disk, using SAMPLE
-  // implementations of RandomAccessSource
+  // API Convenience to open a serialized Parquet file on disk, using built-in IO
+  // interface implementations that were created for testing, and may not be robust for
+  // all use cases.
   static std::unique_ptr<ParquetFileReader> OpenFile(const std::string& path,
       bool memory_map = true, ReaderProperties props = default_reader_properties());
 
--- a/src/parquet/util/input.cc
+++ b/src/parquet/util/input.cc
@@ -134,7 +134,10 @@ void MemoryMapSource::Close() {
 }
 
 void MemoryMapSource::CloseFile() {
-  if (data_ != nullptr) { munmap(data_, size_); }
+  if (data_ != nullptr) {
+    munmap(data_, size_);
+    data_ = nullptr;
+  }
 
   LocalFileSource::CloseFile();
 }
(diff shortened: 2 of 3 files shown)


wdforson:
NOTE: I haven't added any new tests for this change, since intercepting {{munmap}} calls feels like overkill for such a small change, particularly given that it affects only "sample" code. However, I've done a clean rebuild, and all of the existing tests are passing.

(that said, I haven't looked through the test source at all, so if there is some mocking infrastructure in place that would make such testing reasonably convenient, please let me know, and I'll be happy to revisit.)

src/parquet/util/input.cc
```diff
@@ -134,7 +134,10 @@ void MemoryMapSource::Close() {
 }
 
 void MemoryMapSource::CloseFile() {
-  if (data_ != nullptr) { munmap(data_, size_); }
+  if (data_ != nullptr) {
+    munmap(data_, size_);
+    data_ = nullptr;
+  }
```
wesm: It seems reasonable to add a test for idempotency. Calling `MemoryMapSource::Close` twice should trigger the issue?

src/parquet/file/reader.h
```diff
@@ -79,7 +79,8 @@ class PARQUET_EXPORT ParquetFileReader {
   ParquetFileReader();
   ~ParquetFileReader();
 
-  // API Convenience to open a serialized Parquet file on disk
+  // API Convenience to open a serialized Parquet file on disk, using SAMPLE
+  // implementations of RandomAccessSource
```
wesm: It would be clearer to state that this method uses built-in IO interface implementations that were created for testing, and may not be robust to all use cases.

src/parquet/util/input.h
```diff
@@ -55,6 +55,9 @@ class PARQUET_EXPORT RandomAccessSource {
   int64_t size_;
 };
 
+// ----------------------------------------------------------------------
+// SAMPLE implementations of RandomAccessSource, for reading files from local disk
```
wesm: "Implementations of RandomAccessSource used for testing and internal CLI tools. May not be robust for general production use."

src/parquet/util/input.cc
wdforson: Well, {{MemoryMapSource::Close}} is _always_ called at least twice. But the issue will only manifest in an environment where other threads are making concurrent {{mmap}} calls -- and even then, the manifestation is non-deterministic.

In general, the likelihood of manifestation should increase along with the number of threads making concurrent {{mmap}} and {{munmap}} calls. But in order for the issue to be triggered, a sequence of interleaved operations as illustrated in my commit message must occur (to include {{mmap}} having a tendency to reuse very recently unmapped segments, which I suspect could be platform-dependent).

All of that to say: the only reasonable testing approach I can think of here would be to call {{MemoryMapSource::Close}} multiple times for a single {{MemoryMapSource}} instance, and somehow confirm that {{munmap}} is only invoked once. The "cleanest" way to get such confirmation would be by intercepting {{munmap}} system calls...but I don't know how to do that off the top of my head.

src/parquet/file/reader.h
wdforson: ack

src/parquet/util/input.h
wdforson: ack

PR #204, opened 2016-12-19
PARQUET-805: Read Int96 into Arrow Timestamp(ns)

xhochy:


xhochy:
Also verified locally that this correctly reads the timestamps produced by Spark

src/parquet/arrow/reader.cc
```diff
@@ -44,6 +45,15 @@ using ParquetReader = parquet::ParquetFileReader;
 namespace parquet {
 namespace arrow {
 
+constexpr int64_t kJulianToUnixEpochDays = 2440588l;
+constexpr int64_t kNanosecondsInADay = 86400l * 1000l * 1000l * 1000l;
+
+int64_t impala_timestamp_to_nanoseconds(const Int96& impala_timestamp) {
+  int64_t days_since_epoch = impala_timestamp.value[2] - kJulianToUnixEpochDays;
+  int64_t nanoseconds = *(reinterpret_cast<const int64_t*>(&(impala_timestamp.value)));
+  return days_since_epoch * kNanosecondsInADay + nanoseconds;
```
wesm: Is this sufficient? I believe you that it is, but I had seen the Impala timestamp code (https://github.com/apache/incubator-impala/blob/master/be/src/runtime/timestamp-value.h) and had though there might be more work to be done

xhochy: Although there is a lot of code around on various sources on how to convert Julian dates to datetime objects, this is not so complicated as only have to deal with integer Julian dates. That's why I also verified with Spark INT96 output locally.

PR #206, opened 2016-12-20
PARQUET-812: Read BYTE_ARRAY with no logical type as arrow::BinaryArray

wesm:
Depends on ARROW-374. Need to merge that and update the thirdparty git hash


wesm:
It should be fixed by ARROW-374. i'm updating the hash now

PR #209, opened 2016-12-23
PARQUET-816: Workaround for incorrect column chunk metadata in parquet-mr <= 1.2.8

wesm:
This turned up in reading of old data files generated by parquet-mr in 2013. There's a bug in parquet-mr 1.2.8 and lower in which the column chunk metadata in the Parquet file is incorrect. Impala inserted an explicit workaround for this (see See https://github.com/apache/incubator-impala/blob/88448d1d4ab31eaaf82f764b36dc7d11d4c63c32/be/src/exec/hdfs-parquet-scanner.cc#L1227). 

In this particular file, the dictionary page header is 15 bytes, and the entire column chunk is: 15 (dict page header) + 277 (dictionary) + 17 (data page header) + 28 (data page) bytes, making 337 bytes.

But the metadata says the column chunk is only 322 bytes – the dict page header size got dropped from the accounting.


wesm:
Turns out this was more annoying than I thought, updated the title / description to explain the problem

majetideepak:
As part of PARQUET-793, I am implementing something very similar to  `parquet-mr/parquet-common/src/main/java/org/apache/parquet/SemanticVersion.java`. I will now extend the `FileVersion` accordingly.

src/parquet/file/metadata.h
```diff
@@ -91,6 +91,32 @@ class PARQUET_EXPORT RowGroupMetaData {
 
 class FileMetaDataBuilder;
 
+struct PARQUET_EXPORT FileVersion {
```
majetideepak: Should we move this to a separate file `util/file-version.h` / `util/file-version.cc`

src/parquet/file/metadata.h
```diff
@@ -91,6 +91,32 @@ class PARQUET_EXPORT RowGroupMetaData {
...
+    int major;
+    int minor;
+    int patch;
+  } version;
+
+  FileVersion() {}
+  explicit FileVersion(const std::string& created_by);
+
+  /// Returns true if version is strictly less than <major>.<minor>.<patch>
+  bool VersionLt(int major, int minor = 0, int patch = 0) const;
```
majetideepak: Should we just add a single `int compare(FileVersion& fv)` ?

wesm: Well, in this case the version check ("VersionLt(1, 2, 9)") is only carried out for parquet-mr -- it's not an equals/not-equals

src/parquet/file/metadata.h
wesm: It's small, and semantically part of `FileMetaData`. We could also make it a nested class `FileMetaData::Version`

wesm: Moved this to an inner class

src/parquet/file/metadata.h
majetideepak: I mean, we could combine `bool VersionEq` and `bool VersionLt` into `int compare(FileVersion& fv)`. This will require modifying `VersionLt(1, 2, 9)` to `compare(FileVersion(1, 2, 9)) < 0`.
 I also prefer `FileVersion` as the argument since we can extend `struct version` and still use the same `compare or VersionEq + VersionLt`.
PARQUET-793 will require extending `struct version`.

wesm: I'm still not quite following. Can you look at the version check in this patch and let me know what you're proposing changing the code to? We have to parse the application (here, parquet-mr) out of the version string. Having to construct a dummy FileVersion is a bit awkward. I would suggest we merge this patch and you can make changes to suit PARQUET-793

wesm: *created_by string

majetideepak: I will make the changes later to suit PARQUET-793.
I made a mistake in my earlier statement. `FileVersion` should be `Version`.
I will make changes on lines of what parquet-mr `SemanticVersion` class does, something as follows:
```
static FileMetaData::Version MR_PARQUET_129_FIXED_VERSION = FileMetaData::Version(1, 2, 9, "parquet-mr");
static FileMetaData::Version CDH_PARQUET_251_FIXED_VERSION = FileMetaData::Version(2, 5, 1, "cdh-5.5");

const FileMetaData::Version& version = file_metadata_->writer_version();
if (version.compare(MR_PARQUET_129_FIXED_VERSION) < 0) {
}
if (version.compare(CDH_PARQUET_255_FIXED_VERSION) < 0) {
}
```

Now `compare` in `SemanticVersion` is generalized as
```
int compare(FileMetadata::Version& o) {
   cmp = compareIntegers(major, o.major);
   if (cmp != 0) { return cmp; }

   cmp = compareIntegers(minor, o.minor);
   if (cmp != 0) { return cmp; }

   cmp = compareIntegers(patch, o.patch);
   if (cmp != 0) { return cmp; }

   cmp = compareString(o.application, application);
    if (cmp != 0) { return cmp; }
}
```

majetideepak: I now think the Java Implementation is buggy since it combines the application name into the comparator. The application name should be strictly equal / not equal.

majetideepak:
+1 LGTM

PR #207, opened 2016-12-21
PARQUET-813: Build thirdparty dependencies using ExternalProject

xhochy:

PR #210, opened 2016-12-29
PARQUET-818: Refactoring to utilize common IO, buffer, memory management abstractions and implementations

wesm:
This refactoring is a bit of a bloodbath, but I've attempted to preserve as much API backwards compatibility as possible. 

Several points

* Arrow does not use exceptions, so will need to be very careful about making sure that no Status goes unchecked. I've tried to get most of them, but might have missed some
* parquet-cpp still exposes an abstract file read and write API as before, but this makes it easy to pass in an Arrow file handle (e.g. HDFS, OS files, memory maps, etc.)
* Custom memory allocators will need to subclass `arrow::MemoryPool` instead. If this becomes onerous for some reason, we can try to find alternatives, but basically it's the exact same class as `parquet::MemoryAllocator`

Does not require any upstream changes in Arrow.


wesm:
No longer WIP. I got the build and unit tests passing again locally (so painful). Need to do a sweet of the codebase to make sure all buffer resizes and other possible errors are being checked appropriately

wesm:
@xhochy I added Arrow to the thirdparty deps in the README, can you handle adding to the thirdparty build toolchain in #207 after rebasing?

src/parquet/arrow/io.h
```diff
@@ -36,65 +37,25 @@ namespace arrow {
 // existing Arrow memory pool. This way we can direct all allocations to a
 // single place rather than tracking allocations in different locations (for
 // example: without utilizing parquet-cpp's default allocator)
-class PARQUET_EXPORT ParquetAllocator : public MemoryAllocator {
+class PARQUET_EXPORT ParquetAllocator : public MemoryPool {
```
xhochy: Is there a real need for this? Can't we just pass the plain MemoryPool to Parquet?

src/parquet/arrow/utils.h
```diff
@@ -24,31 +24,7 @@
 #include "parquet/exception.h"
```
xhochy: Remove this file, seems no longer needed.

src/parquet/util/memory.h
```diff
@@ -0,0 +1,438 @@
+// Licensed to the Apache Software Foundation (ASF) under one
```
xhochy: I have the feeling that we could later on move more of this file into Arrow.

xhochy:
@wesm #207 is ready for review/merge. Not sure what needs to be done there besides removing the conditional around Arrow.

xhochy:
Also I really like the ratio of lines changed vs lines removed here. Shows nicely the benefit of using Arrow.

src/parquet/arrow/io.h
wesm: good point, removed

wesm:
OK, this should be ready once the build is green. @majetideepak if there are any problems here that will impact your use of parquet-cpp as a library, can you open follow up JIRAs? Thanks!

wesm:
This is failing because of rpath/LD_LIBRARY_PATH issues. I will statically link Arrow in the unit tests to make this go away

src/parquet/file/reader-internal.h
```diff
@@ -103,8 +103,8 @@ class SerializedFile : public ParquetFileReader::Contents {
   // This class does _not_ take ownership of the data source. You must manage its
   // lifetime separately
   static std::unique_ptr<ParquetFileReader::Contents> Open(
-      std::unique_ptr<RandomAccessSource> source,
-      ReaderProperties props = default_reader_properties());
+      const std::shared_ptr<RandomAccessSource>& source,
```
majetideepak: What is the motivation for making this a `shared_ptr` ?

wesm: I can change it back, but semantically, why would we require that the file reader have exclusive access to a resource? You can always create a wrapper object that wraps a shared_ptr to some other IO interface (e.g., we could have `std::unique_ptr<ArrowInputFile>`, wrapping a shared ptr to one of the Arrow IO implementations).

wesm: Reverted the change, I agree that changing the pointer type in this patch is out of scope

wesm:
Added a static/shared linkage flag for the Arrow libraries. You still need the Arrow headers to build against libparquet, but it at least should simplify linking.

wesm:
@xhochy if you wouldn't mind taking a peek at the change I've made since your review, then this can be merged

CMakeLists.txt
```diff
@@ -692,12 +413,47 @@ if (${CLANG_TIDY_FOUND})
...
+  if(NOT PARQUET_BUILD_STATIC)
+    message(SEND_ERROR "Coverage requires the static lib to be built")
+  endif()
+endif()
+
+#############################################################
+# Apache Arrow linkage
+
+if ("${PARQUET_ARROW_LINKAGE}" STREQUAL "shared")
+  set(ARROW_LINK_LIBS
```
xhochy: This defines the overall linkage for both `parquet_shared` and `parquet_static` but  I would expect that with `PARQUET_ARROW_LINKAGE=shared`, `parquet_static` cannot be linked successfully.

xhochy: Ah, as there is no actual linking happening, this is no problem.

wesm: right -- the static linking is only creating an archive of object code

```
/usr/bin/cmake -E cmake_link_script CMakeFiles/parquet_static.dir/link.txt --verbose=1
ccache /usr/bin/ar cr debug/libparquet.a  CMakeFiles/parquet_objlib.dir/src/parquet/exception.cc.o CMakeFiles/parquet_objlib.dir/src/parquet/types.cc.o CMakeFiles/parquet_objlib.dir/src/parquet/column/levels.cc.o CMakeFiles/parquet_objlib.dir/src/parquet/column/reader.cc.o CMakeFiles/parquet_objlib.dir/src/parquet/column/writer.cc.o CMakeFiles/parquet_objlib.dir/src/parquet/column/scanner.cc.o CMakeFiles/parquet_objlib.dir/src/parquet/column/scan-all.cc.o CMakeFiles/parquet_objlib.dir/src/parquet/column/statistics.cc.o CMakeFiles/parquet_objlib.dir/src/parquet/compression/codec.cc.o CMakeFiles/parquet_objlib.dir/src/parquet/compression/brotli-codec.cc.o CMakeFiles/parquet_objlib.dir/src/parquet/compression/snappy-codec.cc.o CMakeFiles/parquet_objlib.dir/src/parquet/compression/gzip-codec.cc.o CMakeFiles/parquet_objlib.dir/src/parquet/file/metadata.cc.o CMakeFiles/parquet_objlib.dir/src/parquet/file/reader.cc.o CMakeFiles/parquet_objlib.dir/src/parquet/file/reader-internal.cc.o CMakeFiles/parquet_objlib.dir/src/parquet/file/writer.cc.o CMakeFiles/parquet_objlib.dir/src/parquet/file/writer-internal.cc.o CMakeFiles/parquet_objlib.dir/src/parquet/schema/converter.cc.o CMakeFiles/parquet_objlib.dir/src/parquet/schema/descriptor.cc.o CMakeFiles/parquet_objlib.dir/src/parquet/schema/printer.cc.o CMakeFiles/parquet_objlib.dir/src/parquet/schema/types.cc.o CMakeFiles/parquet_objlib.dir/src/parquet/util/cpu-info.cc.o CMakeFiles/parquet_objlib.dir/src/parquet/util/memory.cc.o
ccache /usr/bin/ranlib debug/libparquet.a
```

wesm:
thanks all, merging

wesm:
That felt good.

PR #213, opened 2017-01-04
PARQUET-807: Allow user to retain ownership of parquet::FileMetaData. 

wesm:
Also implements PARQUET-808: opening file with existing metadata object. 

This allows a user to create a reader only for the purposes of obtaining the metadata. 

Do you all think it's worth having a convenience method for reading the metadata out of a file?

### DIFF
--- a/src/parquet/arrow/schema.cc
+++ b/src/parquet/arrow/schema.cc
@@ -364,12 +364,6 @@ Status FieldToNode(const std::shared_ptr<Field>& field,
       type = ParquetType::INT64;
       logical_type = LogicalType::TIMESTAMP_MILLIS;
     } break;
-    case ArrowType::TIMESTAMP_DOUBLE:
-      type = ParquetType::INT64;
-      // This is specified as seconds since the UNIX epoch
-      // TODO: Converted type in Parquet?
-      // logical_type = LogicalType::TIMESTAMP_MILLIS;
-      break;
     case ArrowType::TIME:
       type = ParquetType::INT64;
       logical_type = LogicalType::TIME_MILLIS;
--- a/src/parquet/file/metadata.cc
+++ b/src/parquet/file/metadata.cc
@@ -325,6 +325,7 @@ class FileMetaData::FileMetaDataImpl {
 
 std::shared_ptr<FileMetaData> FileMetaData::Make(
     const uint8_t* metadata, uint32_t* metadata_len) {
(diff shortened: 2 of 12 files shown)


src/parquet/arrow/schema.cc
```diff
@@ -364,12 +364,6 @@ Status FieldToNode(const std::shared_ptr<Field>& field,
       type = ParquetType::INT64;
       logical_type = LogicalType::TIMESTAMP_MILLIS;
     } break;
-    case ArrowType::TIMESTAMP_DOUBLE:
-      type = ParquetType::INT64;
-      // This is specified as seconds since the UNIX epoch
-      // TODO: Converted type in Parquet?
-      // logical_type = LogicalType::TIMESTAMP_MILLIS;
-      break;
```
wesm: I'm planning to remove this type code from Arrow for now.

xhochy: As long as Drill has no dependency on that, +1!

src/parquet/file/metadata.cc
```diff
@@ -323,9 +323,9 @@ class FileMetaData::FileMetaDataImpl {
   FileMetaData::Version writer_version_;
 };
 
-std::unique_ptr<FileMetaData> FileMetaData::Make(
+std::shared_ptr<FileMetaData> FileMetaData::Make(
```
xhochy: This function is now just an alias for `make_shared`. Do we really want to keep it?

xhochy:
> Do you all think it's worth having a convenience method for reading the metadata out of a file?

Definitely, especially as tools such as Spark provide files as `_common_metadata` and `_metadata` that are mainly useful for inferring the schema of all files in a directory.

src/parquet/file/metadata.cc
wesm: The FileMetaData ctor used here is private, I will add a comment.

wesm:
OK. Will add a method for reading the metadata only for a file.

In these special `_metadata` files, do you know if they have the `PAR1` footer and metadata length, or are they only the serialized Thrift message? If the latter, then there should be an option whether the input is a full Parquet file or simply a metadata-only file

wesm:
I'll add whatever's needed to read serialized-metadata-only files in pyarrow, for now the convenience method can just read from a file footer.

wesm:
@xhochy I expanded the scope to include PARQUET-808. Could you review again? Thanks

PR #215, opened 2017-01-07
PARQUET-827: Account for arrow::MemoryPool API change and fix bug in reading Int96 timestamps

wesm:
Prior to ARROW-427, the "length()" field was not being properly checked in the implementation of `PrimitiveArray::Equals`.


wesm:
agreed

PR #216, opened 2017-01-09
PARQUET-828: Do not implicitly cast ParquetVersion enum to int

wesm:
See https://github.com/apache/parquet-mr/blob/df9d8e415436292ae33e1ca0b8da256640de9710/parquet-hadoop/src/main/java/org/apache/parquet/hadoop/ParquetFileWriter.java#L86, this number should be 1 for Parquet 1.0 files, I believe.

### DIFF
--- a/src/parquet/file/file-metadata-test.cc
+++ b/src/parquet/file/file-metadata-test.cc
@@ -31,7 +31,10 @@ TEST(Metadata, TestBuildAccess) {
   parquet::schema::NodePtr root;
   parquet::SchemaDescriptor schema;
 
-  std::shared_ptr<WriterProperties> props = WriterProperties::Builder().build();
+  WriterProperties::Builder prop_builder;
+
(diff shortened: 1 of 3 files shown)


src/parquet/file/file-metadata-test.cc
```diff
@@ -142,8 +145,8 @@ TEST(Metadata, TestBuildAccess) {
   ASSERT_EQ(nrows / 2, rg2_column2->num_values());
   ASSERT_EQ(DEFAULT_COMPRESSION_TYPE, rg2_column1->compression());
   ASSERT_EQ(DEFAULT_COMPRESSION_TYPE, rg2_column2->compression());
-  ASSERT_EQ(2, rg2_column1->encodings().size());
```
xhochy: Huh, how did this happen?

xhochy: Ah, it's because of the version switch.

PR #222, opened 2017-01-23
PARQUET-835: Read Arrow columns in parallel with thread pool

wesm:
Also implements PARQUET-836, but need to add a unit test for that

### DIFF
--- a/src/parquet/arrow/arrow-reader-writer-test.cc
+++ b/src/parquet/arrow/arrow-reader-writer-test.cc
@@ -17,6 +17,8 @@
 
 #include "gtest/gtest.h"
 
+#include <sstream>
+
 #include "parquet/api/reader.h"
 #include "parquet/api/writer.h"
 
@@ -44,7 +46,6 @@ using parquet::schema::NodePtr;
 using parquet::schema::PrimitiveNode;
 
 namespace parquet {
-
 namespace arrow {
 
 const int SMALL_SIZE = 100;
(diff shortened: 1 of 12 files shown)


src/parquet/arrow/arrow-reader-writer-test.cc
```diff
@@ -184,6 +185,23 @@ using ParquetDataType = DataType<test_traits<T>::parquet_enum>;
...
+
+  std::shared_ptr<Buffer> buffer = sink->GetBuffer();
+  std::unique_ptr<FileReader> reader;
+  ASSERT_OK_NO_THROW(
+      OpenFile(std::make_shared<BufferReader>(buffer), ::arrow::default_memory_pool(),
+          ::parquet::default_reader_properties(), nullptr, &reader));
+
+  reader->set_num_threads(num_threads);
+  ASSERT_OK_NO_THROW(reader->ReadFlatTable(out));
+}
```
wesm: @xhochy I'm wondering if this test module can be simplified using a slightly more generalized version of this function

xhochy:
In general this seems fine but once we would add S3 support to Arrow IO, this would definitely need to be reworked. The I/O is hidden inside of this code and would need to be more explicit to be tunable for the underlying storage access cost model. 

Not saying that parallel I/O doesn't work with object stores (it actually can be beneficial) but it behaves differently than to an SSD where you can seek randomly.

wesm:
We should look at Impala's S3 support for inspiration (they spent a long time designing / building it as I recall).

xhochy:
The only thing I found in Impala's source code was that they actually redirect to HDFS (read the s3a implementation in Hadoop) in `be/src/runtime/hdfs-fs-cache.{h,cc}` and don't actually implement it in native code.

wesm:
Thanks. I'll add a unit test for PARQUET-836 in a follow up patch

PR #224, opened 2017-01-24
PARQUET-691: Write ColumnChunk metadata after chunk is complete

wesm:
See corresponding logic in Impala's Parquet implementation: https://github.com/apache/incubator-impala/blob/master/be/src/exec/hdfs-parquet-table-writer.cc#L973


majetideepak:
+1  LGTM.

PR #225, opened 2017-01-25
PARQUET-841: Version number being incorrectly written for v1 files

wesm:
In introduced this bug recently and it wasn't properly being tested.


majetideepak:
+1 LGTM

PR #226, opened 2017-01-25
PARQUET-842: Do not set unnecessary fields in the parquet::SchemaElement

wesm:
The Impala Parquet scanner can reject our data otherwise


majetideepak:
+1

PR #227, opened 2017-01-25
PARQUET-843: Impala is thrown off by a REPEATED root schema node

wesm:
We do not use the repetition level of the root node in the schema, and neither does parquet-mr. However, Impala 2.8.0 increases the max_repetition_level and expects to find repetition levels that are not being written out. With this change, Impala is able to read our files again.

### DIFF
--- a/src/parquet/arrow/schema.cc
+++ b/src/parquet/arrow/schema.cc
@@ -406,7 +406,7 @@ Status ToParquetSchema(const ::arrow::Schema* arrow_schema,
     RETURN_NOT_OK(FieldToNode(arrow_schema->field(i), properties, &nodes[i]));
   }
 
-  NodePtr schema = GroupNode::Make("schema", Repetition::REPEATED, nodes);
+  NodePtr schema = GroupNode::Make("schema", Repetition::REQUIRED, nodes);
   *out = std::make_shared<::parquet::SchemaDescriptor>();
   PARQUET_CATCH_NOT_OK((*out)->Init(schema));
 
--- a/src/parquet/file/metadata.cc
+++ b/src/parquet/file/metadata.cc
@@ -485,16 +485,17 @@ class ColumnChunkMetaDataBuilder::ColumnChunkMetaDataBuilderImpl {
     column_chunk_->meta_data.__set_total_uncompressed_size(uncompressed_size);
     column_chunk_->meta_data.__set_total_compressed_size(compressed_size);
     std::vector<format::Encoding::type> thrift_encodings;
-    thrift_encodings.push_back(ToThrift(Encoding::RLE));
     if (has_dictionary) {
-      thrift_encodings.push_back(ToThrift(properties_->dictionary_page_encoding()));
-      // add the encoding only if it is unique
-      if (properties_->version() == ParquetVersion::PARQUET_2_0) {
-        thrift_encodings.push_back(ToThrift(properties_->dictionary_index_encoding()));
+      thrift_encodings.push_back(ToThrift(properties_->dictionary_index_encoding()));
+      if (properties_->version() == ParquetVersion::PARQUET_1_0) {
+        thrift_encodings.push_back(ToThrift(Encoding::PLAIN));
+      } else {
+        thrift_encodings.push_back(ToThrift(properties_->dictionary_page_encoding()));
       }
     } else {  // Dictionary not enabled
       thrift_encodings.push_back(ToThrift(properties_->encoding(column_->path())));
     }
+    thrift_encodings.push_back(ToThrift(Encoding::RLE));
     // Only PLAIN encoding is supported for fallback in V1
     // TODO(majetideepak): Use user specified encoding for V2
     if (dictionary_fallback) { thrift_encodings.push_back(ToThrift(Encoding::PLAIN)); }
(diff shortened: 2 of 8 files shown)


src/parquet/file/metadata.cc
```diff
@@ -485,16 +485,17 @@ class ColumnChunkMetaDataBuilder::ColumnChunkMetaDataBuilderImpl {
...
+      thrift_encodings.push_back(ToThrift(properties_->dictionary_index_encoding()));
+      if (properties_->version() == ParquetVersion::PARQUET_1_0) {
+        thrift_encodings.push_back(ToThrift(Encoding::PLAIN));
+      } else {
+        thrift_encodings.push_back(ToThrift(properties_->dictionary_page_encoding()));
       }
     } else {  // Dictionary not enabled
       thrift_encodings.push_back(ToThrift(properties_->encoding(column_->path())));
     }
+    thrift_encodings.push_back(ToThrift(Encoding::RLE));
```
majetideepak: What is the reason for the re-order of encodings ?

src/parquet/file/metadata.cc
```diff
@@ -485,16 +485,17 @@ class ColumnChunkMetaDataBuilder::ColumnChunkMetaDataBuilderImpl {
...
     column_chunk_->meta_data.__set_total_compressed_size(compressed_size);
     std::vector<format::Encoding::type> thrift_encodings;
-    thrift_encodings.push_back(ToThrift(Encoding::RLE));
     if (has_dictionary) {
-      thrift_encodings.push_back(ToThrift(properties_->dictionary_page_encoding()));
-      // add the encoding only if it is unique
-      if (properties_->version() == ParquetVersion::PARQUET_2_0) {
-        thrift_encodings.push_back(ToThrift(properties_->dictionary_index_encoding()));
+      thrift_encodings.push_back(ToThrift(properties_->dictionary_index_encoding()));
+      if (properties_->version() == ParquetVersion::PARQUET_1_0) {
```
majetideepak: I don't think we need this check anymore. `properties_->dictionary_page_encoding()` also check it ?
I think the spec says: https://github.com/Parquet/parquet-format/blob/master/src/thrift/parquet.thrift
PARQUET_1_0 : data (index) page: PLAIN_DICTIONARY, dictionary page: PLAIN_DICTIONARY
PARQUET_2_0 : data (index) page: RLE_DICTIONARY, dictionary page: PLAIN

majetideepak: I am not sure if Impala has a different convention.

src/parquet/file/metadata.cc
wesm: I can change the order back -- I was trying to account for any differences in the metadata that might be preventing Impala from reading our files.

src/parquet/file/metadata.cc
```diff
@@ -485,16 +485,17 @@ class ColumnChunkMetaDataBuilder::ColumnChunkMetaDataBuilderImpl {
...
+        thrift_encodings.push_back(ToThrift(Encoding::PLAIN));
+      } else {
+        thrift_encodings.push_back(ToThrift(properties_->dictionary_page_encoding()));
```
wesm: We were not adding `Encoding::PLAIN` to the list of encodings in the metadata -- Impala has all 3 when you have dictionary encoding, PLAIN, PLAIN_ENCODING, and RLE. I don't believe it impacts parquet-cpp or parquet-mr

src/parquet/file/metadata.cc
wesm: To be clear, if this causes you a problem, I can change it, otherwise I will leave it as is (this is the order of encodings that Impala writes out)

src/parquet/file/metadata.cc
wesm: The problem is that `dictionary_page_encoding()` returns `PLAIN_DICTIONARY` for V1 files: https://github.com/apache/parquet-cpp/blob/master/src/parquet/column/properties.h#L325

Since Impala writes `Encoding::PLAIN` in the metadata even when dictionary-encoded, I don't want to risk incompatibilities

src/parquet/file/metadata.cc
majetideepak: I have no problem. This is fine.

src/parquet/file/metadata.cc
majetideepak: Thanks for the clarification!

src/parquet/file/metadata.cc
majetideepak: Sounds good!

majetideepak:
+1 LGTM

src/parquet/file/metadata.cc
timarmstrong: It doesn't look like this would make a difference for Impala - we just check each encoding in the list independently to see if it's one Impala knows how to handle.

src/parquet/file/metadata.cc
timarmstrong: Looks like Impala just does this as a shortcut, there's a TODO in the code to fix: https://github.com/apache/incubator-impala/blob/master/be/src/exec/hdfs-parquet-table-writer.cc#L793

Hopefully nobody relies on this.

src/parquet/arrow/schema.cc
```diff
@@ -406,7 +406,7 @@ Status ToParquetSchema(const ::arrow::Schema* arrow_schema,
     RETURN_NOT_OK(FieldToNode(arrow_schema->field(i), properties, &nodes[i]));
   }
 
-  NodePtr schema = GroupNode::Make("schema", Repetition::REPEATED, nodes);
+  NodePtr schema = GroupNode::Make("schema", Repetition::REQUIRED, nodes);
```
timarmstrong: parquet.thrift seems to think that the root schema should not have any repetition set:

https://github.com/apache/parquet-format/blob/master/src/main/thrift/parquet.thrift#L231

I'm assuming other parquet implementations probably just ignore the root schema's repetition level, but it triggers some logic in Impala to handle nested types.

wesm:
thanks @timarmstrong

PR #229, opened 2017-01-26
PARQUET-834: Support I/O of arrow::ListArray

xhochy:


xhochy:
Depends on https://github.com/apache/arrow/pull/304

xhochy:
@wesm Review would be appreciated. I removed the `Flat` prefixes as the API should be usable also for non-flat arrays. As part of this PR I would now implement also the `definition_level == 2` case but that should have no more API impact.

Performance will be okish, we can work on that post-release as it shouldn't have an impact on interfaces.

wesm:
Great, will review today.

wesm:
I'm about halfway through reviewing, will try to finish in a little bit

src/parquet/arrow/reader.cc
```diff
@@ -706,48 +821,52 @@ Status FlatColumnReader::Impl::TypedReadBatch<::arrow::StringType, ByteArrayType
...
-      switch (timestamp_type->unit) {
-        case ::arrow::TimeUnit::MILLI:
-          return TypedReadBatch<::arrow::TimestampType, Int64Type>(batch_size, out);
-          break;
-        case ::arrow::TimeUnit::NANO:
-          return TypedReadBatch<::arrow::TimestampType, Int96Type>(batch_size, out);
-          break;
-        default:
-          return Status::NotImplemented("TimeUnit not supported");
+  if (descr_->max_repetition_level() <= 1) {
```
wesm: if you make this instead `if (descr_->max_repetition_level() > 1)` and exit earlier, it will reduce the diff noise in subsequent patches

src/parquet/arrow/writer.cc
```diff
@@ -342,29 +358,297 @@ Status FileWriter::Impl::WriteFlatColumnChunk(
   return Status::OK();
 }
 
+template <typename ParquetType, typename ArrowType>
+Status FileWriter::Impl::WriteListBatch(ColumnWriter* column_writer,
+    const ListArray* data, int64_t offset, int64_t length, int64_t num_levels,
```
wesm: Why pass in `ListArray` here at all, and these functions below? All you need from the ListArray is the begin and end offset. 

Having list-related code specialized for individual types in general would be best avoided -- you construct the repetition and definition levels representing a certain number of logical "values" (per the discussion above), then write the correct number of values from the innermost leaf node. The trouble with this code is that it will need to be immediately refactored as soon as we are dealing in structs or more levels of lists. 

I suspect that TypedWriteBatch can be generalized or split up until submethods

src/parquet/arrow/writer.cc
```diff
@@ -342,29 +358,297 @@ Status FileWriter::Impl::WriteFlatColumnChunk(
...
+  }
+  std::shared_ptr<Array> def_levels_array;
+  def_levels.Finish(&def_levels_array);
+  const int16_t* def_levels_ptr = reinterpret_cast<const int16_t*>(
+      static_cast<Int16Array*>(def_levels_array.get())->data()->data());
+
+  std::shared_ptr<Array> rep_levels_array;
+  rep_levels.Finish(&rep_levels_array);
+  const int16_t* rep_levels_ptr = reinterpret_cast<const int16_t*>(
+      static_cast<Int16Array*>(rep_levels_array.get())->data()->data());
```
wesm: Can you factor out the construction of the repetition and definition levels into a separate method? When we are dealing with multiple levels of nested types, this function will eventually need to accept a vector of offsets arrays and a vector of bitmaps

src/parquet/arrow/reader.cc
```diff
@@ -521,93 +534,182 @@ Status FlatColumnReader::Impl::TypedReadBatch(
...
+    }
+  }
+  return Status::OK();
+}
+
+template <typename ArrowType, typename ParquetType>
+Status ColumnReader::Impl::TypedReadBatch(int batch_size, std::shared_ptr<Array>* out) {
+  using ArrowCType = typename ArrowType::c_type;
+
+  int values_to_read = batch_size;
```
xhochy: Yes, the values_to_read refer to the actual levels we are going to read. With the current abstractions it isn't simple to ensure that we won't end reading in the middle of a list as there are no methods that provide some kind of peek what will come next. Implementing such methods will probably make this much more complicated here. For a start, I would rather add a warning comment to the header.

xhochy: Also I'll rename this to `levels_to_read`.

wesm: Until batch_size always refers to the number of logical rows, it will be a source of issues. Doesn't have to get done in this patch but it will need to maintain state in between batch read calls

wesm:
I'll hold off on reviewing more until you tell me so

xhochy:
@wesm This PR is ready for review again. Now also supports arbitrary depth list-of-lists

wesm:
Fantastic. I will review today

xhochy:
Yes, it will raise at https://github.com/apache/parquet-cpp/pull/229/files#diff-2f5ceebd1726b16db561185cc620d18eR536

PR #235, opened 2017-02-05
PARQUET-865: Pass all CXXFLAGS to Thrift ExternalProject

xhochy:


xhochy:
I'll extend this path to also include the flags for other dependencies, so please hold of with merging.

xhochy:
ok, done. This was needed that we can also introduce AddressSanitizer in parquet-cpp.

cmake_modules/ThirdpartyToolchain.cmake
```diff
@@ -216,9 +225,9 @@ if(PARQUET_BUILD_TESTS)
 
   if("$ENV{GTEST_HOME}" STREQUAL "")
     if(APPLE)
-      set(GTEST_CMAKE_CXX_FLAGS "-fPIC -std=c++11 -stdlib=libc++ -DGTEST_USE_OWN_TR1_TUPLE=1 -Wno-unused-value -Wno-ignored-attributes")
```
wesm: That's interesting if this does not cause problems on OS X -- I will wait for the build to run

xhochy: That's because all the options I removed are already included in parquet-cpp's `CMAKE_CXX_FLAGS`.

PR #236, opened 2017-02-06
PARQUET-866: API fixes for ARROW-33 patch

wesm:
See ARROW-33 patch https://github.com/apache/arrow/pull/322

@xhochy this fails on Int96 timestamps. I'm not sure why yet


wesm:
per discussion in https://github.com/apache/arrow/pull/322, I think the Int96 thing is a bug that was exposed by the refactoring. if you could take a look and let me know if you agree

wesm:
here's the view from inside gdb

```
(gdb) p left_data
$1 = (const uint8_t *) 0xcd39c0 "/\263\256\243\201t"
(gdb) p right_data
$2 = (const uint8_t *) 0xcd2840 "/\263\375\064", <incomplete sequence \303>
```

The first byte is equal, and because of the bug that was the only byte being compared amazingly.

wesm:
OK, fixed the off-by-one error, the build should pass now

wesm:
build passed on my fork: https://travis-ci.org/wesm/parquet-cpp/builds/198973522. Merging

PR #238, opened 2017-02-08
PARQUET-867: Support writing sliced Arrow arrays

xhochy:


xhochy:
Depends on https://github.com/apache/arrow/pull/329

src/parquet/arrow/writer.cc
```diff
@@ -431,8 +436,8 @@ Status FileWriter::Impl::TypedWriteBatch<ByteArrayType, ::arrow::BinaryType>(
   if (writer->descr()->schema_node()->is_required() || (data->null_count() == 0)) {
     // no nulls, just dump the data
     for (int64_t i = 0; i < num_values; i++) {
-      buffer_ptr[i] =
-          ByteArray(data->value_length(i + offset), data_ptr + data->value_offset(i));
+      buffer_ptr[i] = ByteArray(
+          data->value_length(i + offset), data_ptr + data->value_offset(i + offset));
```
wesm: is `data` unsliced? the `value_length` and `value_offset` functions account for the offset if there is one

src/parquet/arrow/writer.h
```diff
@@ -50,8 +50,7 @@ class PARQUET_EXPORT FileWriter {
   FileWriter(::arrow::MemoryPool* pool, std::unique_ptr<ParquetFileWriter> writer);
 
   ::arrow::Status NewRowGroup(int64_t chunk_size);
-  ::arrow::Status WriteColumnChunk(
-      const ::arrow::Array* data, int64_t offset = 0, int64_t length = -1);
+  ::arrow::Status WriteColumnChunk(const ::arrow::Array* data);
```
wesm: It's not in scope for this patch, but we should consider it before 1.0: this argument should really be `const Array&` instead of `const Array*` -- there's other places where `const T*` should be `const T&`. see https://google.github.io/styleguide/cppguide.html#Reference_Arguments

```
However, there are some instances where using const T* is preferable to const T& for input parameters. For example:

* You want to pass in a null pointer.
* The function saves a pointer or reference to the input.
```

xhochy: I'll do it in this patch, so we get all those API breakages done quickly.

src/parquet/arrow/writer.cc
xhochy: Now using slice instead of length and offset.

src/parquet/arrow/writer.cc
```diff
@@ -389,19 +391,18 @@ NULLABLE_BATCH_FAST_PATH(DoubleType, ::arrow::DoubleType, double)
...
   auto buffer_ptr = reinterpret_cast<bool*>(data_buffer_.mutable_data());
   auto writer = reinterpret_cast<TypedColumnWriter<BooleanType>*>(column_writer);
 
   int buffer_idx = 0;
-  for (int i = 0; i < num_values; i++) {
-    if (!data->IsNull(offset + i)) {
-      buffer_ptr[buffer_idx++] = BitUtil::GetBit(data_ptr, offset + i);
+  for (int i = 0; i < data->length(); i++) {
+    if (!data->IsNull(i)) {
+      buffer_ptr[buffer_idx++] = BitUtil::GetBit(data_ptr, array->offset() + i);
```
wesm: Now that `array` is `shared_ptr<Array>`, this may affect performance (the compiler may not be able to reason about whether `array->offset()` is a loop invariant

src/parquet/arrow/writer.cc
```diff
@@ -430,18 +430,17 @@ Status FileWriter::Impl::TypedWriteBatch<ByteArrayType, ::arrow::BinaryType>(
 
   if (writer->descr()->schema_node()->is_required() || (data->null_count() == 0)) {
     // no nulls, just dump the data
-    for (int64_t i = 0; i < num_values; i++) {
-      buffer_ptr[i] =
-          ByteArray(data->value_length(i + offset), data_ptr + data->value_offset(i));
+    for (int64_t i = 0; i < data->length(); i++) {
+      buffer_ptr[i] = ByteArray(data->value_length(i), data_ptr + data->value_offset(i));
```
wesm: For performance, recommend caching `const int32_t* value_offsets = data->raw_value_offsets()` (which is shifted) and using the offsets to compute the value lengths

src/parquet/arrow/writer.cc
```diff
@@ -430,18 +430,17 @@ Status FileWriter::Impl::TypedWriteBatch<ByteArrayType, ::arrow::BinaryType>(
...
   } else {
     int buffer_idx = 0;
-    for (int64_t i = 0; i < num_values; i++) {
-      if (!data->IsNull(offset + i)) {
-        buffer_ptr[buffer_idx++] = ByteArray(
-            data->value_length(i + offset), data_ptr + data->value_offset(i + offset));
+    for (int64_t i = 0; i < data->length(); i++) {
+      if (!data->IsNull(i)) {
+        buffer_ptr[buffer_idx++] =
+            ByteArray(data->value_length(i), data_ptr + data->value_offset(i));
```
wesm: Same as above

wesm:
Can you enable Travis CI on your fork to get occasional faster builds? https://travis-ci.org/xhochy/parquet-cpp

I've been asking ASF Infra to throttle projects that are hogging the build slaves, but the builds are still taking a long time

xhochy:
Activated and triggered build: https://travis-ci.org/xhochy/parquet-cpp/builds/200022807

wesm:
Very strange error!

xhochy:
No, that's sadly quite common. That's Alt+Space on Mac :(

wesm:
Successful build, despite appearances: https://travis-ci.org/xhochy/parquet-cpp. @xhochy if you enable coveralls for xhochy/parquet-cpp you will get green builds on your fork. forgot that was a dependency

Aside: our coverage stats are messed up because of the thirdparty libs, I'll create a JIRA to fix

PR #244, opened 2017-02-14
PARQUET-878: Remove setup_build_env from rc-verification script

xhochy:

PR #245, opened 2017-02-14
PARQUET-879: Use same configure command for Thrift

xhochy:


wesm:
I'm trying this locally (CMake 2.8.x) due to the failure in #244

PR #248, opened 2017-02-15
PARQUET-882: Improve Application Version parsing

majetideepak:


src/parquet/file/metadata.cc
```diff
@@ -474,29 +475,33 @@ void FileMetaData::WriteTo(OutputStream* dst) {
...
-    ba::split(version_tokens, version_string_trimmed, ba::is_any_of("."));
-    version.major = version_tokens.size() >= 1 ? atoi(version_tokens[0].c_str()) : 0;
-    version.minor = version_tokens.size() >= 2 ? atoi(version_tokens[1].c_str()) : 0;
-    version.patch = version_tokens.size() >= 3 ? atoi(version_tokens[2].c_str()) : 0;
+  bool app_success = std::regex_match(created_by_lower, app_matches, app_regex);
+  bool ver_success = false;
+
+  if (app_success && app_matches.size() >= 4) {
+    // first match is the entire string. sub-matches start from second.
+    application_ = app_matches[1];
```
wesm: Appears `application_` is only being set in this branch, needs to be set to `"unknown"` otherwise

majetideepak:
@wesm, the file `data/nation.dict-malformed.parquet` you added to fix PARQUET-816 has an invalid created_by string `parquet-mr`.
The corresponding test is failing in this patch because of this. Can you tell me how this file was generated? Thanks.

wesm:
It was created by an early version of parquet-mr (prior to 1.2.8). I believe we'll need to be robust to this

majetideepak:
I had to use `boost::regex` since `std::regex` seems to have issues on a MacOS with these regular expressions. 
https://travis-ci.org/apache/parquet-cpp/builds/203466061

wesm:
@xhochy if we are going to do another RC, perhaps we should pull in this patch?

xhochy:
We should include this in the next RC.

PR #250, opened 2017-02-16
Fixed reading of nested columns.

peshopetrov:
Nested repeated columns' count is incorrectly read from row group's metadata. That's correct in cases where there aren't any nested repeated fields but is generally not correct. Instead the num_values from the column's metadata should be used.


mvertes:
Hello, I confirm that reading nested repeated columns is broken. Looking at the number of returned values, It seems that the reader uses the number of rows rather than the number of values in the column. However your fix doesn't compile right now:

```
parquet-cpp/src/parquet/file/reader-internal.cc:199:58: error: no member named 'allocator' in 'parquet::ReaderProperties'
      col->num_values(), col->compression(), properties_.allocator()));
                                             ~~~~~~~~~~~ ^
1 error generated.
```

mvertes:
The line should be:

parquet-cpp/src/parquet/file/reader-internal.cc:199

```
      col->num_values(), col->compression(), properties_.memory_pool()));
```

Then it compiles, and fixes the problem.

wesm:
I'm sorry to have been slow to respond -- between @mvertes and @peshopetrov if you can come up with a unit test for this, that would be very helpful. Since we're trying to get a release out the door we may merge the fix in #261 if making a test case is difficult (leave it up to @xhochy)

PR #251, opened 2017-02-19
Upgrade thrift revision number to 0.10.0

jlafaye:
Default thrift revision is 0.9.1 which gives the following compilation
error with gcc 6 (I am using 6.2.0)

src/generate/t_java_generator.cc:2830:14: error: operands to ?: have
different types ‘bool’ and ‘std::basic_ostream<char>’
        first ? first = false : indent(f_service_) << "else ";
                ~~~~~~^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~

Upgrading to the latest version with the same major
number, 0.9.3, yields the compilation error below. SSLV3
has been deprecated because of security issues (see e.g.
                http://disablessl3.com/)

src/thrift/transport/TSSLSocket.cpp: In constructor 'apache::thrift::transport::SSLContext::SSLContext(
  const apache::thrift::transport::SSLProtocol&)': src/thrift/transport/TSSLSocket.cpp:143:37: error: 'SSLv3_method' was not declared in this scope
ctx_ = SSL_CTX_new(SSLv3_method());
^
Makefile:1280 : la recette pour la cible « src/

Upgrading to 0.10.0 works  fine

### DIFF
--- a/cmake_modules/ThirdpartyToolchain.cmake
+++ b/cmake_modules/ThirdpartyToolchain.cmake
@@ -18,7 +18,7 @@
 set(GTEST_VERSION "1.7.0")
 set(GBENCHMARK_VERSION "1.0.0")
 set(SNAPPY_VERSION "1.1.3")
-set(THRIFT_VERSION "0.9.1")
+set(THRIFT_VERSION "0.10.0")
 
 # Brotli 0.5.2 does not install headers/libraries yet, but 0.6.0.dev does
 set(BROTLI_VERSION "5db62dcc9d386579609540cdf8869e95ad334bbd")
--- a/cmake_modules/ThirdpartyToolchain.cmake
+++ b/cmake_modules/ThirdpartyToolchain.cmake
@@ -42,9 +42,6 @@ set(EP_C_FLAGS "${CMAKE_C_FLAGS} ${CMAKE_C_FLAGS_${UPPERCASE_BUILD_TYPE}} -fPIC"
 find_package(Thrift)
 
 if (NOT THRIFT_FOUND)
-  if (APPLE)
-      message(FATAL_ERROR "thrift compilation under OSX is not currently supported.")
-  endif()
 
   set(THRIFT_PREFIX "${CMAKE_CURRENT_BINARY_DIR}/thrift_ep/src/thrift_ep-install")
   set(THRIFT_HOME "${THRIFT_PREFIX}")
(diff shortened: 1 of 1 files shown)


cmake_modules/ThirdpartyToolchain.cmake
```diff
@@ -18,7 +18,7 @@
 set(GTEST_VERSION "1.7.0")
 set(GBENCHMARK_VERSION "1.0.0")
 set(SNAPPY_VERSION "1.1.3")
-set(THRIFT_VERSION "0.9.1")
+set(THRIFT_VERSION "0.10.0")
```
wesm: Can you update the ExternalProject build instructions to use CMake?

jlafaye: Do you mean the 'Third Party Dependencies' section in README.md ?

wesm: No, I mean this: https://github.com/apache/parquet-cpp/blob/master/cmake_modules/ThirdpartyToolchain.cmake#L56

jlafaye: will do, I hope it will fix CI errors at the same time

jlafaye: Done

wesm:
Sweet. @xhochy can you also take a look?

@jlafaye would you mind opening a JIRA on issues.apache.org and prefixing the PR title with "PARQUET-XXX:" where XXX is the JIRA number? Thanks

cmake_modules/ThirdpartyToolchain.cmake
```diff
@@ -42,41 +42,43 @@ set(EP_C_FLAGS "${CMAKE_C_FLAGS} ${CMAKE_C_FLAGS_${UPPERCASE_BUILD_TYPE}} -fPIC"
...
     ExternalProject_Add(thrift_ep
-      CONFIGURE_COMMAND ${THRIFT_CONFIGURE_COMMAND}
-      BUILD_IN_SOURCE 1
-      # This is needed for 0.9.1 and can be removed for 0.9.3 again
-      BUILD_COMMAND make clean
-      INSTALL_COMMAND make install
-      INSTALL_DIR ${THRIFT_PREFIX}
       URL "http://archive.apache.org/dist/thrift/${THRIFT_VERSION}/thrift-${THRIFT_VERSION}.tar.gz"
-      )
+      BUILD_BYPRODUCTS "${THRIFT_STATIC_LIB}" "${THRIFT_COMPILER}"
```
xhochy: The `else` part shouldn't include `BUILD_BYPRODUCTS`. This unsupported in not-so-old CMake versions, thus we have the two nearly-identical `ExternalProject_Add` calls.

xhochy:
We already have a JIRA for this https://issues.apache.org/jira/browse/PARQUET-849

wesm:
@jlafaye can you change your PR title to start with `PARQUET-849:`? Please also change the OS X build to install boost instead of Thrift, so we can verify that this works for OS X as well: https://github.com/apache/parquet-cpp/blob/master/ci/before_script_travis.sh#L17

thanks

wesm:
I'm going to open a new PR since the other patches are blocked on this

PR #252, opened 2017-02-20
PARQUET-885: Do not search for Thrift in default system paths

wesm:
This will be a better out-of-the-box experience for users who have Thrift installed through some other means, possibly without being built with -fPIC for static linking.


wesm:
Shoot, this will fail on OS X because we are doing `brew install thrift` in the Travis CI scripts. I suggest that we merge #251 and build the external project in Travis CI for OS X also.

wesm:
Merging this, so that the fix for PARQUET-849 can be tested

rdblue:
This works for me. Sorry for the delay!

PR #257, opened 2017-02-22
PARQUET-889: Fix compilation when SSE is enabled

daedric:
bit-util.h was missing a couple of include. cpu-util.h was including
parquet/logging.h which is an internal header so the implementations of
the static functions have been moved in the source file.

### DIFF
--- a/src/parquet/util/cpu-info.h
+++ b/src/parquet/util/cpu-info.h
@@ -24,8 +24,6 @@
 #include <cstdint>
 #include <string>
 
-#include "parquet/util/logging.h"
-
 namespace parquet {
 
 /// CpuInfo is an interface to query for cpu information at runtime.  The caller can
@@ -54,14 +52,10 @@ class CpuInfo {
   static void VerifyCpuRequirements();
 
   /// Returns all the flags for this cpu
-  static int64_t hardware_flags() {
-    DCHECK(initialized_);
-    return hardware_flags_;
-  }
+  static int64_t hardware_flags();
 
   /// Returns whether of not the cpu supports this flag
   inline static bool IsSupported(int64_t flag) {
-    DCHECK(initialized_);
     return (hardware_flags_ & flag) != 0;
   }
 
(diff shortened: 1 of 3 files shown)


src/parquet/util/cpu-info.h
```diff
@@ -54,44 +52,26 @@ class CpuInfo {
...
-    return hardware_flags_;
-  }
+  static int64_t hardware_flags();
 
   /// Returns whether of not the cpu supports this flag
-  inline static bool IsSupported(int64_t flag) {
-    DCHECK(initialized_);
-    return (hardware_flags_ & flag) != 0;
-  }
+  static bool IsSupported(int64_t flag);
```
wesm: This is the one function that really needs to be inlined

daedric: how would you suggest we keep it in the header ?
Here are the solution I can think of in order of preference:

* manual if + throw (or abort)
* Simply remove the DCHECK
* Make the logging.h a public header

wesm: I would say to remove the DCHECK, @majetideepak @xhochy do you have an opinion? whether the `CpuInfo` is initialized is something we can test elsewhere

xhochy: Away with it!

daedric: Fine by me, let me remove it :)

majetideepak: I am fine with removing it as well.

daedric: The fix has been pushed, I guess one can merge it as soon as the CI finishes :)

wesm:
I'm keeping an eye on it. If you enable Travis CI for your fork then we can get faster builds (right now we're in the general ASAF queue, which can sometimes get congested)

wesm:
The build passed https://travis-ci.org/daedric/parquet-cpp/builds/204262490 -- the Linux failures are because of coveralls (we should make coveralls only run on the main repo maybe). Merging

wesm:
Thanks for your contribution!

daedric:
My pleasure :)

PR #258, opened 2017-02-22
PARQUET-891: Do not search for snappy on system paths

wesm:
The `libsnappy.a` from Linux package mangers (e.g. Ubuntu 14.04) may not be built with -fPIC, which will cause a linker failure when creating `libparquet.so`. Add notes to README about this nuance when depending on separately built/installed thirdparty components.


wesm:
Build pending: https://travis-ci.org/wesm/parquet-cpp/builds/204314886

PR #259, opened 2017-02-22
PARQUET-892: Specify public link targets for parquet_static so that transitive dependencies are linked in executables

wesm:


wesm:
@keithgchapman can you try out this branch on your environment? Thanks

keithgchapman:
Sure will do and report back

keithgchapman:
@wesm I'm still getting issues, will post the transcript with VERBOSE=1

wesm:
@keithgchapman does /usr/lib/x86_64-linux-gnu/libboost_regex.so exist on your system?

keithgchapman:
@wesm Yes

ldconfig -p | grep libboost_regex
	libboost_regex.so.1.58.0 (libc6,x86-64) => /usr/lib/x86_64-linux-gnu/libboost_regex.so.1.58.0
	libboost_regex.so (libc6,x86-64) => /usr/lib/x86_64-linux-gnu/libboost_regex.so

wesm:
It looks like there is something wrong with your Boost header files -- possibly from having multiple versions of Boost installed. So the headers in `/usr/include/boost` are not consistent with the compiled symbols in `libboost_regex.so`. 

Can you try removing and reinstalling your system Boost and see if that fixes the problem?

keithgchapman:
Sure, will give that a shot

keithgchapman:
@wesm I tried removing boost and reinstalling it, it still gave the sane error. Inorder to do a sanity check I tried building parquet-cpp on a different machine (again ubuntu 16.04 with boost 1.58) and I get the same error. Here is a log from the other machine [1]. Is it something to do with boost 1.58? I've run into similar issues before where it works with Ubuntu 14.04 (which has boost 1.54) but fails on 16.04 (which has boost 1.58).

[1] https://gist.github.com/keithgchapman/473add1d345174a9017522a16d7112a0

majetideepak:
I think I am hitting a similar issue with boost 1.59 as well
undefined symbol: _ZN5boost9re_detail14verify_optionsEjNS_15regex_constants12_match_flagsE

majetideepak:
The library actually builds fine with boost 1.59. I did not hit the above error. Sorry for the confusion.

wesm:
I recommend building your own Boost, also to be able to statically link the required symbols. I'll set up a 16.04 VM so I can see if it's fixable.

wesm:
@keithgchapman I figured out the problem. The default gcc ABI on Ubuntu Xenial is gcc 5.x, which makes C++ libraries coming from apt repos incompatible with the libstdc++ for gcc 4.9. So you need to build your own boost to use gcc 4.9

wesm:
@xhochy if this and #258 look OK, I think we can do an RC after merging them.

green build for this patch: https://travis-ci.org/wesm/parquet-cpp/builds/204363838

keithgchapman:
Thanks for the update @wesm and thank you very much for looking into it.

PR #261, opened 2017-02-23
PARQUET-895: Fix broken reading of nested repeated columns

mvertes:
Use the column metadata to get the number of levels to process. It
fixes a problem where not enough values are returned in the case
of nested repeated values.

@peshopetrov should get the credit for 1st attempt #250

### DIFF
--- a/src/parquet/file/reader-internal.cc
+++ b/src/parquet/file/reader-internal.cc
@@ -196,7 +196,7 @@ std::unique_ptr<PageReader> SerializedRowGroup::GetColumnPageReader(int i) {
   stream = properties_.GetStream(source_, col_start, col_length);
 
   return std::unique_ptr<PageReader>(new SerializedPageReader(std::move(stream),
-      row_group_metadata_->num_rows(), col->compression(), properties_.memory_pool()));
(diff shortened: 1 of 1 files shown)


mvertes:
I have a file, but it is quite large, complex and with confidential data. I'm not able to generate quickly a sample.

The sample should have a schema with nested columns, for ex (simplified):

```
message hive_schema {
  optional binary odid (UTF8);
  optional group queries (LIST) {
    repeated group bag {
      optional binary array_element (UTF8);
    }
  }
}
```

and the array_element column should have  many more values than there is rows

xhochy:
Pushed the commit to my repo so we hopefully get a faster Travis build: https://travis-ci.org/xhochy/parquet-cpp/builds/204592293

wesm:
Merging in interest of getting an RC out

wesm:
Cherry-picked in https://github.com/apache/parquet-cpp/commit/98f5fa144f7d1e219c37ebc56e69bc0543457d78 (I edited the commit message to close #250 but forgot to close this PR). Can you close? Thanks

PR #264, opened 2017-02-28
PARQUET-901: Publish RCs in apache-parquet-cpp-VERSION in SVN

xhochy:


xhochy:
Green build: https://travis-ci.org/xhochy/parquet-cpp/builds/207314819

PR #267, opened 2017-03-08
PARQUET-908: Fix shared library visibility of some symbols in types.h

wesm:


wesm:
here's the build from my branch, will keep an eye on https://travis-ci.org/wesm/parquet-cpp/builds/209006672

wesm:
merging, as this will also fix the arrow builds

PR #268, opened 2017-03-10
PARQUET-909: Reduce buffer allocations (mallocs) on critical path

majetideepak:


majetideepak:
@wesm @xhochy, The plan here is to avoid some of the mallocs in the write path. The current implementation uses `AllocateBuffer` followed by `Resize` very frequently. I made some changes to the `column/writer.cc`.  I expect similar changes in  `file/writer-internal.cc` in next patches.
I wanted to know your concerns if any before I proceed in this path. Thanks!

src/parquet/column/writer.cc
```diff
@@ -72,51 +75,55 @@ void ColumnWriter::WriteRepetitionLevels(int64_t num_levels, const int16_t* leve
...
   int64_t encoded_size = level_encoder_.len() + sizeof(int32_t);
-  DCHECK(rle_size >= encoded_size);
-  PARQUET_THROW_NOT_OK(buffer_rle->Resize(encoded_size));
-  return std::static_pointer_cast<Buffer>(buffer_rle);
+  return encoded_size;
 }
 
 void ColumnWriter::AddDataPage() {
-  std::shared_ptr<Buffer> definition_levels = definition_levels_sink_->GetBuffer();
-  std::shared_ptr<Buffer> repetition_levels = repetition_levels_sink_->GetBuffer();
```
majetideepak: These previously allocated/freed buffers.

src/parquet/column/writer.cc
```diff
@@ -72,51 +75,55 @@ void ColumnWriter::WriteRepetitionLevels(int64_t num_levels, const int16_t* leve
       reinterpret_cast<const uint8_t*>(levels), sizeof(int16_t) * num_levels);
 }
 
-std::shared_ptr<Buffer> ColumnWriter::RleEncodeLevels(
-    const std::shared_ptr<Buffer>& buffer, int16_t max_level) {
+int64_t ColumnWriter::RleEncodeLevels(std::shared_ptr<ResizableBuffer> dest_buffer,
+    const uint8_t* src_buffer, int16_t max_level) {
```
xhochy: You can pass `dest_buffer` as `ResizableBuffer&` here.

src/parquet/column/writer.cc
```diff
@@ -72,51 +75,55 @@ void ColumnWriter::WriteRepetitionLevels(int64_t num_levels, const int16_t* leve
...
+  int64_t definition_levels_rle_size = 0;
+  int64_t repetition_levels_rle_size = 0;
+
   std::shared_ptr<Buffer> values = GetValuesBuffer();
 
   if (descr_->max_definition_level() > 0) {
-    definition_levels =
-        RleEncodeLevels(definition_levels, descr_->max_definition_level());
+    definition_levels_rle_size =
+        RleEncodeLevels(std::static_pointer_cast<ResizableBuffer>(definition_levels_rle_),
```
xhochy: I guess we should store `definition_levels_rle_` from the beginning as `std::shared_ptr<ResizableBuffer>`

majetideepak:
@wesm @xhochy This patch is ready. Let me know your feedback. Thanks.

majetideepak:
I tested the changes using Callgrind. 
The total number of calls to `Arrow::InternalMemoryPool::Reallocate` and `Arrow::InternalMemoryPool::Allocate` reduced to `1/4` of the original number.

src/parquet/column/writer.cc
```diff
@@ -72,68 +83,89 @@ void ColumnWriter::WriteRepetitionLevels(int64_t num_levels, const int16_t* leve
...
   ResetPageStatistics();
-  std::shared_ptr<Buffer> compressed_data = pager_->Compress(uncompressed_data);
-  CompressedDataPage page(compressed_data, num_buffered_values_, encoding_, Encoding::RLE,
-      Encoding::RLE, uncompressed_size, page_stats);
+
+  std::shared_ptr<Buffer> compressed_data;
+  if (pager_->has_compressor()) {
+    // Use Arrow::Buffer::shrink_to_fit = false
+    // underlying buffer only keeps growing. Resize to a smaller size does not reallocate.
+    PARQUET_THROW_NOT_OK(compressed_data_->Resize(uncompressed_size, false));
```
xhochy: This is a duplicate resize as we will also call `Resize` inside of `Compress`

src/parquet/column/writer.cc
```diff
@@ -72,68 +83,89 @@ void ColumnWriter::WriteRepetitionLevels(int64_t num_levels, const int16_t* leve
...
+    compressed_data = compressed_data_;
+  } else {
+    compressed_data = uncompressed_data_;
+  }
 
   // Write the page to OutputStream eagerly if there is no dictionary or
   // if dictionary encoding has fallen back to PLAIN
   if (has_dictionary_ && !fallback_) {  // Save pages until end of dictionary encoding
+    std::shared_ptr<Buffer> compressed_data_copy;
+    compressed_data->Copy(0, compressed_data->size(), allocator_, &compressed_data_copy);
```
wesm: can this fail?

src/parquet/file/writer-internal.cc
```diff
@@ -70,22 +70,21 @@ void SerializedPageWriter::Close(bool has_dictionary, bool fallback) {
   metadata_->WriteTo(sink_);
 }
 
-std::shared_ptr<Buffer> SerializedPageWriter::Compress(
-    const std::shared_ptr<Buffer>& buffer) {
-  // Fast path, no compressor available.
-  if (!compressor_) return buffer;
+void SerializedPageWriter::Compress(const std::shared_ptr<Buffer>& src_buffer,
+    std::shared_ptr<ResizableBuffer>& dest_buffer) {
```
wesm: We might get into a habit of passing the immutable and mutable args here instead as:

`const Buffer& src_buffer, ResizableBuffer* dest_buffer`

src/parquet/util/memory.h
```diff
@@ -340,6 +340,12 @@ class PARQUET_EXPORT InMemoryOutputStream : public OutputStream {
 
   virtual void Write(const uint8_t* data, int64_t length);
 
+  // Clears the stream
+  void Clear() { size_ = 0; }
+
+  // Get pointer to the underlying buffer
+  std::shared_ptr<Buffer> GetBufferRef() const { return buffer_; }
```
wesm: Maybe return `const Buffer&` (may require the change I described above) unless you expect the caller to need to share ownership over the returned object?

src/parquet/column/writer.cc
majetideepak: Good catch. Add a check.

src/parquet/file/writer-internal.cc
majetideepak: Will fix the API

src/parquet/util/memory.h
majetideepak: I don't require the caller to share ownership in my current patch. Will change it.

src/parquet/column/writer.cc
majetideepak: good catch. Will fix it.

majetideepak:
Thank you for the feedback.

PR #270, opened 2017-03-13
PARQUET-915: Support Arrow Time Types

tebeka:


wesm:
Can you 

* add a unit test for this (similar to https://github.com/apache/parquet-cpp/blob/master/src/parquet/arrow/arrow-schema-test.cc#L97)
* open a Parquet JIRA for time schema conversion and change the PR title? 

Thanks

src/parquet/arrow/reader.cc
```diff
@@ -903,7 +924,22 @@ Status ColumnReader::Impl::NextBatch(int batch_size, std::shared_ptr<Array>* out
...
+          static_cast<::arrow::TimeType*>(field_->type.get());
+      switch (time_type->unit) {
+        case ::arrow::TimeUnit::MILLI:
+          return TypedReadBatch<::arrow::TimeType, Int32Type>(batch_size, out);
+          break;
+        case ::arrow::TimeUnit::NANO:
+          return TypedReadBatch<::arrow::TimeType, Int64Type>(batch_size, out);
+          break;
+        default:
+          return UnsupportedTimeUnit(time_type->unit);
```
wesm: Since we don't have any tests for reading time columns, can you leave this case not implemented?

tebeka: Working on adding tests ...

wesm: The time type is undergoing discussion in https://issues.apache.org/jira/browse/ARROW-617 so most likely these will have to be revisited after the time type is finalized in the Arrow format.

tebeka: OK. Will remove this code then :)

PR #271, opened 2017-03-20
PARQUET-897: Only use designated public headers from libarrow

xhochy:


xhochy:
Build succeeded, coverage upload is flaky.

PR #272, opened 2017-03-23
PARQUET-919: Account for ARROW-683 changes, but make no functional changes. Set PARQUET_ARROW=on by default

wesm:


wesm:
Bumped the Arrow version

wesm:
green build: https://travis-ci.org/wesm/parquet-cpp

PR #273, opened 2017-03-24
PARQUET-923: Account for Time type changes in Arrow

wesm:
Needs ARROW-709 (https://github.com/apache/arrow/pull/439), will update Arrow commit hash once that's done


wesm:
Yes, we need to make some follow up patches to test the read/writes. I think it's acceptable to expect implicit coercion to milliseconds or microseconds

PR #274, opened 2017-03-26
PARQUET-925: Use detected ARROW_LIBS from ARROW_LIB_PATH

xhochy:


cmake_modules/FindArrow.cmake
```diff
@@ -56,7 +56,7 @@ if (ARROW_INCLUDE_DIR AND ARROW_LIB_PATH)
   set(ARROW_LIB_NAME libarrow)
   set(ARROW_IO_LIB_NAME libarrow_io)
 
-  set(ARROW_LIBS ${ARROW_SEARCH_LIB_PATH})
```
xhochy: The magic here is that CMake also tries all paths for libraries appended with the platform bit-width.

PR #275, opened 2017-03-27
PARQUET-918: FromParquetSchema API crashes on nested schemas

itaiin:
FromParquetSchema and ReadTable APIs crashed on nested schemas parquet on partial schema reads.

The fix creates a partial schema while building the arrow::Schema object.

4 tests were added - Two for FromParquetSchema API and two for ReadTable, each function is tested with a full schema read and a partial one from a simple nested parquet file.

### DIFF
--- a/src/parquet/arrow/schema.cc
+++ b/src/parquet/arrow/schema.cc
@@ -188,6 +188,11 @@ Status FromPrimitive(const PrimitiveNode* primitive, TypePtr* out) {
 Status NodeToFieldInternal(const NodePtr& node,
     const std::unordered_set<NodePtr>* needed_nodes, std::shared_ptr<Field>* out);
 
+bool IsNeededNode(const NodePtr& node, const std::unordered_set<NodePtr>* needed_nodes) {
+  return ((needed_nodes == nullptr) ||
+          (needed_nodes->count(node) > 0));
+}
+
 Status StructFromGroup(const GroupNode* group,
     const std::unordered_set<NodePtr>* needed_nodes, TypePtr* out) {
   std::vector<std::shared_ptr<Field>> fields;
@@ -227,7 +232,8 @@ Status NodeToList(const GroupNode* group,
           !str_endswith_tuple(list_node->name())) {
         // List of primitive type
         std::shared_ptr<Field> item_field;
-        RETURN_NOT_OK(NodeToFieldInternal(list_group->field(0), needed_nodes, &item_field));
+        RETURN_NOT_OK(NodeToFieldInternal(
+          list_group->field(0), needed_nodes, &item_field));
         if (item_field != nullptr) {
           *out = ::arrow::list(item_field);
         }
@@ -243,9 +249,9 @@ Status NodeToList(const GroupNode* group,
     } else if (list_node->is_repeated()) {
       // repeated primitive node
       std::shared_ptr<::arrow::DataType> inner_type;
-      PrimitiveNode* primitive = static_cast<PrimitiveNode*>(list_node.get());
-      if ((needed_nodes == nullptr) ||
-          (needed_nodes->count(static_cast<NodePtr>(primitive)) > 0)) {
+      if (IsNeededNode(static_cast<NodePtr>(list_node), needed_nodes)) {
+        const PrimitiveNode* primitive =
+          static_cast<const PrimitiveNode*>(list_node.get());
         RETURN_NOT_OK(FromPrimitive(primitive, &inner_type));
         auto item_field = std::make_shared<Field>(list_node->name(), inner_type, false);
         *out = ::arrow::list(item_field);
@@ -276,8 +282,7 @@ Status NodeToFieldInternal(const NodePtr& node,
   if (node->is_repeated()) {
     // 1-level LIST encoding fields are required
     std::shared_ptr<::arrow::DataType> inner_type;
-    if ((needed_nodes == nullptr) ||
-        (needed_nodes->count(static_cast<NodePtr>(node)) > 0)) {
+    if (IsNeededNode(static_cast<NodePtr>(node), needed_nodes)) {
       const PrimitiveNode* primitive = static_cast<const PrimitiveNode*>(node.get());
       RETURN_NOT_OK(FromPrimitive(primitive, &inner_type));
       auto item_field = std::make_shared<Field>(node->name(), inner_type, false);
@@ -293,8 +298,7 @@ Status NodeToFieldInternal(const NodePtr& node,
     }
   } else {
     // Primitive (leaf) node
-    if ((needed_nodes == nullptr) ||
-        (needed_nodes->count(static_cast<NodePtr>(node)) > 0)) {
+    if (IsNeededNode(static_cast<NodePtr>(node), needed_nodes)) {
       const PrimitiveNode* primitive = static_cast<const PrimitiveNode*>(node.get());
       RETURN_NOT_OK(FromPrimitive(primitive, &type));
     }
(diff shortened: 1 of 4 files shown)


src/parquet/arrow/schema.cc
```diff
@@ -183,12 +184,31 @@ Status FromPrimitive(const PrimitiveNode* primitive, TypePtr* out) {
   return Status::OK();
 }
 
-Status StructFromGroup(const GroupNode* group, TypePtr* out) {
-  std::vector<std::shared_ptr<Field>> fields(group->field_count());
+// Forward declaration
+Status NodeToFieldInternal(const NodePtr& node,
+    const std::unordered_set<NodePtr>* needed_nodes, std::shared_ptr<Field>* out);
+
+bool IsNeededNode(const NodePtr& node, const std::unordered_set<NodePtr>* needed_nodes) {
```
wesm: What is a needed node?

itaiin: A needed node is a parquet schema leaf node which is required in the partial schema. 
So for example IsNeededNode is a shortcut used later to test if we want the parquet leaf node in the arrow schema so should maybe be an inline function or even a macro.
Maybe I should rename it to partial schema leaf node or something similar? 

About the tests - no problem, but I also want to test ReadTable, which needs a parquet file and not just the parquet schema. Is there a test suite that construct one?

itaiin:
Note that new tests correctness rely on nested fields comparison being correct
I have found a bug in nested comparisons and submitted a fix to the arrow project - https://github.com/apache/arrow/pull/450

wesm:
See https://github.com/apache/parquet-cpp/blob/master/src/parquet/arrow/arrow-reader-writer-test.cc#L779. I think if you do a variant of that, but you'll need to use the lower-level file writer API to write the nested data, then that is good.

itaiin:
I have no idea why the last tiny change failed the CI check

wesm:
Can you rebase? An API change in Arrow was just fixed in https://github.com/apache/parquet-cpp/commit/a48bfaa7e9c649b120eb3c88e0234695042d5d4e

wesm:
Hm. not sure what just happened

wesm:
Looks like you rebased on something other than apache/parquet-cpp master branch

src/parquet/arrow/schema.cc
```diff
@@ -238,40 +277,54 @@ Status NodeToList(const GroupNode* group, TypePtr* out) {
...
   if (node->is_repeated()) {
     // 1-level LIST encoding fields are required
     std::shared_ptr<::arrow::DataType> inner_type;
-    const PrimitiveNode* primitive = static_cast<const PrimitiveNode*>(node.get());
-    RETURN_NOT_OK(FromPrimitive(primitive, &inner_type));
-    auto item_field = std::make_shared<Field>(node->name(), inner_type, false);
-    type = ::arrow::list(item_field);
-    nullable = false;
+    if (IsIncludedLeaf(static_cast<NodePtr>(node), included_leaf_nodes)) {
+      const PrimitiveNode* primitive = static_cast<const PrimitiveNode*>(node.get());
```
advancedxy: Hi,  I think this doesn't handles the repeated group field. For example,
```
message TestProtobuf.ArrayWithNestedGroupAndArray {
  optional int32 primitive;
  repeated group myComplex {
    optional int32 id;
    repeated group repeatedMessage {
      optional int32 someId;
    }
  }
}
```
So, I made an attempt to fix that issue, but didn't report earlier. The fix is shown as below against old code.
``` cpp
  if (node->is_repeated()) {
    // 1-level LIST encoding fields are required
    std::shared_ptr<::arrow::DataType> inner_type;
    if (node->is_group()) {
      const GroupNode* group = static_cast<const GroupNode*>(node.get());
      RETURN_NOT_OK(StructFromGroup(group, &inner_type));
    } else {
      const PrimitiveNode* primitive = static_cast<const PrimitiveNode*>(node.get());
      RETURN_NOT_OK(FromPrimitive(primitive, &inner_type));
    }
    auto item_field = std::make_shared<Field>(node->name(), inner_type, false);
    type = ::arrow::list(item_field);
    nullable = false;
```
Hope this helps.

wesm:
I will review in more detail sometime this weekend. Haven't looked into the issue that @advancedxy mentioned

itaiin:
Seems to me the issue @advancedxy mentioned is not related the original one but since its in the same code paths its possible to fix both in the same PR

advancedxy:
@itaiin yes, it's not in the selected column code path, but it do relate to nested schema in the FromParquetSchema api.  I believe it's better to fix both in the same pr.

wesm:
@advancedxy @itaiin sorry for my slow response. This PR looks good. Since I'm looking at this today let me rebase and pull in the @advancedxy fix and see if I can write a test case for it

wesm:
I didn't get to this today. @itaiin if you have time tomorrow to update this patch, could you do that? otherwise I can try to pick it up in the next couple days

itaiin:
I already have a patch for @advancedxy issue and a test case (his own example).
I'll add them and rebase.

itaiin:
Had a bit of trouble with the rebase and merge again - hope I did it right

wesm:
Doesn't look right, includes a bunch of other patches

wesm:
When you are rebasing, the only commands you should be using are

git rebase BRANCH

<resolve conflicts>
git add ...
git rebase --continue

Using "pull" or "merge" at any point will put you in a broken state

wesm:
You can try using `git merge --squash` to get a clean commit including only your changes

itaiin:
umm yes thats what I did, however it refused to accept them, I guess the last commit screwed everything. i'll try to fix it

wesm:
Looking good now =)

itaiin:
No prob. sorry for the git problems.

itaiin:
Hmm.. travis failed... that's odd everything builds on my ubuntu machine

wesm:
Looks like there's something wrong with the Arrow version pinning (it's building against master)

wesm:
I fixed a small API break that was just merged to Arrow, here is my rebased branch with the fix if you can update this PR https://github.com/wesm/parquet-cpp/tree/PARQUET-918 -- sorry about the disruption

itaiin:
Sorry I wasn't available. Thanks for merging the PR.

src/parquet/arrow/schema.cc
```diff
@@ -285,11 +343,22 @@ Status FromParquetSchema(const SchemaDescriptor* parquet_schema,
...
+    auto column_desc = parquet_schema->Column(column_indices[i]);
+    included_leaf_nodes.insert(column_desc->schema_node());
+  }
 
-  std::vector<std::shared_ptr<Field>> fields(num_fields);
-  for (int i = 0; i < num_fields; i++) {
-    RETURN_NOT_OK(NodeToField(schema_node->field(column_indices[i]), &fields[i]));
+  std::vector<std::shared_ptr<Field>> fields;
+  std::shared_ptr<Field> field;
+  for (int i = 0; i < schema_node->field_count(); i++) {
```
advancedxy: Later to party. But I think this breaks the column_indices ordering with old behaviour...

Let's we have three simple columns(int32, boolean, string), and the column indices is [1, 0].
then the converted schema is {int32, bool} instead of {bool, int32}.

@itaiin do you have time to fix this issue? If not, I will try to send a pr.
cc @wesm

wesm: @advancedxy can you send a PR?

itaiin: @advancedxy Sorry I'm working on too many things right now... would be very happy too if you send a PR
(@wesm)

advancedxy: I will send a PR this weekend.  I am pretty busy on work days.

PR #276, opened 2017-03-27
PARQUET-926: Use pkg-config to find Apache Arrow

kou:
pkg-config is a tool to detect package. It's suitable for this propose.

If we can't find Apache Arrow by pkg-config, we use the current Apache
Arrow detect system as fallback.


cmake_modules/ThirdpartyToolchain.cmake
```diff
@@ -348,7 +348,23 @@ if(PARQUET_BUILD_BENCHMARKS)
...
 ## Apache Arrow
-find_package(Arrow)
+pkg_check_modules(ARROW arrow)
+pkg_check_modules(ARROW_IO arrow-io)
+if (ARROW_FOUND AND ARROW_IO_FOUND)
+  set(ARROW_INCLUDE_DIR ${ARROW_INCLUDE_DIRS})
+  set(ARROW_LIBS ${ARROW_LIBRARY_DIRS})
+
+  set(ARROW_LIB_NAME libarrow)
+  set(ARROW_IO_LIB_NAME libarrow_io)
```
wesm: Is there a way to make this work for MSVC (no lib prefix)? We aren't supporting visual studio yet, but hopefully soon

kou: Yes.
[`pkg_check_modules`](https://cmake.org/cmake/help/latest/module/FindPkgConfig.html#command:pkg_check_modules) provides `ARROW_LDFLAGS`. It includes all needed for each platform.

But pkg-config isn't used for MSVC normally.

So it's better that we use `link_directories()` and [`target_link_libraries()`](https://cmake.org/cmake/help/latest/command/target_link_libraries.html#command:target_link_libraries) with `arrow` (no `lib` prefix and no `.lib` suffix) instead of `IMPORTED_LOCATION`. `target_link_libraries()` cares about real linker flags (`-larrow` or `arrow.lib`).

wesm: Got it, thank you

xhochy: You could use https://cmake.org/cmake/help/v3.0/variable/CMAKE_SHARED_LIBRARY_PREFIX.html for that, just for reference, no need for it here (yet).

kou: OK. I changed to use `CMAKE_SHARED_LIBRARY_PREFIX`.
I borrowed the code from `FindArrow.cmake`. We may need to do the same change in `FindArrow.cmake` later.

xhochy:
@kou Is there a possibility to determine the SOVERSION via pkg-config? 

I'm currently looking into how to determining inside of CMake all files & symlinks that belong to `libarrow.so.0.0.0` and I would like to avoid using calls to `readlink`.

kou:
> @kou Is there a possibility to determine the SOVERSION via pkg-config?

Yes.
We can put custom variables to `.pc` file. If we put a variable for `SOVERSION` to `arrow.pc`, we can determine `SOVERSION` from `arrow.pc`.

If you like this approach, I can create a pull request.

xhochy:
Yes, that would be really nice.

PR #279, opened 2017-03-30
PARQUET-932: Add option to build parquet library with minimal dependency

majetideepak:
If users want to build only the parquet library, thirdparty headers are sufficient and GTEST and GBENCHMARK are not required.


CMakeLists.txt
```diff
@@ -103,10 +103,14 @@ if ("${CMAKE_SOURCE_DIR}" STREQUAL "${CMAKE_CURRENT_SOURCE_DIR}")
   option(PARQUET_RPATH_ORIGIN
     "Build Parquet libraries with RATH set to \$ORIGIN"
     OFF)
+  option(PARQUET_MINIMAL_DEPENDENCY
+    "Depend only on Thirdparty headers to build libparquet. Always OFF if building binaries"
+    OFF)
```
majetideepak: We could also implicitly set this value to ``OFF`` by default and set it to `ON` if `PARQUET_BUILD_TESTS` and `PARQUET_BUILD_EXECUTABLES` and `PARQUET_BUILD_BENCHMARKS` are all `OFF`

majetideepak:
@xhochy, @wesm  Can you give your feedback on this? I would like this patch to be committed soon.Thanks!

cmake_modules/FindArrow.cmake
```diff
@@ -51,8 +51,10 @@ find_library(ARROW_IO_LIB_PATH NAMES arrow_io
   ${ARROW_SEARCH_LIB_PATH}
   NO_DEFAULT_PATH)
 
-if (ARROW_INCLUDE_DIR AND ARROW_LIB_PATH)
+if (ARROW_INCLUDE_DIR AND (PARQUET_MINIMAL_DEPENDENCY OR ARROW_LIB_PATH))
   set(ARROW_FOUND TRUE)
+  set(ARROW_HEADER_NAME arrow/arrow.h)
```
wesm: It's called `arrow/api.h`

cmake_modules/FindArrow.cmake
```diff
@@ -51,8 +51,10 @@ find_library(ARROW_IO_LIB_PATH NAMES arrow_io
   ${ARROW_SEARCH_LIB_PATH}
   NO_DEFAULT_PATH)
 
-if (ARROW_INCLUDE_DIR AND ARROW_LIB_PATH)
+if (ARROW_INCLUDE_DIR AND (PARQUET_MINIMAL_DEPENDENCY OR ARROW_LIB_PATH))
```
wesm: Just so I understand, you have a build environment where only the headers are available, and no libraries?

majetideepak: That is correct.

cmake_modules/FindArrow.cmake
majetideepak: Fixed.

wesm:
+1, merging this

PR #282, opened 2017-03-31
PARQUET-935: Set version to shared library

kou:
.deb package builder assumes that shared libraries have version.
See also: https://www.debian.org/doc/debian-policy/ch-sharedlibs.html


wesm:
Oops, can you rebase? thanks

kou:
Done!

PR #291, opened 2017-04-05
PARQUET-946: Add ReadRowGroup and num_row_group methods to arrow::FileReader

wesm:
There's a lot of room for improvement / further refactoring here -- the assumption that an entire column in a file is being read runs very deep in the Arrow reader, so I tried to do the minimum work to decouple the row group iteration. There's some code duplication in ReadRowGroup, but we should maybe save further cleanup for a future patch.


xhochy:
The size of code duplications is far less than I would have expected ;)

PR #292, opened 2017-04-09
PARQUET-947: Account for Arrow library consolidation in ARROW-795, API changes in ARROW-782

wesm:


wesm:
@majetideepak Uwe is out this week, does this look OK to you?

wesm:
Updated the version to get a passing build

wesm:
I'm going to cherry-pick #293 to try to get a passing build

majetideepak:
+1 LGTM. @wesm on a seperate note, does arrow produce only one library now? Can you tell me the reason for this change? Thanks!

wesm:
@majetideepak yes, it produces only one library. It's mainly to simplify development for third parties -- if you only need some Arrow symbols, then you can link to libarrow.a, but otherwise you can link to the single `libarrow.so`. Here we were already linking against both `libarrow.so` and `libarrow_io.so` (but not `libarrow_ipc.so`). The only additional piece in the toolchain from `libarrow_ipc` is Flatbuffers, which isn't a very intrusive dependency. If users have problems with this, we could add an option to leave the arrow/ipc pieces out of the build.

majetideepak:
Adding option to leave `arrow/ipc` pieces out of the build sounds good.
For only Parquet Library users without ARROW support, this will be helpful.
I will add an ARROW JIRA.

PR #294, opened 2017-04-10
PARQUET-949: Pin Arrow version using git archives instead of GIT_TAG

wesm:
There has been some flakiness with this on PRs today, so this should help prevent accidentally building against master


majetideepak:
+1 LGTM

wesm:
There was some oddness where the version string was being dropped, I guess that was the issue before.

PR #296, opened 2017-04-13
PARQUET-953: Add static constructors to arrow::FileWriter for initializing from schema, add WriteTable method

wesm:
I preserved the existing WriteTable top level methods, but this will unblock ARROW-528


wesm:
I switched to clang-format-3.9 which has some minor changes; I guess we should document which clang-format is the "official" one. Hopefully it will stop changing things

PR #298, opened 2017-04-15
PARQUET-955: Only look for Arrow pkgconfig if ARROW_HOME is not set in environment

wesm:
cc @kou 

Ran into this when testing out https://github.com/apache/arrow/pull/548


cmake_modules/ThirdpartyToolchain.cmake
```diff
@@ -357,7 +357,14 @@ if(PARQUET_BUILD_BENCHMARKS AND NOT IGNORE_OPTIONAL_PACKAGES)
 endif()
 
 ## Apache Arrow
-pkg_check_modules(ARROW arrow)
+if ("$ENV{ARROW_HOME}" STREQUAL "")
+  # PARQUET-955. If the user has set $ARROW_HOME in the environment, we respect
+  # this, otherwise try to locate the pkgconfig in the system environment
+  pkg_check_modules(ARROW arrow)
```
majetideepak: Not related to this bug, but should this be moved to `FindArrow.cmake`?

wesm: Good question for @kou, I am not sure

kou:
I think that it's better that users should use `PKG_CONFIG_PATH` instead of `ARROW_HOME` when they can use it. (For example, Windows users will use `ARROW_HOME` instead of `PKG_CONFIG_PATH`.) Because `PKG_CONFIG_PATH` is more general way than `ARROW_HOME`.

But `ARROW_HOME` is respected than `PKG_CONFIG_PATH` is reasonable. Because `ARROW_HOME` is more specific than `PKG_CONFIG_PATH`. So +1.

cmake_modules/ThirdpartyToolchain.cmake
kou: Yes. We should move to `FindArrow.cmake`.

majetideepak:
I ran into this today. The build system did not honor `ARROW_HOME`  and was picking old arrow headers causing a very weird segmentation fault. I would like to print the include path along with the library path that gets printed. I would also like move this to `FindArrow.cmake`. I can open a separate JIRA to do this if it is not in the scope of this JIRA.

wesm:
Just did both, could you both take a look and see if things look reasonable?

cmake_modules/FindArrow.cmake
```diff
@@ -18,52 +18,82 @@
...
-  ${ARROW_SEARCH_HEADER_PATHS}
-  # make sure we don't accidentally pick up a different version
-  NO_DEFAULT_PATH
-)
+  set(ARROW_LIB_NAME ${CMAKE_SHARED_LIBRARY_PREFIX}arrow)
+
+  set(ARROW_SHARED_LIB ${ARROW_LIBDIR}/${ARROW_LIB_NAME}${CMAKE_SHARED_LIBRARY_SUFFIX}${ARROW_SHARED_LIB_SUFFIX})
+  set(ARROW_STATIC_LIB ${ARROW_LIBDIR}/${ARROW_LIB_NAME}${CMAKE_STATIC_LIBRARY_SUFFIX})
+else()
+  if( NOT "$ENV{ARROW_HOME}" STREQUAL "")
```
majetideepak: Do we need this check again here? if `ARROW_HOME` is set in the env, `ARROW_FOUND` is always false right?

wesm: I moved the `set(ARROW_HOME ...)` to the top

majetideepak:
+1 LGTM.  I tried this patch locally and it does seem to honor `ARROW_HOME` if set.
Thanks @wesm

xhochy:
@wesm Can you also update this for `pyarrow` so that the `FindArrow` scripts don't get out-of-sync (which they probably are already)?

wesm:
yep

PR #300, opened 2017-04-18
PARQUET-931: Add option to pin Arrow version with environment variable

wesm:
Also added console output to show the version to help with debugging. I need this to help with pinning the Arrow version in conda-forge without necessarily making Arrow itself a build dependency (since it's cheap to build Arrow)


majetideepak:
+1

PR #303, opened 2017-04-20
PARQUET-959: Set CMAKE_INSTALL_LIBDIR in Arrow thirdparty build to avoid failures on multiarch systems

wesm:
Arrow libraries are now installed in `lib64` on multiarch systems, which breaks the thirdparty build


wesm:
cc @kou

kou:
+1!

wesm:
thanks!

PR #304, opened 2017-04-20
PARQUET-957: Add PARQUET_BUILD_TOOLCHAIN env variable option to simplify setting build environment

wesm:
See also ARROW-849


majetideepak:
+1 LGTM

PR #307, opened 2017-04-20
PARQUET-508: Add ParquetFilePrinter

majetideepak:


majetideepak:
@wesm, @xhochy  I am planning on adding `ParquetFilePrinter::JsonPrint()` as part of PARQUET-958. Do you have any feedback on this API? Thanks.

PR #308, opened 2017-04-21
PARQUET-963: Return NotImplemented when attempting to read a struct field

wesm:
cc @xhochy @itaiin @advancedxy We are not yet able to correctly to read structs. This showed up in ARROW-601, so this raises an exception for now

### DIFF
--- a/src/parquet/arrow/arrow-reader-writer-test.cc
+++ b/src/parquet/arrow/arrow-reader-writer-test.cc
@@ -948,26 +948,14 @@ class TestNestedSchemaRead : public ::testing::Test {
 
 TEST_F(TestNestedSchemaRead, ReadIntoTableFull) {
   std::shared_ptr<Table> table;
-  ASSERT_OK_NO_THROW(reader_->ReadTable(&table));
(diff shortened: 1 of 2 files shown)


advancedxy:
I am on vacation, sorry for the late response. I am ok with the change for now. But when do we want to support struct or nested complex type? I haven't had time to look at that issue, but hopefully It's done it get done in the near future.

wesm:
I think @itaiin is working on it. We would like to have complete nested types support for Arrow as soon as possible, but it's a fairly complex project because you have to deal with merging decoded data. So if we had

```
records: list<
  struct<a: int32, b: double>
>
```

Then in Parquet, we have two "columns":

```
records.a
records.b
```

And the record sizes and nullability is duplicated amongst the repetition/definition levels. In Arrow there is only one null bitmap for the struct/list nulls and the list offsets. So we have to deal with merging together the decoded intermediate structure.

itaiin:
@advancedxy Indeed I'm working on the issue. hope to open a PR for it soon.

advancedxy:
Glad to hear that.  Cann't wait to see your pr.

PR #309, opened 2017-04-22
PARQUET-595: Add API for key-value metadata

cpcloud:


cpcloud:
Definitely to format this.

cpcloud:
Fixing linting now.

src/parquet/file/metadata.h
```diff
@@ -178,6 +178,8 @@ class PARQUET_EXPORT FileMetaData {
   // Return const-pointer to make it clear that this object is not to be copied
   const SchemaDescriptor* schema() const;
 
+  std::unordered_map<std::string, std::string> key_value_metadata() const;
```
wesm: I'm a bit torn about exposing `unordered_map<string, string>` in the public API. The alternative would be something like:

```c++
class KeyValueMetadata {
 public:
  int length() const;

  const std::string& key(int i) const;
  const std::string& value(int i) const;

  void Append(const std::string& key, const std::string& value) const;

 private:
  std::vector<std::string> keys_;
  std::vector<std::string> values_;
};
```

then

```c++
const KeyValueMetadata& key_value_metadata() const;
```

cpcloud: One reason to keep `unordered_map` is so that we don't have to redefine another version of this class in `arrow-cpp`.

wesm: Since Arrow is a build dependency, we could have a common implementation in arrow/util. If that isn't too fancy

cpcloud: Ah ok. Makes sense then. I'll implement that.

cpcloud:
This will fail until I get the arrow patch up, and the parquet tests there will fail until this is merged. How should we handle that?

cpcloud:
Here's the arrow patch: https://github.com/apache/arrow/pull/587

wesm:
We can merge the Arrow patch with a broken build (assuming things check out otherwise) and then rebase this here (updating the hash in ThirdpartyToolchain.cmake) to get a passing build. This is sort of the worst case scenario for the two projects

cpcloud:
As I mentioned in https://github.com/apache/arrow/pull/587, I'm going to split that patch into two: One that implements the metadata API in arrow, and one that uses the metadata to implement pandas index reconstruction. That way, we can merge the first arrow patch, then this one, then the second arrow patch.

wesm:
After ARROW-898 https://github.com/apache/arrow/pull/605 is in, this will need a little work. I am happy to help, let me know

PR #297, opened 2017-04-14
PARQUET-918: Keep ordering in column indices when converting Parquet Schema

advancedxy:
This is a follow up fix for [PARQUET-918](https://github.com/apache/parquet-cpp/pull/295), do I need to create another jira for this?

Looks like some .idea files are included by accident. It looks no harm. Do I need to revert them?@wesm 

cc @wesm @itaiin  for reviewing

PR #312, opened 2017-04-26
PARQUET-911: [C++] Support nested structs in parquet_arrow

itaiin:
Support for simple StructArray reads.
Not supported in conjunction with lists yet.

### DIFF
--- a/src/parquet/arrow/arrow-reader-writer-test.cc
+++ b/src/parquet/arrow/arrow-reader-writer-test.cc
@@ -1250,7 +1250,7 @@ class TestNestedSchemaRead : public ::testing::TestWithParam<Repetition::type> {
     }
     auto schema_node = GroupNode::Make("schema", Repetition::REQUIRED, parquet_fields);
 
-    int num_columns = num_trees * std::pow(num_children, tree_depth);
+    int num_columns = num_trees * static_cast<int>((std::pow(num_children, tree_depth)));
 
     std::vector<int16_t> def_levels(num_rows);
     std::vector<int16_t> rep_levels(num_rows);
--- a/src/parquet/arrow/arrow-reader-writer-test.cc
+++ b/src/parquet/arrow/arrow-reader-writer-test.cc
@@ -50,6 +50,7 @@ using arrow::PrimitiveArray;
 using arrow::Status;
 using arrow::Table;
 using arrow::TimeUnit;
+using arrow::ArrayVisitor;
 
 using ArrowId = ::arrow::Type;
 using ParquetType = parquet::Type;
@@ -1080,10 +1081,10 @@ TEST(TestArrowWrite, CheckChunkSize) {
       Invalid, WriteTable(*table, ::arrow::default_memory_pool(), sink, chunk_size));
 }
 
-class TestNestedSchemaRead : public ::testing::Test {
+class TestNestedSchemaRead : public ::testing::TestWithParam<Repetition::type> {
  protected:
   // make it *3 to make it easily divisible by 3
-  const int num_test_rows = SMALL_SIZE * 3;
+  const int NUM_SIMPLE_TEST_ROWS = SMALL_SIZE * 3;
   std::shared_ptr<::arrow::Int32Array> values_array_ = nullptr;
 
   void InitReader() {
@@ -1106,6 +1107,12 @@ class TestNestedSchemaRead : public ::testing::Test {
   void FinalizeParquetFile() {
     row_group_writer_->Close();
     writer_->Close();
+  }
+
+  void MakeValues(int num_rows) {
+    std::shared_ptr<Array> arr;
+    ASSERT_OK(NullableArray<::arrow::Int32Type>(num_rows, 0, kDefaultSeed, &arr));
+    values_array_ = std::dynamic_pointer_cast<::arrow::Int32Array>(arr);
   }
 
   void WriteColumnData(size_t num_rows, int16_t* def_levels,
@@ -1142,13 +1149,14 @@ class TestNestedSchemaRead : public ::testing::Test {
     }
   }
 
-  void CreateNestedParquet(Repetition::type struct_repetition) {
+  // A parquet with a simple nested schema
+  void CreateSimpleNestedParquet(Repetition::type struct_repetition) {
     std::vector<NodePtr> parquet_fields;
     // TODO(itaiin): We are using parquet low-level file api to create the nested parquet
     // this needs to change when a nested writes are implemented
 
     // create the schema:
-    // <required/optional/repeated> group group1 {
+    // <struct_repetition> group group1 {
     //   required int32 leaf1;
     //   optional int32 leaf2;
     // }
@@ -1164,41 +1172,146 @@ class TestNestedSchemaRead : public ::testing::Test {
 
     // Create definition levels for the different columns that contain interleaved
     // nulls and values at all nesting levels
-    int16_t default_def_levels[num_test_rows];
-    memset(default_def_levels, 0, sizeof(default_def_levels));
-    int16_t default_rep_levels[num_test_rows];
-    memset(default_rep_levels, 0, sizeof(default_rep_levels));
+
     //  definition levels for optional fields
-    int16_t leaf1_def_levels[num_test_rows];
-    int16_t leaf2_def_levels[num_test_rows];
-    for (int i = 0; i < num_test_rows; i++)  {
+    int16_t leaf1_def_levels[NUM_SIMPLE_TEST_ROWS];
+    int16_t leaf2_def_levels[NUM_SIMPLE_TEST_ROWS];
+    int16_t leaf3_def_levels[NUM_SIMPLE_TEST_ROWS];
+    for (int i = 0; i < NUM_SIMPLE_TEST_ROWS; i++)  {
       // leaf1 is required within the optional group1, so it is only null
       // when the group is null
       leaf1_def_levels[i] = (i % 3 == 0) ? 0 : 1;
       // leaf2 is optional, can be null in the primitive (def-level 1) or
       // struct level (def-level 0)
       leaf2_def_levels[i] = i % 3;
+      // leaf3 is required
+      leaf3_def_levels[i] = 0;
     }
 
+    int16_t rep_levels[NUM_SIMPLE_TEST_ROWS];
+    memset(rep_levels, 0, sizeof(rep_levels));
+
     // Produce values for the columns
-    std::shared_ptr<Array> arr;
-    ASSERT_OK(NullableArray<::arrow::Int32Type>(num_test_rows, 0, kDefaultSeed, &arr));
-    values_array_ = std::dynamic_pointer_cast<::arrow::Int32Array>(arr);
+    MakeValues(NUM_SIMPLE_TEST_ROWS);
     int32_t* values = reinterpret_cast<int32_t*>(values_array_->data()->mutable_data());
 
     // Create the actual parquet file
-    InitNewParquetFile(std::static_pointer_cast<GroupNode>(schema_node), num_test_rows);
+    InitNewParquetFile(std::static_pointer_cast<GroupNode>(schema_node),
+      NUM_SIMPLE_TEST_ROWS);
 
     // leaf1 column
-    WriteColumnData(num_test_rows, leaf1_def_levels, default_rep_levels, values);
+    WriteColumnData(NUM_SIMPLE_TEST_ROWS, leaf1_def_levels, rep_levels, values);
     // leaf2 column
-    WriteColumnData(num_test_rows, leaf2_def_levels, default_rep_levels, values);
+    WriteColumnData(NUM_SIMPLE_TEST_ROWS, leaf2_def_levels, rep_levels, values);
     // leaf3 column
-    WriteColumnData(num_test_rows, default_def_levels, default_rep_levels, values);
+    WriteColumnData(NUM_SIMPLE_TEST_ROWS, leaf3_def_levels, rep_levels, values);
 
     FinalizeParquetFile();
     InitReader();
   }
+
+  NodePtr CreateDeepGroup(int depth, int num_children, Repetition::type node_repetition) {
+    std::vector<NodePtr> children;
+
+    for (int i = 0; i < num_children; i++) {
+      if (depth <= 1) {
+        children.push_back(PrimitiveNode::Make("leaf",
+          node_repetition, ParquetType::INT32));
+      } else {
+        children.push_back(CreateDeepGroup(depth - 1, num_children, node_repetition));
+      }
+    }
+
+    return NodePtr(GroupNode::Make("group", node_repetition, children));
+  }
+
+  // A deeply nested schema
+  void CreateMultiLevelNestedParquet(int num_trees, int tree_depth,
+      int num_children, int num_rows, Repetition::type node_repetition) {
+    // Create the schema
+    std::vector<NodePtr> parquet_fields;
+    for (int i = 0; i < num_trees; i++) {
+      parquet_fields.push_back(CreateDeepGroup(tree_depth, num_children,
+        node_repetition));
+    }
+    auto schema_node = GroupNode::Make("schema", Repetition::OPTIONAL, parquet_fields);
+
+    int num_columns = num_trees * std::pow(num_children, tree_depth);
+
+    int16_t def_levels[num_rows];
+    int16_t rep_levels[num_rows];
+    for (int i = 0; i < num_rows; i++) {
+      if (node_repetition == Repetition::REQUIRED) {
+        def_levels[i] = 0; // all is required
+      } else {
+        def_levels[i] = i % tree_depth; // all is optional
+      }
+      rep_levels[i] = 0; // none is repeated
+    }
+
+    // Produce values for the columns
+    MakeValues(num_rows);
+    int32_t* values = reinterpret_cast<int32_t*>(values_array_->data()->mutable_data());
+
+    // Create the actual parquet file
+    InitNewParquetFile(std::static_pointer_cast<GroupNode>(schema_node), num_rows);
+
+    for (int i = 0; i < num_columns; i++) {
+      WriteColumnData(num_rows, def_levels, rep_levels, values);
+    }
+    FinalizeParquetFile();
+    InitReader();
+  }
+
+  class DeepParquetTestVisitor : public ArrayVisitor {
+   public:
+    DeepParquetTestVisitor(Repetition::type node_repetition,
+      std::shared_ptr<::arrow::Int32Array> expected) :
+      node_repetition_(node_repetition), expected_(expected) {}
+
+    Status Validate(std::shared_ptr<Array> tree) {
+      return tree->Accept(this);
+    }
+
+    virtual Status Visit(const ::arrow::Int32Array& array) {
+      if (node_repetition_ == Repetition::REQUIRED) {
+        if (!array.Equals(expected_)) {
+          return Status::Invalid("leaf array data mismatch");
+        }
+      } else if (node_repetition_ == Repetition::OPTIONAL) {
+        if (array.length() != expected_->length()) {
+          return Status::Invalid("Bad leaf array length");
+        }
+        // expect only 1 value every `depth` row
+        if (array.null_count() != SMALL_SIZE) {
+          return Status::Invalid("Unexpected null count");
+        }
+      } else {
+        return Status::NotImplemented("Unsupported repetition");
+      }
+      return Status::OK();
+    }
+
+    virtual Status Visit(const ::arrow::StructArray& array) {
+      for (auto& child : array.fields()) {
+        if (node_repetition_ == Repetition::REQUIRED) {
+          RETURN_NOT_OK(child->Accept(this));
+        } else if (node_repetition_ == Repetition::OPTIONAL) {
+          // Null count Must be a multiple of SMALL_SIZE
+          if (array.null_count() % SMALL_SIZE != 0) {
+            return Status::Invalid("Unexpected struct null count");
+          }
+        } else {
+          return Status::NotImplemented("Unsupported repetition");
+        }
+      }
+      return Status::OK();
+    }
+
+   private:
+    Repetition::type node_repetition_;
+    std::shared_ptr<::arrow::Int32Array> expected_;
+  };
 
   std::shared_ptr<InMemoryOutputStream> nested_parquet_;
   std::shared_ptr<FileReader> reader_;
@@ -1207,11 +1320,11 @@ class TestNestedSchemaRead : public ::testing::Test {
 };
 
 TEST_F(TestNestedSchemaRead, ReadIntoTableFull) {
-  CreateNestedParquet(Repetition::OPTIONAL);
+  CreateSimpleNestedParquet(Repetition::OPTIONAL);
 
   std::shared_ptr<Table> table;
   ASSERT_OK_NO_THROW(reader_->ReadTable(&table));
-  ASSERT_EQ(table->num_rows(), num_test_rows);
+  ASSERT_EQ(table->num_rows(), NUM_SIMPLE_TEST_ROWS);
   ASSERT_EQ(table->num_columns(), 2);
   ASSERT_EQ(table->schema()->field(0)->type()->num_children(), 2);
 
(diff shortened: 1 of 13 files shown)


itaiin:
cc @wesm @advancedxy

advancedxy:
@itaiin is this ready for review? And do you want to support repeated struct support(list of struct  and struct of list) in this pr? 

If ready, I will allocate some time to do a proper review as this is a quite big change.

wesm:
I think it's ready for review. It would be nice to have complete nested types support, but I'm ok with proceeding in stages as long as the test cases don't have a lot of hard-coded assumptions about the completeness of the code

itaiin:
It is almost ready to review - after posting the PR I've played with it with some more complex parquet files I had and unfortunately found a bug. I'll fix it and extend the tests before your review.

About the hard-coded test cases - I dont think they are more specialized than before. I can think of  column-reader-test where there is one, but a) there was already an assumption there, and b) I think its non-essential and can be extended/eliminated, I just didn't want it to block the PR.

itaiin:
Fixed and added more testing. Should be ok to review.

advancedxy:
It's late in my time zone. I will do the review tomorrow morning

src/parquet/arrow/reader.cc
```diff
@@ -179,7 +183,13 @@ class FileReader::Impl {
   virtual ~Impl() {}
 
   Status GetColumn(int i, std::unique_ptr<ColumnReader>* out);
+  Status ReadSchemaField(int i, std::shared_ptr<Array>* out);
+  Status ReadSchemaField(int i, const std::vector<int>& indices,
```
advancedxy: Since we are here, I think we can make it clear in the document that(if we all agree):
`Column`: refers to leaf node in the parquet/arrow schema definition.
`Field`: refers to the base node(a.k.a top level) in the parquet schema definition, mapping to arrow's Field.

And change all the occurrence of `indices` to the name of `included_column_indices` to avoid any confusion. 

Also we can use include_column_indices to represent fields selection, and I think we can build an unified api to use include_column_indices only.

src/parquet/arrow/reader.cc
```diff
@@ -1098,5 +1299,79 @@ Status ColumnReader::NextBatch(int batch_size, std::shared_ptr<Array>* out) {
...
+      ::arrow::BitUtil::ClearBit(null_bitmap_ptr, i);
+    }
+  }
+
+  *null_count_out = null_count;
+  *null_bitmap_out = (null_count == 0) ? nullptr : null_bitmap;
+  return Status::OK();
+}
+
+Status StructImpl::DefLevels(ValueLevelsPtr* out) {
```
advancedxy: Is caching needed here? For now, it looks like we only calculated this once. But it may be used multiple times.  I think we can add a comment about caching.

src/parquet/arrow/arrow-reader-writer-test.cc
```diff
@@ -1074,61 +1073,296 @@ class TestNestedSchemaRead : public ::testing::Test {
...
+      if (array->IsNull(i)) {
+        continue;
+      }
+      ASSERT_EQ(array->Value(i), values_array_->Value(j));
+      j++;
+    }
+  }
+
+  // A parquet with a simple nested schema
+  void CreateSimpleNestedParquet(Repetition::type struct_repetition) {
```
advancedxy: As the impl indicates, we can handle struct of struct. Maybe we should also add a test case for that.

advancedxy:
ping @itaiin any updates?

itaiin:
@advancedxy thanks a lot for the review! sorry for not being responsive, its just that i am currently away and wont  be back in a few days. i will get right on it as soon as i can.

src/parquet/arrow/arrow-reader-writer-test.cc
```diff
@@ -1074,35 +1072,96 @@ class TestNestedSchemaRead : public ::testing::Test {
...
+    parquet_fields.push_back(GroupNode::Make("group1", struct_repetition,
         {PrimitiveNode::Make("leaf1", Repetition::REQUIRED, ParquetType::INT32),
-            PrimitiveNode::Make("leaf2", Repetition::REQUIRED, ParquetType::INT32)}));
+         PrimitiveNode::Make("leaf2", Repetition::OPTIONAL, ParquetType::INT32)}));
     parquet_fields.push_back(
         PrimitiveNode::Make("leaf3", Repetition::REQUIRED, ParquetType::INT32));
 
-    const int num_columns = 3;
-    auto schema_node = GroupNode::Make("schema", Repetition::REQUIRED, parquet_fields);
+    auto schema_node = GroupNode::Make("schema", Repetition::OPTIONAL, parquet_fields);
```
itaiin: Good catch.

src/parquet/arrow/reader.cc
itaiin: Yes I pondered about this a bit..
But I figured its probably best to keep it simple for now, and maybe add caching after a struct writer is introduced.
I can add a comment about that.

src/parquet/arrow/arrow-reader-writer-test.cc
itaiin: Right. Basically that is why I added the single type deep schema parquet tests (which uses `CreateMultiLevelNestedParquet`).
I thought the hand-crafted schema test (which uses this `CreateSimpleNestedParquet` function) will become too elaborated if I test that too. 
Do you think that there's an added edge case there to make it worthwhile?

src/parquet/arrow/reader.cc
itaiin: I'm 100% with you on the definitions.
Also the naming make sense to me however I think the coherency and the docs are the most crucial part so I don't have as strong feelings about it :)

src/parquet/arrow/arrow-reader-writer-test.cc
```diff
@@ -1074,61 +1073,296 @@ class TestNestedSchemaRead : public ::testing::Test {
...
+    // leaf2 column
+    WriteColumnData(NUM_SIMPLE_TEST_ROWS, leaf2_def_levels, rep_levels, values);
+    // leaf3 column
+    WriteColumnData(NUM_SIMPLE_TEST_ROWS, leaf3_def_levels, rep_levels, values);
+
+    FinalizeParquetFile();
+    InitReader();
+  }
+
+  NodePtr CreateDeepGroup(int depth, int num_children, Repetition::type node_repetition) {
```
itaiin: Right. I'll add a `leaf_type` parameter to it.

src/parquet/arrow/arrow-reader-writer-test.cc
advancedxy: DeepNestedSchemaRead  test should do it. I must have overlooked them when reviewing.

src/parquet/arrow/arrow-reader-writer-test.cc
```diff
@@ -1074,61 +1073,296 @@ class TestNestedSchemaRead : public ::testing::Test {
...
+    for (int i = 0; i < num_children; i++) {
+      if (depth <= 1) {
+        children.push_back(PrimitiveNode::Make("leaf",
+          node_repetition, ParquetType::INT32));
+      } else {
+        children.push_back(CreateDeepGroup(depth - 1, num_children, node_repetition));
+      }
     }
 
+    return NodePtr(GroupNode::Make("group", node_repetition, children));
```
itaiin: also adding depth

src/parquet/schema.cc
```diff
@@ -620,6 +620,15 @@ const ColumnDescriptor* SchemaDescriptor::Column(int i) const {
   return &leaves_[i];
 }
 
+int SchemaDescriptor::ColumnIndex(const NodePtr& node) const {
```
itaiin: It seemed to me like a good idea at first, but unfortunately this creates a bit of a problem of using `shared_ptr` (which `NodePtr` are) as keys to an `unordered_map`. Doing it with std objects seems complicated (seems like boost solved it but don't think we want to convert all std::shared_ptr to boost:shared_ptr).

I can use Node* instead but that might be bug prone.

src/parquet/util/schema-util.h
```diff
@@ -0,0 +1,58 @@
...
+//   If the name is array or ends in _tuple, this should be a list of struct
+//   even for single child elements.
+inline bool HasStructListName(const GroupNode* node) {
+  return (node->name() == "array" ||
+          str_endswith_tuple(node->name()));
+}
+
+// Find the index of the field node under the schema root of a leaf node
+// return a negative result if not found
+inline int ColumnIndexToSchemaFieldIndex(const SchemaDescriptor* schema, int i) {
```
itaiin: Well actually the intention was that ColumnIndex is needed mostly so I can reuse the existing Column reader API.
But now that you mention it, maybe this is unnecessary and I should extend it to get it from the NodePtr object directly. I'll check this option out.

src/parquet/schema.cc
advancedxy: @wesm do you have a solution for this situation? shared_ptr as keys to an `unordered_map` indeed seems complicated. But I also think O(n) must be avoided.

advancedxy: @itaiin It occurred to me this morning, we shouldn't use shared_ptr as keys as they are mutable. And we don't need to: we can use ColumnPath of NodePtr as keys. And it's fair easy to reconstruct ColumnPath from NodePtr

wesm: Sorry, for my low activity on this patch, let me take a closer look tomorrow morning

src/parquet/util/schema-util.h
itaiin: After looking a bit into it, it seems to me that using indices to read columns decouples the schema object hierarchy from the reading mechanism which is probably a good thing in terms of simplicity and separation of concerns. 
So ColumnIndex  is necessary to keep it this way.

src/parquet/schema.cc
itaiin: @advancedxy well, the shared_ptr itself is mutable because of the refcounting, but the `Node` object itself can be immutable just the same as a `ColumnPath` object. And imho it seems that what matters anyways in terms of unordered_map is that they are hashable and comparable which should be defined on the object and not the `shared_ptr`.
Another option is to use `std::string` for the key in conjunction with `ColumnPath::ToDotString`. But I'm guessing this will practically be slower than the O(n) array pass because of all the string manipulation.
Another possibility is to define the compare+hash over `std::vector<std::string>` and use `ColumnPath::ToDotVector`, but I don't see any major benefit over doing it with `ColumnPath`.

advancedxy: >And imho it seems that what matters anyways in terms of unordered_map is that they are hashable and comparable which should be defined on the object and not the shared_ptr.

I understand that, but it requires we supply our Hash and Predict class which adds complexity.
>Another option is to use std::string for the key in conjunction with ColumnPath::ToDotString. But I'm guessing this will practically be slower than the O(n) array pass because of all the string manipulation.

This is what I want to proposal.  Didn't make it clear in the previous comment.
As for performance, It might be slower for small size of leaf columns in the schema.  But it's bounded to O(1) look up if we be careful and cache ColumnPath::ToDotString.

It's also easy for user to use this api, they can just supply a string and query.

itaiin: I agree that adding the classes will add unnecessary code complexity. 
I will make the changes for the ToDotString method. I'm just think that about caching, since `NodePtr`s can be moved around (when manipulating schemas for example), it can be a bit tricky and the PR is kind of big as it is. 
So I offer that for the time being I won't add caching but leave a TODO comment for it, since it seems to me that in any case this is not called a lot and not worth the effort, at least currently.

src/parquet/column/levels.h
```diff
@@ -82,5 +83,32 @@ class PARQUET_EXPORT LevelDecoder {
   std::unique_ptr<BitReader> bit_packed_decoder_;
 };
 
+// ----------------------------------------------------------------------
+// Definition and Repetition level vectors
```
xhochy: This seems like dead code?

itaiin: Actually it is being used (except `StructRepLevels` which is there for future repeated+struct impl, maybe I should remove it).
I have avoided converting all of the existing code that uses `uint16[]` to use these datatypes though because it has occured to me that these things may change with the full complex schema impl. That's is also why I've only put there some basic encapsulations of the logic in the form of static factory functions.

itaiin:
@xhochy I've pushed a commit with all of the fixes according to the conversation so far. 
If @advancedxy thinks they're cool then I think its ready.

src/parquet/arrow/reader.h
```diff
@@ -99,10 +99,18 @@ class PARQUET_EXPORT FileReader {
   // Read column as a whole into an Array.
   ::arrow::Status ReadColumn(int i, std::shared_ptr<::arrow::Array>* out);
 
-  // Read a table of flat columns into a Table.
+  // Reads a specific top level schema field into an Array
+  ::arrow::Status ReadSchemaField(int i, std::shared_ptr<::arrow::Array>* out);
+
+  // Reads a specific top level schema field into an Array
+  // Read only the indicated child columns
```
advancedxy: nit: child -> leaf may be more precise.

src/parquet/util/schema-util.h
```diff
@@ -0,0 +1,102 @@
...
+// Coalesce a list of schema fields indices which are the roots of the
+// columns referred by a list of column indices
+inline bool ColumnIndicesToFieldIndices(const SchemaDescriptor& descr,
+    const std::vector<int>& column_indices, std::vector<int>* out) {
+  auto group = descr.group_node();
+  auto num_fields = group->field_count();
+  std::vector<bool> fields_to_read(descr.group_node()->field_count(), false);
+  // Copy the indices vector and then sort it.
+  // This ensures the corresponding schema field indices of the columns are
+  // monotonic too which enables a single-pass.
```
advancedxy: I believe we made some effort to keep the ordering of included_column_indices.. see my related [pr](https://github.com/apache/parquet-cpp/pull/297).  I this this behaviour should be kept too.

src/parquet/arrow/arrow-reader-writer-test.cc
```diff
@@ -1104,61 +1103,300 @@ class TestNestedSchemaRead : public ::testing::Test {
...
+    WriteColumnData(NUM_SIMPLE_TEST_ROWS, leaf2_def_levels, rep_levels, values);
+    // leaf3 column
+    WriteColumnData(NUM_SIMPLE_TEST_ROWS, leaf3_def_levels, rep_levels, values);
+
+    FinalizeParquetFile();
+    InitReader();
+  }
+
+  NodePtr CreateSingleTypedNestedGroup(int index, int depth, int num_children,
+      Repetition::type node_repetition, ParquetType::type leaf_type) {
```
advancedxy: I was thinking something like `template <typename primitive_type>` with `std::enable_if`
But I can accept this in test code. Leave comments to other reviewers..

advancedxy:
@itaiin some minor comments plus the included_column_indices ordering..

src/parquet/util/schema-util.h
itaiin: Right. Will fix. Seems like a test case for this is also needed.

src/parquet/schema.h
```diff
@@ -261,16 +267,23 @@ class PARQUET_EXPORT GroupNode : public Node {
...
       field->SetParent(this);
+      field_name_to_idx_[field->name()] = field_idx++;
     }
   }
 
   NodeVector fields_;
   bool EqualsInternal(const GroupNode* other) const;
 
+  // Mapping between field name to the field index
+  std::unordered_map<std::string, int> field_name_to_idx_;
```
advancedxy: Being finicky here... Wondering if it's too much to add an unordered_map for every GroupNode in the schema, is SchemaDescriptor a better place to go?

No objections here.

src/parquet/arrow/arrow-reader-writer-test.cc
```diff
@@ -1074,22 +1075,20 @@ TEST(TestArrowWrite, CheckChunkSize) {
...
-    InitReader(&reader_);
-  }
+  // make it *3 to make it easily divisible by 3
+  const int NUM_SIMPLE_TEST_ROWS = SMALL_SIZE * 3;
+  std::shared_ptr<::arrow::Int32Array> values_array_ = nullptr;
 
-  void InitReader(std::shared_ptr<FileReader>* out) {
+  void InitReader() {
     std::shared_ptr<Buffer> buffer = nested_parquet_->GetBuffer();
     std::unique_ptr<FileReader> reader;
```
xhochy: With your changes, we should directly use `reader_` here instead of adding another local variable.

src/parquet/arrow/reader.cc
```diff
@@ -243,6 +263,11 @@ class ColumnReader::Impl {
   Status WrapIntoListArray(const int16_t* def_levels, const int16_t* rep_levels,
       int64_t total_values_read, std::shared_ptr<Array>* array);
 
+  virtual Status DefLevels(ValueLevelsPtr* out);
```
xhochy: Use `override` instead of `virtual` for implementations

src/parquet/arrow/reader.cc
```diff
@@ -272,6 +297,33 @@ class ColumnReader::Impl {
...
+ public:
+  explicit StructImpl(const std::vector<std::shared_ptr<Impl>>& children,
+      int16_t struct_def_level, MemoryPool* pool, const NodePtr& node)
+      : children_(children), struct_def_level_(struct_def_level), pool_(pool) {
+    NodeToField(node, &field_);
+  }
+
+  virtual ~StructImpl() {}
+
+  virtual Status NextBatch(int batch_size, std::shared_ptr<Array>* out);
```
xhochy: Use `override` instead of `virtual` for implementations

src/parquet/schema.h
itaiin: I think its ok. I dont think its a big overhead.
I have been tempted to use the same map on `SchemaDescriptor` used for mapping paths to leaves for all nodes and use it instead but this way seemed easier to navigate the tree, even where you don't have the descriptor object.

src/parquet/util/schema-util.h
```diff
@@ -0,0 +1,84 @@
+// Licensed to the Apache Software Foundation (ASF) under one
```
itaiin: Thought of sticking them there (or schema.cc), but it seems to me that schema.h is long as it plus it should be more about parquet schema objects declaration, and whats in schema-util is just aux aux. functions which are more implementation details.

src/parquet/arrow/reader.cc
itaiin: I'm guessing virtual for the destructor is ok though?

src/parquet/column/levels.h
```diff
@@ -82,5 +83,32 @@ class PARQUET_EXPORT LevelDecoder {
...
+
+typedef std::vector<int16_t> ValueLevels;
+typedef std::shared_ptr<std::vector<int16_t>> ValueLevelsPtr;
```
itaiin: My thinking was that the lower level API could benefit not messing around with all the int16_t* (even though I haven't implemented it in this PR).
Didn't take into account the pool allocators though. I'll use a PoolBuffer in `arrow/reader.cc` instead.

wesm:
Sorry I haven't gotten down to review this yet. I think we are going to merge this right after parquet-cpp 1.1.0 and then we can make sure this gets into 1.2.0 and Arrow 0.5.0 (which should be released within a month of 1.1.0 / 0.4.0, respectively)

itaiin:
@wesm no worries. Also I think the reviews so far were very helpful.

src/parquet/arrow/reader.cc
xhochy: Sure, that's a different thing than overriding methods 😉

advancedxy:
Any updates on this pr?

wesm:
I'm reviewing finally, can you rebase? thanks

src/parquet/util/schema-util.h
```diff
@@ -0,0 +1,84 @@
...
+    //   even for single child elements.
+  GroupNode* group = static_cast<GroupNode*>(node.get());
+  if (group->field_count() == 1 && HasStructListName(group)) return false;
+
+  return true;
+}
+
+// Coalesce a list of schema fields indices which are the roots of the
+// columns referred by a list of column indices
+inline bool ColumnIndicesToFieldIndices(const SchemaDescriptor& descr,
```
wesm: I'm a bit confused by the terminology in this PR

In a Parquet file, the list of thrift `parquet::SchemaElement` values define a semantic schema with columns of the form:

```
0 foo.bar
1 foo.bar.baz
2 foo.qux
3 foo2
4 foo3
5 foo4.qux
```

Our intent in the Arrow API was to only reference these semantic column numbers, but if you were to read `{0, 1, 2}` then the reader must deal with reading the tree for the 3 leaf nodes. How does the terminology here (column indices vs field indices) correspond to this? Maybe a text illustration in the comments would help

itaiin:
@wesm thanks for reviewing. 
I've found yesterday a small bug with partial struct reads, I'll commit a change + test along with the rebase.

src/parquet/arrow/arrow-reader-writer-test.cc
```diff
@@ -1104,61 +1100,314 @@ class TestNestedSchemaRead : public ::testing::Test {
...
+         PrimitiveNode::Make("leaf2", Repetition::OPTIONAL, ParquetType::INT32)}));
     parquet_fields.push_back(
         PrimitiveNode::Make("leaf3", Repetition::REQUIRED, ParquetType::INT32));
 
-    const int num_columns = 3;
     auto schema_node = GroupNode::Make("schema", Repetition::REQUIRED, parquet_fields);
 
-    InitNewParquetFile(std::static_pointer_cast<GroupNode>(schema_node), 0);
+    // Create definition levels for the different columns that contain interleaved
+    // nulls and values at all nesting levels
```
itaiin: Using JSON and fuzzing sounds great -  would make testing less of a pain and would definitely find way more bugs (e.g. the one I found yesterday..). 
I do believe that working on writing should happen right after solving full nested data reads.

src/parquet/arrow/reader.cc
```diff
@@ -204,9 +214,21 @@ class FileReader::Impl {
   int num_threads_;
 };
 
+typedef const int16_t* ValueLevelsPtr;
```
itaiin: With that naming I was trying to mimic other typedef and hide implementation details, but when you mention it - maybe visibility is better here? I don't have a solid opinion here.. I'll make the change if you think it makes sense.

src/parquet/arrow/arrow-reader-writer-test.cc
```diff
@@ -1104,61 +1100,314 @@ class TestNestedSchemaRead : public ::testing::Test {
...
+  }
+
+  void WriteColumnData(size_t num_rows, int16_t* def_levels,
+      int16_t* rep_levels, int32_t* values) {
+    auto typed_writer = static_cast<TypedColumnWriter<Int32Type>*>(
+      row_group_writer_->NextColumn());
+    typed_writer->WriteBatch(num_rows, def_levels, rep_levels, values);
+  }
+
+  void ValidateArray(std::shared_ptr<Array> array, size_t expected_nulls) {
```
itaiin: Good tip. Since i'm changing the tests a bit, I'll change it too.

src/parquet/arrow/reader.cc
```diff
@@ -272,6 +299,35 @@ class ColumnReader::Impl {
...
+ private:
+  std::vector<std::shared_ptr<Impl>> children_;
+  int16_t struct_def_level_;
+  MemoryPool* pool_;
+  std::shared_ptr<Field> field_;
+  PoolBuffer def_levels_buffer_;
+
+  Status CreateBitmap(size_t num_elements, std::shared_ptr<PoolBuffer>* null_bitmap_out);
+  Status DefLevelsToNullArray(std::shared_ptr<PoolBuffer>* null_bitmap,
+    int64_t* null_count);
```
itaiin: Good catch

src/parquet/arrow/reader.cc
```diff
@@ -281,11 +337,95 @@ Status FileReader::Impl::GetColumn(int i, std::unique_ptr<ColumnReader>* out) {
...
+      walker = group->field(0);
+    }
+    auto column_index = reader_->metadata()->schema()->ColumnIndex(*walker.get());
+
+    if (std::find(indices.begin(), indices.end(), column_index) != indices.end()) {
+      // The index of the column is found, therefore a reader is needed
+      std::unique_ptr<ColumnReader> reader;
+      RETURN_NOT_OK(GetColumn(column_index, &reader));
+      *out = std::move(reader->impl_);
+    }
```
itaiin: the column reader is not needed, which is a normal behavior so no error, so the `out` parameter keeps being nullptr. (and nullity is being checked in the callers levels)
I'll add a comment for clarity.

src/parquet/arrow/reader.cc
```diff
@@ -1159,5 +1328,102 @@ Status ColumnReader::NextBatch(int batch_size, std::shared_ptr<Array>* out) {
...
+  RETURN_NOT_OK(children_[0]->GetDefLevels(&child_def_levels, &child_length));
+  auto size = child_length * sizeof(int16_t);
+  def_levels_buffer_.Resize(size);
+  // Initialize with the minimal def level
+  std::memset(def_levels_buffer_.mutable_data(), 0, size);
+  auto result_levels = reinterpret_cast<int16_t*>(def_levels_buffer_.mutable_data());
+
+  // A struct is defined if at least one of its children is defined,
+  // so the definition level of a struct value is the max definition level
+  // of its children
```
itaiin: Its true. The `max` logic is still correct (given the def data isn't botched somehow) but I have to make the first line of the comment more accurate since it neglects that scenario.

itaiin: Now it got me wondering: even though `max` works (and is probably the fastest), maybe I should change it too to detect contradictory struct def levels?

src/parquet/arrow/reader.h
```diff
@@ -99,10 +99,18 @@ class PARQUET_EXPORT FileReader {
   // Read column as a whole into an Array.
   ::arrow::Status ReadColumn(int i, std::shared_ptr<::arrow::Array>* out);
 
-  // Read a table of flat columns into a Table.
+  // Reads a specific top level schema field into an Array
+  ::arrow::Status ReadSchemaField(int i, std::shared_ptr<::arrow::Array>* out);
+
+  // Reads a specific top level schema field into an Array
+  // Read only the indicated leaf columns
```
itaiin: Yes, I wasn't sure about this too - I think that there should be a decision about terminology which also derives the APIs.
cc @advancedxy

src/parquet/arrow/reader.cc
```diff
@@ -1159,5 +1328,102 @@ Status ColumnReader::NextBatch(int batch_size, std::shared_ptr<Array>* out) {
...
 
+// StructImpl methods
+
+Status StructImpl::CreateBitmap(size_t num_elements,
+    std::shared_ptr<PoolBuffer>* null_bitmap_out) {
+  *null_bitmap_out = std::make_shared<::arrow::PoolBuffer>(pool_);
+  auto bitmap_size = ::arrow::BitUtil::CeilByte(num_elements) / 8;
+  RETURN_NOT_OK((*null_bitmap_out)->Resize(bitmap_size));
+  uint8_t* null_bitmap_ptr = (*null_bitmap_out)->mutable_data();
+  memset(null_bitmap_ptr, 0xFF, bitmap_size);
```
itaiin: The fact that its zero'd is not a real problem, however there's the issue of getting a `MutableBuffer` instead of a `PoolBuffer`. Is there any downside of using the first kind in this context instead of the other?

src/parquet/util/schema-util.h
advancedxy: Thanks for the clarifying. 

> Our intent in the Arrow API was to only reference these semantic column numbers

This intent is unclear to me as SchemaDescriptor only uses leaves to get Column, see code below:
```cpp
const ColumnDescriptor* SchemaDescriptor::Column(int i) const {
  DCHECK(i >= 0 && i < static_cast<int>(leaves_.size()));
  return &leaves_[i];
}
```
I used to think column reference to leaf node until I saw your clarification and some parquet-mr code digging.

Back to column vs field indices, I believe it's clear what field means in arrow::Schema, the top(base) node, which would be {foo, foo2, foo3, foo4} in your example, and I think it's much easier that if there is a ReadField api so that we can simply read corresponding field by index.

Also for API usage, it would be quite common to selected nodes(field or column) of interested by user. For example, user only wants to read the first and third field in the file(foo and foo3 in your example), or by name path: {foo, foo2, foo4.qux}, {foo.bar, foo4}.  To be able to support that,  I think we should also do column index mapping.
> if you were to read {0, 1, 2} then the reader must deal with reading the tree for the 3 leaf nodes.

I think we should merge this read request to generate an arrow array constrained by `Schema<Field foo>`

itaiin: Not sure if I am missing anything or not - is there an actual difference between parquet columns and what the leaf nodes of its schema refer to? In my mind the column indices are refering to either because they are mapped one-to-one.
As of the field indices, I do mean them in the exact same way @advancedxy does,
however while I think choosing any subset of columns to read is a very necessary feature, I am not 100% sure whether `ReadField` should be exposed as an API. In any case its an implementation detail that is an easy way to keep the parallel reads on many parquet files,

src/parquet/arrow/reader.cc
itaiin: Now using `GetEmptyBitmap`

src/parquet/arrow/reader.cc
itaiin: Ended up changing the doc and also the implementation a bit to be more robust.

src/parquet/util/schema-util.h
wesm: I think it's OK to leave the `ReadSchemaField` functions but we need a docstring that makes it abundantly clear what's going on. So in my example:

```
0 foo.bar
1 foo.bar.baz
2 foo.qux
3 foo2
4 foo3
5 foo4.qux
```

If we did `ReadSchemaField(0, ...)` then it's reading the 3 columns `{0, 1, 2}`. Perhaps this can also be marked experimental until we have complete nested type support

wesm:
@itaiin can you fix the Appveyor failure (it looks like just a compiler warning, may require a static_cast or two)? Then I think we can merge this and move on to the next iteration of nested data support.

itaiin:
@wesm cool. will fix right after the weekend

wesm:
@xhochy can you have a last look? I think we should probably merge this and then start a design document about implementing nested data "properly" (implementing a state machine capable of decoding any nested / repeated tree of types into Arrow memory), which adequate testing (perhaps automatically generated, using Arrow's JSON file format that we now use for integration testing) so that we're able to exhaust all of the possibilities -- null structs, empty / null lists, lists with null or empty structs, and so forth.

xhochy:
@wesm That sounds fine to me. Feel free to merge.

wesm:
@itaiin thank you!

itaiin:
@xhochy  @advancedxy & @wesm thanks for the reviews.
And thanks for merging the PR, I'm glad I could be helpful! 
Next phase I guess is full read support.

advancedxy:
@itaiin you are welcome. Our internal compute engine (will open source in this year if possible) also needs nested struct support from parquet-cpp.

PR #314, opened 2017-04-28
PARQUET-595: API for KeyValue metadata

wesm:
This supersedes #309 and incorporates the `std::shared_ptr<const KeyValueMetadata>` pattern so less copying is needed in Parquet for metadata inbound from Arrow (and vice versa).

close #309


src/parquet/arrow/schema.h
```diff
@@ -44,13 +45,17 @@ ::arrow::Status PARQUET_EXPORT NodeToField(
...
-    const std::vector<int>& column_indices, std::shared_ptr<::arrow::Schema>* out);
+    const std::vector<int>& column_indices,
+    const std::shared_ptr<const KeyValueMetadata>& key_value_metadata,
+    std::shared_ptr<::arrow::Schema>* out);
 
-::arrow::Status PARQUET_EXPORT FromParquetSchema(
-    const SchemaDescriptor* parquet_schema, std::shared_ptr<::arrow::Schema>* out);
+::arrow::Status PARQUET_EXPORT FromParquetSchema(const SchemaDescriptor* parquet_schema,
+    const std::shared_ptr<const KeyValueMetadata>& key_value_metadata,
+    std::shared_ptr<::arrow::Schema>* out);
```
advancedxy: Interface question: 
I think it's better to keep the previous exported interface.
```
::arrow::Status PARQUET_EXPORT FromParquetSchema(
      const SchemaDescriptor* parquet_schema, std::shared_ptr<::arrow::Schema>* out);
```

so it doesn't break user's code and if user doesn't care about key_value_metadata, he can just ignore them.

what do you think? @wesm

wesm: I agree. I will make a pass through and add backwards compatible APIs.

wesm: Done. Does this patch look OK?

src/parquet/arrow/schema.h
```diff
@@ -19,6 +19,7 @@
 #define PARQUET_ARROW_SCHEMA_H
 
 #include <memory>
+#include <string>
```
cpcloud: `std::string` isn't used in this file right?

wesm: Removed. We could use a include-what-you-use pass on the whole codebase

src/parquet/arrow/writer.cc
```diff
@@ -18,6 +18,7 @@
 #include "parquet/arrow/writer.h"
 
 #include <algorithm>
+#include <string>
```
cpcloud: Is `std::string` being used in this file?

src/parquet/file/writer.h
```diff
@@ -103,12 +124,14 @@ class PARQUET_EXPORT ParquetFileWriter {
...
-      const std::shared_ptr<WriterProperties>& properties = default_writer_properties());
+      const std::shared_ptr<WriterProperties>& properties = default_writer_properties(),
+      const std::shared_ptr<const KeyValueMetadata>& key_value_metadata = nullptr);
 
   static std::unique_ptr<ParquetFileWriter> Open(
       const std::shared_ptr<OutputStream>& sink,
       const std::shared_ptr<schema::GroupNode>& schema,
-      const std::shared_ptr<WriterProperties>& properties = default_writer_properties());
+      const std::shared_ptr<WriterProperties>& properties = default_writer_properties(),
+      const std::shared_ptr<const KeyValueMetadata>& key_value_metadata = nullptr);
```
cpcloud: If someone is using this API, and they're linking against the `libparquet` shared object, will changing this signature still allow existing code to continue working?

src/parquet/file/writer.h
```diff
@@ -20,8 +20,10 @@
 
 #include <cstdint>
 #include <memory>
+#include <string>
```
cpcloud: Is `std::string` being used here?

src/parquet/file/writer-internal.h
```diff
@@ -19,6 +19,7 @@
 #define PARQUET_FILE_WRITER_INTERNAL_H
 
 #include <memory>
+#include <string>
```
cpcloud: `string` used in this file?

src/parquet/file/writer.h
wesm: Yep, I think this actually will generate two symbols, one with the new argument and one without

src/parquet/file/writer-internal.h
wesm: removed and above, artifact of the old data structure

cpcloud:
Thanks for getting the `shared_ptr` stuff up! I'm cranking on ARROW-881 as we speak.

cpcloud:
(Along with decimal madness)

advancedxy:
I did another quick review. LGTM.

wesm:
Thanks. Merging

PR #316, opened 2017-05-02
PARQUET-936: Return Invalid Status if chunk_size <= 0 when WriteTable in parquet-arrow

advancedxy:


src/parquet/arrow/writer.cc
```diff
@@ -717,6 +717,10 @@ Status FileWriter::WriteTable(const Table& table, int64_t chunk_size) {
     }
   }
 
+  if (chunk_size <= 0) {
+    return Status::Invalid("chunk size per row_group must be greater than 0");
```
advancedxy: Invalid Status or DCHECK? I prefer Invalid Status

Also, is a test needed? I prefer not. 
@wesm

wesm: Can you test that error status is returned when chunk_size is equal to 0? Better than segfault or infinite loop

advancedxy:
ping @wesm, added a test. 

Some morning work, _^^_.

PR #317, opened 2017-05-04
PARQUET-914: Rewording exception message in column writer.

advancedxy:


src/parquet/column/writer.cc
```diff
@@ -187,9 +187,10 @@ int64_t ColumnWriter::Close() {
   }
 
   if (num_rows_ != expected_rows_) {
-    throw ParquetException(
-        "Less than the number of expected rows written in"
-        " the current column chunk");
+    std::stringstream ss;
```
advancedxy: Do we prefer to not use stringstream?

advancedxy:
@xhochy I thinks PARQUET-914 should go to cpp-1.1.0 too.

src/parquet/column/writer.cc
xhochy: `std::stringstream` is fine. Exceptions are not performance critical.

PR #319, opened 2017-05-07
PARQUET-679: [C++] Resolve unit tests issues on Windows; Run unit tes…

MaxRis:
…ts with Appveyor; Appveyor build jobs for Visual Studio cmake generator and Debug configuration.

### DIFF
--- a/cmake_modules/ThirdpartyToolchain.cmake
+++ b/cmake_modules/ThirdpartyToolchain.cmake
@@ -82,7 +82,7 @@ if (PARQUET_BOOST_USE_SHARED)
   endif()
 
   find_package(Boost COMPONENTS regex REQUIRED)
-  if ("${CMAKE_BUILD_TYPE}" STREQUAL "DEBUG")
+  if ("${UPPERCASE_BUILD_TYPE}" STREQUAL "DEBUG")
     set(BOOST_SHARED_REGEX_LIBRARY ${Boost_REGEX_LIBRARY_DEBUG})
   else()
     set(BOOST_SHARED_REGEX_LIBRARY ${Boost_REGEX_LIBRARY_RELEASE})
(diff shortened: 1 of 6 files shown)


cmake_modules/ThirdpartyToolchain.cmake
```diff
@@ -175,7 +179,11 @@ if (NOT THRIFT_FOUND)
   set(THRIFT_HOME "${THRIFT_PREFIX}")
   set(THRIFT_INCLUDE_DIR "${THRIFT_PREFIX}/include")
   IF (${UPPERCASE_BUILD_TYPE} STREQUAL "DEBUG")
-    set(THRIFT_STATIC_LIB "${THRIFT_PREFIX}/lib/libthriftd.a")
+    IF (MSVC)
+      set(THRIFT_STATIC_LIB "${THRIFT_PREFIX}/lib/thriftmdd.lib")
```
xhochy: No need here for `if(MSVC)`, we should be able to handle this better by using `CMAKE_STATIC_LIBRARY_PREFIX` and `CMAKE_SHARED_LIBRARY_SUFFIX`

src/parquet/arrow/arrow-reader-writer-test.cc
```diff
@@ -661,10 +661,10 @@ TEST_F(TestInt96ParquetIO, ReadIntoTimestamp) {
 
   // 2nd January 1970, 11:35min 145738543ns
   Int96 day;
-  day.value[2] = 2440589l;
+  day.value[2] = (uint32_t)2440589;
```
xhochy: We should use here C++'s `static_cast<uint32_t>`

cmake_modules/ThirdpartyToolchain.cmake
wesm: MSVC has the complexity with the MT vs MD variants of libraries (whether they link to the dynamic or static MSVC c/c++ libraries), so this may be necessary

MaxRis: as @wesm mentioned, in case of thrift the required lib is with MD suffix, but others 3rd party dependencies of Parquet might be named without MD suffix at all. So it seems additional checks should be placed somewhere in any case, why not to use this solution at least for now :)

xhochy:
@MaxRis Can you activate AppVeyor on your fork? Thus we should get faster builds. It seems like the ASF AppVeyor queue is stuck / quite delayed.

PR #321, opened 2017-05-09
PARQUET-930: Add timestamp[us] to schema test

xhochy:


xhochy:
@wesm This is the only remaining thing for PARQUET-930 I found. Did I miss something?

PR #323, opened 2017-05-13
PARQUET-977: retire pre-3.2 cmake to simplify and cleanup cmake files

rip-nsk:


wesm:
Realistically we need to support cmake 2.8.x for Linux (e.g Ubuntu 14.04 is on 2.8.x)

rip-nsk:
On my Ubuntu (16.04) "apt-get install cmake" installs 3.5...

wesm:
Right, but 14.04 LTS has a pretty large install base

wesm:
We could possibly vendor CMake, which is not that uncommon (e.g. see Apache Kudu https://github.com/apache/kudu/tree/master/thirdparty)

rip-nsk:
Windows build will fail on 2.8

rip-nsk:
And actually it would be nive to have 3.7 to cleanup arrow_ep by using SOURCE_SUBDIR

wesm:
Perhaps the build can error out on MSVC for CMake below a certain version. I think it's fine to only support CMake 2.8.x on Linux

rip-nsk:
Hm.. - "The Travis CI build passed"

wesm:
The Travis CI Trusty VM has cmake 3.2 installed, which is different from what comes from apt-get http://packages.ubuntu.com/trusty/cmake

wesm:
To summarize, I'm +0 on dropping cmake < 3.2 in the build. If we do drop it, we have to provide instructions how to install a newer CMake in the project README. I am interested in opinion of @xhochy and @majetideepak or others

xhochy:
I'm `-0.5` on dropping support for CMake < 3.2. Currently, we need at least 2.8.11 for gtest but all other dependencies should build with 2.8. Given the current build toolchain, cmake, boost and gcc are the main requirements, all other dependencies can be build directly through CMake's external project mechanism. 

Personally I got used to do `pip install cmake` or `conda install cmake` on all my systems but as cmake is really the "first impression" you get when building parquet-cpp (and also arrow which is quite dependent on what we do here), I would like to keep 2.8 support for a bit until we have more users (This could already be in some months).

majetideepak:
I agree with @wesm and @xhochy. Apart from some Ubuntu versions, even Centos versions come with CMake 2.8. @wesm's idea of vendoring CMake for windows is a good way to resolve this.

rip-nsk:
Goal of my proposal - avoiding duplicated code to simplify ThirdpartyToolchain.
If support of pre-3.2 cmake is critical, I can remove using of post-3.2 features (actually, AFAIK, it is only BUILD_BYPRODUCTS)
On Windows using of latest version cmake is not a problem, it is 3rdparty anyway.

wesm:
I would like to see more what is involved with getting a newer CMake on the major Linux platforms. Building cmake from source is a bit of a can of worms because of the dependence on on various system libraries. See for example the build toolchain on conda-forge: https://github.com/conda-forge/cmake-feedstock/blob/master/recipe/meta.yaml#L26

PR #327, opened 2017-05-15
 PARQUET-986: Improve MSVC build - ThirdpartyToolchain - Thrift

rip-nsk:


wesm:
@MaxRis could you take a look?

rip-nsk:
libevent is optional for thrift and is not required for parquet (IMO).
Anyway, it can't be built in default windows environment due to dependency on openssl.
Likely it should be removed.

cmake_modules/ThirdpartyToolchain.cmake
```diff
@@ -180,25 +180,18 @@ endif()
...
-  ELSE()
-    IF (MSVC)
-      set(THRIFT_STATIC_LIB "${THRIFT_PREFIX}/lib/thriftmd.lib")
-    ELSE()
-      set(THRIFT_STATIC_LIB "${THRIFT_PREFIX}/lib/libthrift.a")
-    ENDIF()
-  ENDIF()
+  set(THRIFT_STATIC_LIB_NAME "${CMAKE_STATIC_LIBRARY_PREFIX}thrift")
+  if (MSVC)
+    set(THRIFT_STATIC_LIB_NAME "${THRIFT_STATIC_LIB_NAME}${THRIFT_MSVC_STATIC_LIB_SUFFIX}")
```
MaxRis: `THRIFT_MSVC_STATIC_LIB_SUFFIX` is configurable by users and should be used to found custom thrift builds made by users. Here we build thrift from the sources and have full control of what we want to have at the output. It seems better to use hardcoded `md` suffix here for MSVC. Usage of `THRIFT_MSVC_STATIC_LIB_SUFFIX` might break the build, if user will overwrite the default value, but installed thrift will be not found.

cmake_modules/ThirdpartyToolchain.cmake
```diff
@@ -214,15 +207,27 @@ if (NOT THRIFT_FOUND)
...
+
+    set(WINFLEXBISON_PREFIX "${CMAKE_CURRENT_BINARY_DIR}/winflexbison_ep/src/winflexbison_ep-install")
+    ExternalProject_Add(winflexbison_ep
+      URL https://github.com/lexxmark/winflexbison/releases/download/v.${WINFLEXBISON_VERSION}/win_flex_bison-${WINFLEXBISON_VERSION}.zip
+      URL_HASH MD5=a2e979ea9928fbf8567e995e9c0df765
+      SOURCE_DIR ${WINFLEXBISON_PREFIX}
+      CONFIGURE_COMMAND "" BUILD_COMMAND "" INSTALL_COMMAND "")
+    set(THRIFT_DEPENDENCIES ${THRIFT_DEPENDENCIES} winflexbison_ep)
+
+    set(THRIFT_CMAKE_ARGS "-DLIBEVENT_ROOT=${CMAKE_CURRENT_BINARY_DIR}/thrift_ep-prefix/src/thrift_ep/thirdparty/src/Libevent-release-2.1.7-rc"
```
MaxRis: It might make sense to move definition of `LIBEVENT_VERSION` somewhere above and use it to set the version of libevent in path `/thrift_ep-prefix/src/thrift_ep/thirdparty/src/Libevent-release-2.1.7-rc`

MaxRis:
@rip-nsk @wesm you are right, it [seems that](https://github.com/apache/thrift/tree/master/lib/cpp) libevent is required only for libthriftnb - "This library contains the Thrift nonblocking server, which uses libevent. To link this library you will also need to link libevent." And libthriftnb is not used with parquet-cpp.

cmake_modules/ThirdpartyToolchain.cmake
rip-nsk: Agree. BTW, md it is Run-Time Library POSTFIX:
    #For visual studio the library naming is as following:
    # Dynamic libraries:
    #  - thrift.dll  for release library
    #  - thriftd.dll for debug library
    #
    # Static libraries:
    #  - thriftmd.lib for /MD release build
    #  - thriftmt.lib for /MT release build
    #
    #  - thriftmdd.lib for /MD debug build
    #  - thriftmtd.lib for /MT debug build

wesm:
Is it possible to exclude libevent and libthriftnb here? We only need serialization

rip-nsk:
Yes, but I don't have time to merge my change several times, so going to wait until master is stabilized.

wesm:
Ok -- note that rebasing is the preferred approach for resolving conflicts; merges make following the change history more difficult.

PR #329, opened 2017-05-15
PARQUET-981: Set msvc string options only if msvc build

MaxRis:

PR #330, opened 2017-05-15
PARQUET-988: Add toolchain Linux build using conda

wesm:


wesm:
@maxris does this look ok? Then you can rebase your MSVC toolchain patch on this

ci/travis_script_toolchain.sh
```diff
@@ -0,0 +1,66 @@
...
+
+conda config --set show_channel_urls True
+
+# Help with SSL timeouts to S3
+conda config --set remote_connect_timeout_secs 12
+
+conda info -a
+
+conda create -y -q -p $CPP_TOOLCHAIN \
+      boost-cpp zlib thrift-cpp snappy brotli cmake git \
```
MaxRis: Possibly `arrow-cpp` should be added here too

wesm: Ok, it's possible there will be build failures when there's an arrow API change but I'll add that for now

wesm: I'll make this build in the matrix an allowed failure

MaxRis: In pending PARQUET-981 [changelist](https://github.com/apache/parquet-cpp/pull/329) I've put [specific versions](https://github.com/apache/parquet-cpp/pull/329/files#diff-180360612c6b8c4ed830919bbb4dd459R43) of libs for "Toolchain" Appveyor configuration
`conda install -y boost-cpp=1.63 arrow-cpp=0.3.0 brotli=0.6.0 zlib=1.2.11 snappy=1.1.4 thrift-cpp=0.10.0`

wesm:
Merging. @MaxRis can you rebase your patch after this lands in master?

PR #331, opened 2017-05-15
PARQUET-989: [C++] Fix toolchain Travis CI failure

wesm:
I mistakenly thought that the toolchain build had passed in #330, but when I added arrow-cpp it introduced failures.  I dropped the `allow_failures` to require the toolchain build to pass to see build success on GitHub. @majetideepak if you could take a look and let me know if this is OK (assuming the Travis CI build passes).


majetideepak:
@wesm This is OK.

majetideepak:
+1 LGTM

wesm:
There appears to be an ABI conflict when between gcc 4.9 in Travis CI and the devtoolset-2 arrow-cpp package (???), so I've removed that from here for now.

PR #333, opened 2017-05-17
PARQUET-992: Do not transitively include zlib.h in public API

wesm:
Use PIMPL pattern to hide zlib from public API


wesm:
These builds are running now in https://ci.appveyor.com/project/wesm/parquet-cpp/build/1.0.12 -- can merge as soon as they are green

wesm:
Build has passed, merging

PR #337, opened 2017-05-18
PARQUET-995: Use sizeof(Int96) instead of Int96Type

wesm:


wesm:
Added a test case that had a valgrind error showing the bad memory access

src/parquet/arrow/arrow-reader-writer-test.cc
```diff
@@ -1159,6 +1159,26 @@ TEST_F(TestNestedSchemaRead, ReadTablePartial) {
   ASSERT_EQ(table->schema()->field(0)->type()->num_children(), 0);
 }
 
+TEST(TestArrowReaderAdHoc, NationDictTruncatedDataPage) {
```
wesm: I need to change the name of this test

xhochy:
Green build: https://ci.appveyor.com/project/wesm/parquet-cpp/build/1.0.16

PR #338, opened 2017-05-21
PARQUET-997: Fix override compiler warnings

cpcloud:


CMakeLists.txt
```diff
@@ -402,11 +408,13 @@ endif ()
...
 
+
 set(CMAKE_CXX_FLAGS "${PARQUET_CXXFLAGS} ${CMAKE_CXX_FLAGS}")
+
 if (MSVC)
-  set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -W3")
+  set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} /W3 /WX")
 else()
-  set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -std=c++11 -fno-strict-aliasing -Wall")
+  set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -fno-strict-aliasing -Wall -Werror")
```
wesm: We don't want /WX and /Werror to be in there by default; these should be passed with `$ARROW_CXXFLAGS` in CI builds, though

cpcloud: Cool, I'll change.

cpcloud: Hm. It's a little mysterious why this wasn't failing since `-Werror` is being passed in to CI builds.

wesm: Only on Linux: https://github.com/apache/arrow/blob/master/ci/travis_before_script_cpp.sh#L37. We should have a warning-free build on clang

wesm:
The current set of warnings in MSVC is a bit out of control, and speaks to code changes that need to be made. I think we should try to have a warning-free build with `/W3`

wesm:
I think @MaxRis is looking at the compiler warnings -- i created a JIRA about it in https://issues.apache.org/jira/browse/PARQUET-991. We have been building Arrow with `/W3 /WX` with no problems for a while now

rip-nsk:
Ideally it should be clean with /Wall, but for beginning,
it would be better clean "priority-1" warnings, which are buried by tons others for now.

wesm:
A large portion of them can be fixed by cleaning up the extern templates (https://github.com/apache/arrow/blob/master/cpp/src/arrow/util/visibility.h#L41). I don't think it's as bad as it looks

cpcloud:
@rip-nsk Please make your comments in the actual diff, alongside the code. Reviewing in a GitHub comment thread is not as systematic as commenting on specific lines of code.

cpcloud:
@wesm Is this good to go?

MaxRis:
I'm still looking to resolve https://issues.apache.org/jira/browse/PARQUET-991 with currently used warnings level, should be ready soon

kou:
`CMAKE_CXX_STANDARD` is added since CMake 3.1: https://cmake.org/cmake/help/latest/release/3.1.html#properties

Debian GNU/Linux jessie ships CMake 3.0.2: https://packages.debian.org/jessie/cmake

Is it OK to us that we drop Debian GNU/Linux jessie support? Or should we support CMake 3.0?

kou:
Workaround for CMake 3.0.2: `cmake ... -DCMAKE_CXX_FLAGS="-std=gnu++11"`

PR #340, opened 2017-05-24
PARQUET-991: Resolve msvc warnings; Appveyor treats msvc warnings as …

MaxRis:
…errors (/WX flag)

### DIFF
--- a/src/parquet/arrow/writer.cc
+++ b/src/parquet/arrow/writer.cc
@@ -63,7 +63,7 @@ class LevelBuilder {
   Status VisitInline(const Array& array);
 
   Status Visit(const ::arrow::PrimitiveArray& array) {
-    array_offsets_.push_back(array.offset());
+    array_offsets_.push_back(static_cast<int32_t>(array.offset()));
     valid_bitmaps_.push_back(array.null_bitmap_data());
     null_counts_.push_back(array.null_count());
     values_type_ = array.type_id();
@@ -72,7 +72,7 @@ class LevelBuilder {
   }
 
   Status Visit(const ::arrow::BinaryArray& array) {
-    array_offsets_.push_back(array.offset());
+    array_offsets_.push_back(static_cast<int32_t>(array.offset()));
     valid_bitmaps_.push_back(array.null_bitmap_data());
     null_counts_.push_back(array.null_count());
     values_type_ = array.type_id();
@@ -81,7 +81,7 @@ class LevelBuilder {
   }
 
   Status Visit(const ListArray& array) {
-    array_offsets_.push_back(array.offset());
+    array_offsets_.push_back(static_cast<int32_t>(array.offset()));
     valid_bitmaps_.push_back(array.null_bitmap_data());
     null_counts_.push_back(array.null_count());
     offsets_.push_back(array.raw_value_offsets());
@@ -111,7 +111,7 @@ class LevelBuilder {
       std::shared_ptr<Buffer>* rep_levels, const Array** values_array) {
     // Work downwards to extract bitmaps and offsets
     min_offset_idx_ = 0;
-    max_offset_idx_ = array.length();
+    max_offset_idx_ = static_cast<int32_t>(array.length());
     RETURN_NOT_OK(VisitInline(array));
     *num_values = max_offset_idx_ - min_offset_idx_;
     *values_offset = min_offset_idx_;
@@ -143,7 +143,7 @@ class LevelBuilder {
           std::fill(def_levels_ptr, def_levels_ptr + array.length(), 1);
         } else {
           const uint8_t* valid_bits = array.null_bitmap_data();
-          INIT_BITSET(valid_bits, array.offset());
+          INIT_BITSET(valid_bits, static_cast<int>(array.offset()));
           for (int i = 0; i < array.length(); i++) {
             if (bitset_valid_bits & (1 << bit_offset_valid_bits)) {
               def_levels_ptr[i] = 1;
@@ -396,7 +396,7 @@ Status FileWriter::Impl::WriteNullableBatch(TypedColumnWriter<ParquetType>* writ
 
   RETURN_NOT_OK(data_buffer_.Resize(num_values * sizeof(ParquetCType)));
   auto buffer_ptr = reinterpret_cast<ParquetCType*>(data_buffer_.mutable_data());
-  INIT_BITSET(valid_bits, valid_bits_offset);
+  INIT_BITSET(valid_bits, static_cast<int>(valid_bits_offset));
   for (int i = 0; i < num_values; i++) {
     if (bitset_valid_bits & (1 << bit_offset_valid_bits)) {
       buffer_ptr[i] = static_cast<ParquetCType>(data_ptr[i]);
@@ -417,7 +417,7 @@ Status FileWriter::Impl::WriteNullableBatch<Int32Type, ::arrow::Date64Type>(
     const int64_t* data_ptr) {
   RETURN_NOT_OK(data_buffer_.Resize(num_values * sizeof(int32_t)));
   auto buffer_ptr = reinterpret_cast<int32_t*>(data_buffer_.mutable_data());
-  INIT_BITSET(valid_bits, valid_bits_offset);
+  INIT_BITSET(valid_bits, static_cast<int>(valid_bits_offset));
   for (int i = 0; i < num_values; i++) {
     if (bitset_valid_bits & (1 << bit_offset_valid_bits)) {
       // Convert from milliseconds into days since the epoch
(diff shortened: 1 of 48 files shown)


src/parquet/column/test-specialization.h
```diff
@@ -115,7 +115,8 @@ void PrimitiveTypedTest<TestType>::SyncValuesOut() {}
 
 template <>
 void PrimitiveTypedTest<BooleanType>::SyncValuesOut() {
-  std::copy(bool_buffer_out_.begin(), bool_buffer_out_.end(), values_out_.begin());
+  std::transform(bool_buffer_out_.begin(), bool_buffer_out_.end(), values_out_.begin(),
```
MaxRis: Change made to overcome MSVC warning [c4800](https://msdn.microsoft.com/en-us/library/b6801kcy.aspx). Warning happens on assigning of bool to non-boolean value. Type casting doesn't affect appearance of this warning.
As alternative solution, we can just suppress this kind of warnings, like this done in some open source projects.

src/parquet/arrow/writer.cc
```diff
@@ -63,7 +63,7 @@ class LevelBuilder {
   Status VisitInline(const Array& array);
 
   Status Visit(const ::arrow::PrimitiveArray& array) {
-    array_offsets_.push_back(array.offset());
+    array_offsets_.push_back(static_cast<int>(array.offset()));
```
wesm: We should use a more precise type here, like `int32_t`?

src/parquet/arrow/writer.cc
```diff
@@ -497,7 +497,7 @@ Status FileWriter::Impl::TypedWriteBatch<BooleanType, ::arrow::BooleanType>(
   auto writer = reinterpret_cast<TypedColumnWriter<BooleanType>*>(column_writer);
 
   int buffer_idx = 0;
-  int32_t offset = array->offset();
+  int32_t offset = static_cast<int32_t>(array->offset());
```
wesm: Let's use `int64_t` here instead. In the instances above we should maybe use int64 also, @xhochy what do you think?

src/parquet/column/levels.cc
```diff
@@ -36,7 +36,8 @@ void LevelEncoder::Init(Encoding::type encoding, int16_t max_level,
       break;
     }
     case Encoding::BIT_PACKED: {
-      int num_bytes = BitUtil::Ceil(num_buffered_values * bit_width_, 8);
+      int num_bytes =
+          static_cast<int>(BitUtil::Ceil(num_buffered_values * bit_width_, 8));
```
wesm: Use more precise integer type here (e.g. `int64_t`)?

src/parquet/column/reader.cc
```diff
@@ -169,7 +169,8 @@ bool TypedColumnReader<DType>::ReadNewPage() {
             throw ParquetException("Unknown encoding type.");
         }
       }
-      current_decoder_->SetData(num_buffered_values_, buffer, data_size);
+      current_decoder_->SetData(
+          num_buffered_values_, buffer, static_cast<int>(data_size));
```
wesm: We should use a more precise integer type in `SetData`, and the decoders. We can leave this as is though

src/parquet/column/reader.h
```diff
@@ -102,7 +102,7 @@ class PARQUET_EXPORT ColumnReader {
 
 // API to read values from a single column. This is the main client facing API.
 template <typename DType>
-class PARQUET_EXPORT TypedColumnReader : public ColumnReader {
+class TypedColumnReader : public ColumnReader {
```
wesm: Is this right?

src/parquet/column/reader.h
```diff
@@ -383,10 +383,10 @@ inline int64_t TypedColumnReader<DType>::Skip(int64_t num_rows_to_skip) {
 
       do {
         batch_size = std::min(batch_size, rows_to_skip);
-        values_read =
-            ReadBatch(batch_size, reinterpret_cast<int16_t*>(def_levels->mutable_data()),
-                reinterpret_cast<int16_t*>(rep_levels->mutable_data()),
-                reinterpret_cast<T*>(vals->mutable_data()), &values_read);
+        values_read = ReadBatch(static_cast<int>(batch_size),
```
wesm: I think we should make `ReadBatch` take `int64_t`, but in a separate patch

src/parquet/column/statistics-test.cc
```diff
@@ -145,8 +145,10 @@ class TestRowGroupStatistics : public PrimitiveTypedTest<TestType> {
     // simulate the case when data comes from multiple buffers,
     // in which case special care is necessary for FLBA/ByteArray types
     for (int i = 0; i < 2; i++) {
-      int batch_num_values = i ? num_values - num_values / 2 : num_values / 2;
-      int batch_null_count = i ? null_count : 0;
+      int batch_num_values =
+          i ? static_cast<int>(num_values) - static_cast<int>(num_values) / 2
+            : static_cast<int>(num_values) / 2;
+      int batch_null_count = i ? static_cast<int>(null_count) : 0;
```
wesm: Use `int64_t` here to avoid these casts?

src/parquet/column/test-specialization.h
wesm: We could also use a for loop here with a cast. Since this is test code I don't think it matters too much

src/parquet/column/writer.cc
```diff
@@ -171,7 +171,7 @@ void ColumnWriter::AddDataPage() {
 
 void ColumnWriter::WriteDataPage(const CompressedDataPage& page) {
   int64_t bytes_written = pager_->WriteDataPage(page);
-  total_bytes_written_ += bytes_written;
+  total_bytes_written_ += static_cast<int>(bytes_written);
```
wesm: We should make `total_bytes_written_` either int32_t or int64_t. If it's 64-bit in the Thrift metadata, then int64_t

src/parquet/column/writer.cc
```diff
@@ -377,8 +377,8 @@ inline int64_t TypedColumnWriter<DType>::WriteMiniBatch(int64_t num_values,
     page_statistics_->Update(values, values_to_write, num_values - values_to_write);
   }
 
-  num_buffered_values_ += num_values;
-  num_buffered_encoded_values_ += values_to_write;
+  num_buffered_values_ += static_cast<int>(num_values);
+  num_buffered_encoded_values_ += static_cast<int>(values_to_write);
```
wesm: Can we change these members to `int64_t`?

src/parquet/reader-test.cc
```diff
@@ -88,13 +88,15 @@ TEST_F(TestAllTypesPlain, TestBatchRead) {
 
   ASSERT_TRUE(col->HasNext());
   int64_t values_read;
-  int levels_read = col->ReadBatch(4, def_levels, rep_levels, values, &values_read);
+  int levels_read =
+      static_cast<int>(col->ReadBatch(4, def_levels, rep_levels, values, &values_read));
```
wesm: Could also use `auto` or `int64_t` here to avoid cast

src/parquet/util/bit-stream-utils.inline.h
```diff
@@ -227,7 +249,7 @@ inline bool BitReader::GetZigZagVlqInt(int32_t* v) {
   int32_t u_signed;
   if (!GetVlqInt(&u_signed)) return false;
   uint32_t u = static_cast<uint32_t>(u_signed);
-  *reinterpret_cast<uint32_t*>(v) = (u >> 1) ^ -(u & 1);
+  *reinterpret_cast<uint32_t*>(v) = (u >> 1) ^ -(static_cast<int32_t>(u & 1));
```
wesm: Should this be `uint32_t`?

src/parquet/util/cpu-info.cc
```diff
@@ -148,7 +148,7 @@ void CpuInfo::Init() {
 #endif
 
   if (max_mhz != 0) {
-    cycles_per_ms_ = max_mhz * 1000;
+    cycles_per_ms_ = static_cast<int64_t>(max_mhz) * 1000;
   } else {
     cycles_per_ms_ = 1000000;
```
wesm: Does getting the cache sizes in `Init` work in MSVC?

wesm:
Looks like the coverage build failed but otherwise the Linux/macOS tests passed

src/parquet/column/reader.h
MaxRis: Thank you for catching this. Having `PARQUET_EXPORT` in this place triggers msvc compilation error that extern template class is not compatible with __dllexport. I'm going to solve this by structure the code in similar way how it's done in Arrow: separate header file type_fwd.h will contain forward definition of `TypedColumnReader` without `PARQUET_EXPORT`, and this forward declaration will be resolved by compiler on processing of extern template class definition. So, no warning generated.

src/parquet/arrow/writer.cc
MaxRis: @xhochy , could you please help with above :)

src/parquet/column/levels.cc
MaxRis: it seems not a lot sense here, since later `num_bytes` is passed to `BitWriter` constructor, which accepts it as `int`

src/parquet/column/reader.h
MaxRis: [PARQUET-1008](https://issues.apache.org/jira/browse/PARQUET-1008) created to address this

src/parquet/util/cpu-info.cc
MaxRis: it doesn't, have created [ticket](https://issues.apache.org/jira/browse/PARQUET-1009)

MaxRis:
@wesm , have updated changelist, it seems coverage passed without any special changes

PR #341, opened 2017-05-25
PARQUET-978: [C++] Minimizing footer reads for small(ish) metadata

majetideepak:
I realized we don't have a full end to end `writer-reader-test` for parquet. Would have been easier to test this PR. Does it make sense adding one?

### DIFF
--- a/src/parquet/file/reader-internal.cc
+++ b/src/parquet/file/reader-internal.cc
@@ -268,8 +268,8 @@ void SerializedFile::ParseMetaData() {
     throw ParquetException("Invalid parquet file. Corrupt footer.");
   }
 
-  uint32_t metadata_len = *reinterpret_cast<uint32_t*>(
-                               footer_buffer + footer_read_size - FOOTER_SIZE);
+  uint32_t metadata_len =
+      *reinterpret_cast<uint32_t*>(footer_buffer + footer_read_size - FOOTER_SIZE);
   int64_t metadata_start = file_size - FOOTER_SIZE - metadata_len;
   if (FOOTER_SIZE + metadata_len > file_size) {
     throw ParquetException(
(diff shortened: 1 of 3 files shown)


src/parquet/file/reader-internal.cc
```diff
@@ -255,15 +257,19 @@ void SerializedFile::ParseMetaData() {
     throw ParquetException("Corrupted file, smaller than file footer");
   }
 
-  uint8_t footer_buffer[FOOTER_SIZE];
+  uint8_t footer_buffer[DEFAULT_FOOTER_READ_SIZE];
```
wesm: is 64K large for the stack?

src/parquet/file/reader-internal.cc
```diff
@@ -273,10 +279,17 @@ void SerializedFile::ParseMetaData() {
...
       AllocateBuffer(properties_.memory_pool(), metadata_len);
-  bytes_read =
-      source_->ReadAt(metadata_start, metadata_len, metadata_buffer->mutable_data());
-  if (bytes_read != metadata_len) {
-    throw ParquetException("Invalid parquet file. Could not read metadata bytes.");
+
+  // Check if the footer_buffer contains the entire metadata
+  if (footer_read_size >= (metadata_len + FOOTER_SIZE)) {
+    memcpy(metadata_buffer->mutable_data(),
+        footer_buffer + (footer_read_size - metadata_len - FOOTER_SIZE), metadata_len);
```
wesm: This memcpy can be avoided by doing the initial read into allocated memory rather than stack memory. If any buffer slicing needs to be performed, this can be done with zero copy using `arrow::SliceBuffer`. But I don't think this is a big deal either way

PR #342, opened 2017-05-26
[PARQUET-1004] Building fails on windows with PARQUET_ZLIB_VENDORED is turned on

atkawa7:


wesm:
Is this issue not addressed by the `ZLIB_MSVC_STATIC_LIB_SUFFIX` CMake option?

atkawa7:
@wesm Thanks.  I didn't see that option. It  would be great if that is added to the docs

wesm:
@MaxRis could you add a note to the README/build instructions about this?

MaxRis:
@wesm , sure, here it's [PR](https://github.com/apache/parquet-cpp/pull/345)

PR #344, opened 2017-05-26
PARQUET-967: Combine libparquet, libparquet_arrow libraries

wesm:


CMakeLists.txt
```diff
@@ -609,6 +609,10 @@ set(LIBPARQUET_SRCS
   src/parquet/exception.cc
   src/parquet/types.cc
 
+  src/parquet/arrow/reader.cc
```
majetideepak: Can we make this optional with `PARQUET_ARROW` flag?

wesm: Ah I need to remove the PARQUET_ARROW flag. I'm -0 on making it optional as the dependencies are unchanged  and it doesn't have a meaningful impact on packaging or linking.

majetideepak: This makes sense. Thanks!

PR #346, opened 2017-05-30
PARQUET-999: Improve MSVC build - Enable PARQUET_BUILD_BENCHMARKS

rip-nsk:


rip-nsk:
How can I find out what is wrong here? job log is empty..

PR #350, opened 2017-06-08
PARQUET-1003: Modify DEFAULT_CREATED_BY value for every new release v…

majetideepak:
…ersion


src/parquet/parquet_version.h
```diff
@@ -0,0 +1,24 @@
...
+// specific language governing permissions and limitations
+// under the License.
+
+#ifndef PARQUET_VERSION_H
+#define PARQUET_VERSION_H
+
+// define the parquet created by version
+#define CREATED_BY_VERSION "parquet-cpp version 1.1.1-SNAPSHOT"
+
+#endif // PARQUET_VERSION_H
```
wesm: Is it typical to check the generated file in?

PR #352, opened 2017-06-13
PARQUET-1029: [C++] Some extern template symbols not being exported in gcc

wesm:
Extern template visibility contributes to be a mystery to me, but this fixes the regression from PARQUET-991.

cc @saatvikshah1994

### DIFF
--- a/examples/CMakeLists.txt
+++ b/examples/CMakeLists.txt
@@ -17,5 +17,5 @@
 
 if (PARQUET_BUILD_EXECUTABLES)
   add_executable(reader-writer reader-writer.cc)
-  target_link_libraries(reader-writer parquet_static)
+  target_link_libraries(reader-writer parquet_shared arrow)
 endif()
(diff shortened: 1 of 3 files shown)


wesm:
This regression was hidden because all of our tests and executables use static linking. I changed `examples/reader-writer.cc` to use dynamic linking and that exposed the linking issue

majetideepak:
+1 LGTM

MaxRis:
@wesm , in changelist which caused the symbols to be hidden, I've [disabled](https://github.com/apache/parquet-cpp/commit/fc5228af3eee2ec8176e404ecb34b7ba985d0e4d#diff-66be436bc1d7bc37dee806a5ae15299eR30) msvc warning for incompatibility between `extern` and `_dllexport` for the same exported symbol record. It seems that even with warning the symbols finally are exported by msvc. But during local testing by making some unit tests to be linked dynamically, I have seen linkage errors for others symbols. It seems appveyor build might fail for unit test. It might make sense to apply dynamic linking for non-msvc platforms and resolve msvc issues in separate changelist, since it might be longer story.

wesm:
So it seems that there isn't consistency when gcc doesn't like the extra `PARQUET_EXPORT` attribute on the extern template declaration. I think it might be because these templates have a number of explicit specializations, but I'm no expert.

wesm:
Appveyor build is OK: https://ci.appveyor.com/project/wesm/parquet-cpp/build/1.0.28

PR #353, opened 2017-06-13
PARQUET-1007: Update parquet.thrift

majetideepak:


src/parquet/parquet.thrift
```diff
@@ -177,6 +188,12 @@ enum ConvertedType {
    */
   INTERVAL = 21;
 
+  /**
+   * Annotates a column that is always null
+   * Sometimes when discovering the schema of existing data
+   * values are always null
+   */
+  NULL = 25;
```
xhochy: Temporarily change this to NA so we can build it in C++?

majetideepak: Yes! Will update shortly.

PR #356, opened 2017-06-20
PARQUET-1035 Write Int96 from Arrow timestamp(ns)

c-nichols:


src/parquet/arrow/writer.h
```diff
@@ -94,6 +94,12 @@ ::arrow::Status PARQUET_EXPORT WriteTable(const ::arrow::Table& table,
     int64_t chunk_size,
     const std::shared_ptr<WriterProperties>& properties = default_writer_properties());
 
+/**
+ * Helper function for converting nanosecond timestamps to Impala (Int96) format
+ */
+inline void nanosecondsToImpalaTimestamp(const int64_t nanoseconds,
```
wesm: You can put the whole implementation here, but put it in an `parquet::arrow::internal` namespace. Also use `NanosecondsToImpalaTimestamp`

wesm:
@xhochy I'll let you also take a look

src/parquet/arrow/schema.cc
```diff
@@ -485,14 +485,19 @@ Status FieldToNode(const std::shared_ptr<Field>& field,
...
       auto timestamp_type = static_cast<::arrow::TimestampType*>(field->type().get());
       auto unit = timestamp_type->unit();
-      type = ParquetType::INT64;
       if (unit == ::arrow::TimeUnit::MILLI) {
+        type = ParquetType::INT64;
         logical_type = LogicalType::TIMESTAMP_MILLIS;
       } else if (unit == ::arrow::TimeUnit::MICRO) {
+        type = ParquetType::INT64;
         logical_type = LogicalType::TIMESTAMP_MICROS;
+      } else if (unit == ::arrow::TimeUnit::NANO) {
```
xhochy: As weird as a it sounds, I would like to have a flag that needs to be set to true to write int96 timestamps, e.g. `support_deprecated_int96_timestamps`. Otherwise e.g. timestamps coming from Pandas will always be saved to Int96. That would be quite destructive in the deprecation of Int96.

wesm: I was thinking that we'd create an `ArrowWriterProperties` where:

* default is write nanos to int64 with casting to micros
* add option to coerce nanoseconds to millis or micros
* add option to permit unsafe casts (e.g. losing nanoseconds), default to raise if nanoseconds will be dropped

It should be like https://github.com/apache/parquet-cpp/blob/master/src/parquet/column/properties.h, and not break the current API. @c-nichols would you mind starting something simple? @xhochy we can also do this in a follow up patch

c-nichols: Unfortunately I am leaving for vacation (without computer) tonight until July 5 :) happy to proceed however you guys think is best.

wesm: No problem, we'll take over the patch

c-nichols: Thank you!

xhochy: @wesm Implemented your suggestion in https://github.com/apache/parquet-cpp/pull/371

PR #358, opened 2017-06-23
PARQUET-1041: Support Arrow's NullArray

xhochy:
This also fixes a segmentation fault on reading a duplicate column.

### DIFF
--- a/src/parquet/arrow/reader.cc
+++ b/src/parquet/arrow/reader.cc
@@ -495,10 +495,6 @@ Status FileReader::Impl::ReadTable(
   std::shared_ptr<::arrow::Schema> schema;
   RETURN_NOT_OK(GetSchema(indices, &schema));
 
-  int num_fields = static_cast<int>(schema->num_fields());
-  int nthreads = std::min<int>(num_threads_, num_fields);
-  std::vector<std::shared_ptr<Column>> columns(num_fields);
-
   // We only need to read schema fields which have columns indicated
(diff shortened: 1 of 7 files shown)


xhochy:
cc @crepererum

src/parquet/arrow/reader.cc
```diff
@@ -1262,6 +1260,10 @@ Status PrimitiveImpl::NextBatch(
   }
 
   switch (field_->type()->id()) {
+    case ::arrow::Type::NA:
+      *out = std::make_shared<::arrow::NullArray>(batch_size);
+      return Status::OK();
+      break;
```
wesm: This `break` is unreachable, any reason to keep it?

src/parquet/types.h
```diff
@@ -81,7 +81,8 @@ struct LogicalType {
     INT_64,
     JSON,
     BSON,
-    INTERVAL
+    INTERVAL,
+    NA = 25
```
wesm: Is this from parquet.thrift?

xhochy: Yes

src/parquet/arrow/reader.cc
xhochy: Yes, getting rid of compiler warnings.

wesm: gotcha, I hadn't seen that before

PR #359, opened 2017-06-23
PARQUET-1042: Fix Compilation breaks on GCC 4.8

xhochy:
Not the most beautiful, but verified locally in docker.


wesm:
Can you instead make `Impl` protected scope?

PR #360, opened 2017-06-23
PARQUET-1041: Support Arrow's NullArray

wesm:
Closes #358. This only includes an Arrow version bump to pick up ARROW-1143


wesm:
Appveyor looking good https://ci.appveyor.com/project/wesm/parquet-cpp/build/1.0.34. Merging so we can get Arrow PRs back to green

PR #362, opened 2017-06-23
PARQUET-1044: Use compression libraries from Apache Arrow

wesm:
Depends on https://github.com/apache/arrow/pull/771; will update Arrow version once that's merged


wesm:
If @advancedxy, @majetideepak, or others could take a look when they get a chance, that would be great! Thanks

src/parquet/util/memory.h
```diff
@@ -31,35 +31,38 @@
...
-    if (!_s.ok()) {                                 \
-      std::stringstream ss;                         \
-      ss << "Arrow error: " << _s.ToString();       \
-      ::parquet::ParquetException::Throw(ss.str()); \
-    }                                               \
-  } while (0);
-
 namespace parquet {
 
+static inline std::unique_ptr<::arrow::Codec> GetArrowCodec(Compression::type codec) {
```
advancedxy: How about change to `GetCodecFromArrow` and reuse use codec_type?

ArrowCodec seems a bit weird..

wesm: Sure, no problem

wesm:
Sure, I'll back out the clang-format changes and get the build passing

benchmarks/decode_benchmark.cc
```diff
@@ -277,19 +278,20 @@ uint64_t TestBinaryPackedEncoding(const char* name, const std::vector<int64_t>&
...
   const uint8_t* raw_data = reinterpret_cast<const uint8_t*>(&data[0]);
   int uncompressed_len = data.size() * sizeof(int64_t);
   uint8_t* decompressed_data = new uint8_t[uncompressed_len];
 
   int max_compressed_size = codec->MaxCompressedLen(uncompressed_len, raw_data);
   uint8_t* compressed_data = new uint8_t[max_compressed_size];
-  int compressed_len =
-      codec->Compress(uncompressed_len, raw_data, max_compressed_size, compressed_data);
+  int64_t compressed_len;
+  codec->Compress(
```
xhochy: Unchecked status?

wesm:
I opened https://issues.apache.org/jira/browse/ARROW-1150 about the compiler warning that showed up in AdaptiveIntBuilder. Interesting that it only happened in the debug build on MSVC

benchmarks/decode_benchmark.cc
wesm: Fixed

wesm:
Here's the Appveyor build: https://ci.appveyor.com/project/wesm/parquet-cpp/build/1.0.43.

majetideepak:
I am seeing parquet test linking issues after this change when statically linking with libarrow.
```
libarrow.a(compression.cc.o): In function `arrow::SnappyCodec::Decompress(long, unsigned char const*, long, unsigned char*)':
arrow/cpp/src/arrow/util/compression.cc:282: undefined reference to `snappy::RawUncompress(char const*, unsigned long, char*)'
libarrow.a(compression.cc.o): In function `arrow::SnappyCodec::MaxCompressedLen(long, unsigned char const*)':
arrow/cpp/src/arrow/util/compression.cc:289: undefined reference to `snappy::MaxCompressedLength(unsigned long)'
....................
....................
```
```
nm libarrow.a | grep snappy
                 U _ZN6snappy11RawCompressEPKcmPcPm
                 U _ZN6snappy13RawUncompressEPKcmPc
                 U _ZN6snappy19MaxCompressedLengthEm
```
How do I get the arrow link line to parquet?

wesm:
Now that snappy is in the Arrow toolchain, if you're statically linking you'll have to also statically link libsnappy.a. I guess we should not support static linking in `libparquet.so` now

PR #364, opened 2017-06-26
PARQUET-1045: Remove code that's being moved to Apache Arrow in ARROW-1154

wesm:
I verified locally that this works on Linux with the current iteration of https://github.com/apache/arrow/pull/785, will update the commit in ThirdpartyToolchain.cmake after that is merged to get a passing build


wesm:
Just updated the Arrow version after merging ARROW-1154

wesm:
Small Windows buglet that I'll have to fix in Arrow

PR #366, opened 2017-07-03
PARQUET-1048: Restore Apache Arrow static transitive dependencies to thirdparty toolchain 

wesm:


cmake_modules/ThirdpartyToolchain.cmake
```diff
@@ -18,6 +18,8 @@
 set(GTEST_VERSION "1.8.0")
 set(GBENCHMARK_VERSION "1.1.0")
 set(THRIFT_VERSION "0.10.0")
+set(SNAPPY_VERSION "1.1.3")
+set(BROTLI_VERSION "v0.6.0")
```
rip-nsk: #minor
set(BROTLI_VERSION "0.6.0")
?

cmake_modules/ThirdpartyToolchain.cmake
```diff
@@ -360,26 +357,139 @@ if (NOT ARROW_FOUND)
...
+  ADD_THIRDPARTY_LIB(brotli_dec
+    STATIC_LIB ${BROTLI_STATIC_LIBRARY_DEC})
+  add_dependencies(brotli_dec brotli_ep)
+
+  ADD_THIRDPARTY_LIB(brotli_common
+    STATIC_LIB ${BROTLI_STATIC_LIBRARY_COMMON})
+  add_dependencies(brotli_common brotli_ep)
+
+  if (NOT DEFINED ENV{SNAPPY_HOME})
+    set(ENV{SNAPPY_HOME} "${SNAPPY_HOME}")
```
majetideepak: This should be `set(SNAPPY_HOME "$ENV{SNAPPY_HOME}")`

cmake_modules/ThirdpartyToolchain.cmake
```diff
@@ -360,26 +357,139 @@ if (NOT ARROW_FOUND)
...
+  endif()
+
+  if (NOT DEFINED ENV{BROTLI_HOME})
+    set(ENV{BROTLI_HOME} "${BROTLI_HOME}")
```
majetideepak: Same as above.

cmake_modules/ThirdpartyToolchain.cmake
wesm: This is here so that the libraries aren't built multiple times (once in parquet-cpp, then again in arrow_ep). I can modify this to use toolchain libraries

wesm:
I ran into a bit of a roadblock as it seems that the linker expects to resolve all transitive dependencies even if they are not used in the final binary. Since we just added ZSTD compressor to Arrow, we now have:

```
debug/libarrow.a(compression.cc.o): In function `arrow::ZSTDCodec::Decompress(long, unsigned char const*, long, unsigned char*)':
/home/wesm/code/parquet-cpp/toolchain-build/arrow_ep-prefix/src/arrow_ep/cpp/src/arrow/util/compression.cc:339: undefined reference to `ZSTD_decompress'
debug/libarrow.a(compression.cc.o): In function `arrow::ZSTDCodec::MaxCompressedLen(long, unsigned char const*)':
/home/wesm/code/parquet-cpp/toolchain-build/arrow_ep-prefix/src/arrow_ep/cpp/src/arrow/util/compression.cc:347: undefined reference to `ZSTD_compressBound'
debug/libarrow.a(compression.cc.o): In function `arrow::ZSTDCodec::Compress(long, unsigned char const*, long, unsigned char*, long*)':
/home/wesm/code/parquet-cpp/toolchain-build/arrow_ep-prefix/src/arrow_ep/cpp/src/arrow/util/compression.cc:353: undefined reference to `ZSTD_compress'
/home/wesm/code/parquet-cpp/toolchain-build/arrow_ep-prefix/src/arrow_ep/cpp/src/arrow/util/compression.cc:354: undefined reference to `ZSTD_isError'
collect2: error: ld returned 1 exit status
```

In light of this, and given the relatively little packaging and build system support we're getting from the community compared with our feature development efforts, and unless anyone knows how to make a "super archive" of all the transitive static dependencies in Arrow, I think we should nix the static linking feature in parquet-cpp and instead expect users to ship `libarrow.so` and `libparquet.so` together.

xhochy:
+1 on making it only via a "super archive" in Arrow. I'm not really sure how to integrate it into CMake but basically, you would have to run the following commands on Unix:

```
ar -x libarrow.a
ar -x libzstd.a
…
ar -c libarrow_fat.a  *.o
```

majetideepak:
I don't think a "super archive" in Arrow is a practical solution.
How will a super archive work with users who have their own third-party (compression) libraries? If a codebase depends on only `libsnappy.a` in another module, are they expected to link with `libarrow_fat.a`?

majetideepak:
Vertica prefers static linking of the libraries. Dynamic linking is complicated in our build process.
The main problem with static linking here is due to the re-use of the compression library interface between Arrow and Parquet. Re-use make sense if the same code is being used. But once Arrow and Parquet have non-overlapping compression library dependencies (ZSTD currently), I don't think re-use is helpful anymore.
I would vote for adding the compression library interface back inside Parquet.
Practically, this code is not going to change much and it seems reasonable to keep it inside both Arrow and Parquet.

wesm:
Re-introducing the compression libraries in Parquet won't fix the problem as is. If `libarrow.a` has any object code that depends on any other libraries (like `libzstd.a`) then they will be required in the static link step with libparquet.a.

I am confident that a solution to this problem exists (using tools for dealing with multiple CMake projects -- for example we are not exporting any CMake targets from Arrow), but there's a limit to how much time I can personally spend on it. For example, if Arrow can export multiple "leaf" static libraries and make these visible to parquet-cpp build system then that would help, this is something that Apache Kudu does in its various subdirectories https://github.com/apache/kudu/blob/master/src/kudu/common/CMakeLists.txt

wesm:
We could also add an option to Arrow's build system to create a minimal build for use in Parquet. It's not ideal but maybe the quickest fix

majetideepak:
Transitively, `libparquet.a` now requires `libzstd.a`. I don't see a way out of this.
I will take a look at the export solution used in Apache Kudu.

wesm:
Yeah, I think this PR should be closed. I think the best way right now is to add an Arrow build option to produce a minimal libarrow.a containing only transitive dependencies that are used in Parquet. I assume you have been statically linking libparquet.a and it's transitive dependencies in a different build system, so I'm not sure we should be spending our energy producing a statically linked libparquet.so

majetideepak:
I agree that a minimal `libarrow.a` solution is ideal for the time being. I am filing an ARROW JIRA.

wesm:
Sounds good. Further, I think we should remove the `PARQUET_ARROW_LINKAGE` option because it only affects `libparquet.so` and nothing else AFAIK. So if you're static linking in another project you have to deal with the transitive dependencies either way

majetideepak:
I am finding it more convenient to use/share the C++ Parquet tools and examples via static linking.
As a temporary measure, can we enforce users to provide the missing transitive dependencies for static linking? We just need 5 values and I have a patch ready.

```
set(BROTLI_STATIC_LIB_ENC "$ENV{BROTLI_STATIC_LIB_ENC}")
set(BROTLI_STATIC_LIB_DEC "$ENV{BROTLI_STATIC_LIB_DEC}")
set(BROTLI_STATIC_LIB_COMMON "$ENV{BROTLI_STATIC_LIB_COMMON}")
set(SNAPPY_STATIC_LIB "$ENV{SNAPPY_STATIC_LIB}")
set(ZLIB_STATIC_LIB "$ENV{ZLIB_STATIC_LIB}")
```

wesm:
I'm a little bit confused about your build environment. If you are statically linking with `libparquet.a`, aren't you doing that in a different codebase? The only static linking that was provided in the 1.1.0 release was to create a `libparquet.so` that has all its dependencies included.

I see two distinct use cases for users of parquet-cpp:

* Users who want to use shared libraries should use libarrow.so (which has the compression symbols statically linked, so can be used standalone) and libparquet.so
* Users who want to statically link Parquet will have to include libarrow.a and its transitive dependencies

I don't think including Arrow's transitive dependencies in parquet-cpp assists with either of these use cases.

majetideepak:
I have no problem with linking libparquet.a in my codebase (Vertica).
I just can't develop/build Parquet tests/examples/benchmarks in static linking mode.

wesm:
I see. I guess my question is why that use case is important, or at least important enough to justify the extra complexity in Parquet's build system. Is the concern that the statically linked build might have test failures not present with shared libraries (which seems very esoteric)? 

If you have a patch already, then that seems OK to me, because we can delete all the CMake and thirdparty cruft anyway. We could later set up a "static toolchain" build in Travis CI so that it doesn't bitrot

majetideepak:
One advantage of using static linking tests is guarding against missing symbols. I think this is useful given the Parquet dependency on Arrow and its dependencies.
The patch is quite small. I agree we need a static toolchain build in Travis CI. I will make an attempt.

wesm:
Thanks @majetideepak and sorry for the additional trouble! I will submit another patch to do some further scrubbing of the build system when I can

PR #367, opened 2017-07-08
PARQUET-1048:  Apache Arrow static transitive dependencies

majetideepak:


wesm:
This looks OK, is this no longer WIP?

majetideepak:
Yes. Thank you!

PR #371, opened 2017-07-15
PARQUET-1035: Write Int96 from Arrow timestamp(ns)

xhochy:
Closes #356 

cc @c-nichols


c-nichols:
Awesome! Thank you @xhochy @wesm !

PR #375, opened 2017-07-25
PARQUET-1068: Modify .clang-format to use straight Google format with 90-character line width

wesm:
The main change is horizontal alignment. We should also do a clang-tidy pass sometime to do some further scrubbing


majetideepak:
+1 LGTM

PR #377, opened 2017-07-31
PARQUET-1074: Use long key ids in KEYS file

lekv:
Created like so: gpg --import < KEYS 2>&1 | grep key | awk '{print $3}'
| tr -d ':' | while read k; do gpg --list-sigs --keyid-format long $k;
gpg --export --armor $k; done > newkeys

### DIFF
--- a/KEYS
+++ b/KEYS
@@ -5,16 +5,14 @@ Users: pgp < KEYS
 Developers: 
   pgp -kxa <your name> and append it to this file.
   (pgpk -ll <your name> && pgpk -xa <your name>) >> this file.
-  (gpg --list-sigs <your name>
+  (gpg --list-sigs --keyid-format long <your name>
     && gpg --armor --export <your name>) >> this file.
 
-pub   2048R/7AE7E47B 2013-04-10 [expires: 2017-04-10]
-uid                  Julien Le Dem <julien@ledem.net>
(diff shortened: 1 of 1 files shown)


lekv:
Let me know what you think. If this is something we want to change I'll go ahead and create a JIRA.

xhochy:
I would be fine with this change but I'm a bit confused why the public keys changed. Is there any simple explanation for this?

lekv:
The shell one-liner seems to have split Alex's two keys into two separate blocks (because it exported by key, not by email address). I can try to fix that.

Your key seems to have gotten two more signatures since you added it. You can inspect them like so: diff <(gpg --list-packets < ~/src/parquet-cpp/apache-parquet-cpp-1.2.0/KEYS) <(gpg --list-packets ~/checkout/parquet-cpp/newkeys)

xhochy:
Build is broken because of https://issues.apache.org/jira/browse/PARQUET-1075

xhochy:
Also updated the KEYS file in SVN.

@lekv we might also need to update the KEYS file in parquet-format git

lekv:
@xhochy Thanks for taking care of SVN. I didn't know it even existed. Maybe it'd be best to copy the same file to all projects and leave a KEYS.README to explain how to make changes, e.g. change in one project, copy to SVN, copy to all other projects) to keep them in sync?

PR #378, opened 2017-08-02
PARQUET-1037: allow arbitrary size row-groups

TobyShaw:
The main change made is that you don't have to specify the size of a row-group upfront when writing it. This is signalled through a "row_count_determined" flag which is threaded through the relevant classes.

The AppendRowGroup(int64_t num_rows) method should have identical behaviour as before.
The AppendRowGroup() method should fix its row group after completing a single column, and enforce all
future columns match that. Empty row groups are still not allowed.

(You'll definitely want to squash these commits. Apologies for the poor git usage.)

### DIFF
--- a/src/parquet/column_writer.h
+++ b/src/parquet/column_writer.h
@@ -73,11 +73,13 @@ static constexpr int WRITE_BATCH_SIZE = 1000;
 class PARQUET_EXPORT ColumnWriter {
  public:
   ColumnWriter(ColumnChunkMetaDataBuilder*, std::unique_ptr<PageWriter>,
+               bool row_count_determined,
                int64_t expected_rows, bool has_dictionary, Encoding::type encoding,
                const WriterProperties* properties);
 
   static std::shared_ptr<ColumnWriter> Make(ColumnChunkMetaDataBuilder*,
                                             std::unique_ptr<PageWriter>,
+                                            bool row_count_determined,
                                             int64_t expected_rows,
                                             const WriterProperties* properties);
 
@@ -91,6 +93,8 @@ class PARQUET_EXPORT ColumnWriter {
    * @return Total size of the column in bytes
    */
   int64_t Close();
+
+  int64_t RowsWritten();
 
  protected:
   virtual std::shared_ptr<Buffer> GetValuesBuffer() = 0;
@@ -137,6 +141,9 @@ class PARQUET_EXPORT ColumnWriter {
   const ColumnDescriptor* descr_;
 
   std::unique_ptr<PageWriter> pager_;
+
+  // Whether the number of rows is determined.
+  bool row_count_determined_;
 
   // The number of rows that should be written in this column chunk.
   int64_t expected_rows_;
@@ -195,8 +202,9 @@ class PARQUET_EXPORT TypedColumnWriter : public ColumnWriter {
   typedef typename DType::c_type T;
 
   TypedColumnWriter(ColumnChunkMetaDataBuilder* metadata,
-                    std::unique_ptr<PageWriter> pager, int64_t expected_rows,
-                    Encoding::type encoding, const WriterProperties* properties);
+                    std::unique_ptr<PageWriter> pager, bool row_count_determined,
+                    int64_t expected_rows, Encoding::type encoding,
+                    const WriterProperties* properties);
 
   // Write a batch of repetition levels, definition levels, and values to the
   // column.
--- a/src/parquet/column_writer.h
+++ b/src/parquet/column_writer.h
@@ -73,14 +73,11 @@ static constexpr int WRITE_BATCH_SIZE = 1000;
 class PARQUET_EXPORT ColumnWriter {
  public:
   ColumnWriter(ColumnChunkMetaDataBuilder*, std::unique_ptr<PageWriter>,
-               bool row_count_determined,
-               int64_t expected_rows, bool has_dictionary, Encoding::type encoding,
+               bool has_dictionary, Encoding::type encoding,
                const WriterProperties* properties);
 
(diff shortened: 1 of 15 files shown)


src/parquet/file/metadata.cc
```diff
@@ -731,8 +731,14 @@ class RowGroupMetaDataBuilder::RowGroupMetaDataBuilderImpl {
     row_group_->__set_total_byte_size(total_byte_size);
   }
 
+  void SetNumRows(int num_rows) {
+    row_group_->num_rows = num_rows;
```
majetideepak: `row_group_->__set_num_rows(num_rows);`

src/parquet/file/writer-internal.cc
```diff
@@ -174,9 +174,29 @@ int64_t SerializedPageWriter::WriteDictionaryPage(const DictionaryPage& page) {
...
+}
 
 ColumnWriter* RowGroupSerializer::NextColumn() {
+  if (current_column_writer_) {
+   if (!row_count_determined_) {
+      num_rows_ = current_column_writer_->RowsWritten();
+     metadata_->SetNumRows(num_rows_);
+      row_count_determined_ = true;
+   } else {
+     if (current_column_writer_->RowsWritten() != num_rows()) {
```
majetideepak: Redundant here since this is already checked in `column_writer.cc` `ColumnWriter::Close()`

src/parquet/file/writer-internal.cc
```diff
@@ -174,9 +174,29 @@ int64_t SerializedPageWriter::WriteDictionaryPage(const DictionaryPage& page) {
...
+  } else {
+    return 0;
+  }
+}
 
 ColumnWriter* RowGroupSerializer::NextColumn() {
+  if (current_column_writer_) {
+   if (!row_count_determined_) {
+      num_rows_ = current_column_writer_->RowsWritten();
+     metadata_->SetNumRows(num_rows_);
```
majetideepak: You should check if the same number of values are written across all columns in the arbitrary size case. You could make a short cut by just comparing with the first column written.

src/parquet/column_writer.h
```diff
@@ -161,6 +165,10 @@ class PARQUET_EXPORT ColumnWriter {
   // number may be lower than num_buffered_values_.
   int64_t num_buffered_encoded_values_;
 
+  // Whether the number of rows is determined.
+  bool row_count_determined_;
+  int expected_rows;
```
TobyShaw: Good point, I'll modify these to match convention.

src/parquet/file/writer-internal.cc
TobyShaw: Well spotted, I'll remove this

src/parquet/file/writer-internal.cc
TobyShaw: I believe I am checking that? I'm also not sure what this shortcut means.

majetideepak: You are correct. I overlooked the `row_count_determined_ = true;` below. Thanks!

majetideepak:
This looks good to me. Travis CI seems to be failing due to formatting. You can use `make format` to fix this issue. Don't worry about the commits. The merge to master has been set to automatically squash all the commits. Thanks!

wesm:
There's a legitimate MSVC failure (compiler warning) that needs to be fixed

wesm:
@TobyShaw can you change the PR title to start with "PARQUET-1037:"? Thanks

TobyShaw:
Will fix it up sometime this week. Cheers for the feedback guys 👍

mrajcok:
I have a streaming application where I need to close files every "x" records or every 5-min -- whichever comes first.  This feature would make it much easier to support the "every 5-min" requirement, since I don't know how many rows might come in during a 5-min period.

TobyShaw:
Everything should be fixed up now.

wesm:
I'll give this a review when I can so we can get this into 1.3.0

src/parquet/file/file-serialize-test.cc
```diff
@@ -41,13 +41,17 @@ class TestSerialize : public PrimitiveTypedTest<TestType> {
...
     this->SetUpSchema(Repetition::OPTIONAL, num_columns_);
   }
 
  protected:
   int num_columns_;
+  int num_rowgroups_;
+  int rows_per_rowgroup_;
 
-  void FileSerializeTest(Compression::type codec_type) {
+  void FileSerializeTest_(Compression::type codec_type, bool row_count_known) {
```
wesm: reason for this rename?

majetideepak:
I agree with @wesm that `expected_rows` argument is redundant with this PR change. We are not doing anything useful (w.r.t resource allocations) using this argument anyway.

wesm:
I think we should be able to preserve the existing API for backwards compatibility

xhochy:
+1 with removing `expected_rows` in a backwards-compatible manner.

Once this is merged, I would push out an RC.

wesm:
@TobyShaw would you mind adding me as a collaborator on your fork? I can also attach a patch for you to cherry pick with these changes, either way is fine with me

src/parquet/file/file-serialize-test.cc
TobyShaw: So FileSerializeTest is implemented further down, in terms of this helper function, keeping the same signature.

wesm:
OK, I think I have got this sorted out. Please review again to make sure I didn't change an API that I shouldn't have (I think I preserved all public APIs but changed some internal APIs). I checked that the Python-Parquet side in Arrow 0.7.0 / master has no regressions

wesm:
thanks @TobyShaw!

PR #379, opened 2017-08-04
PARQUET-1072: Build with ARROW_NO_DEPRECATED_API in Travis CI

wesm:
Also fixes a number of Clang warnings and uses of deprecated API calls as of Arrow 0.6.0


wesm:
Looks like I accidentally force pushed the wrong branch. I will restore the original patch for PARQUET-1072 when I can

wesm:
@majetideepak does this look OK?

.travis.yml
```diff
@@ -40,27 +40,41 @@ matrix:
...
+    before_script:
+    - export PARQUET_CXXFLAGS="-Werror -DARROW_NO_DEPRECATED_API"
+    - source $TRAVIS_BUILD_DIR/ci/before_script_travis.sh
     before_install:
     - mkdir $TRAVIS_BUILD_DIR/parquet-build
     - pushd $TRAVIS_BUILD_DIR/parquet-build
+  - compiler: gcc
+    os: linux
+    env: PARQUET_BUILD_GROUP=toolchain
+    before_script:
```
majetideepak: Can you add `export PARQUET_CXXFLAGS="-Werror -DARROW_NO_DEPRECATED_API"` to static linking as well?

majetideepak:
+1 LGTM apart from a minor suggestion.

.travis.yml
majetideepak: I see that -Werror is already present. Not sure if adding `-DARROW_NO_DEPRECATED_API` makes a difference.

wesm: Done

PR #380, opened 2017-08-06
PARQUET-1078: Add option to coerce Arrow timestamps to a particular unit

wesm:
See ARROW-622 and ARROW-1076. We have been coercing timestamp units on ingest to Arrow format, rather than letting the data be whatever it is and converting only when writing to Parquet. This adds an option to coerce via the `ArrowWriterProperties`, and we also will return error Status if casting to a lower-resolution unit would lose data. We can later add an option to allow unsafe casts should that be desired.


wesm:
@cpcloud can you take a look at this? I'm going to write patches for Arrow a bit later to use this, and deprecate the `timestamps_to_ms` argument

cpcloud:
I will review this tomorrow morning.

src/parquet/arrow/arrow-reader-writer-test.cc
```diff
@@ -993,6 +995,145 @@ TEST(TestArrowReadWrite, DateTimeTypes) {
...
+                                    1489271000000001LL, 1489272000000001LL,
+                                    1489272000000001LL, 1489273000000001LL};
+
+  std::shared_ptr<Array> a_s, a_ms, a_us, a_ns;
+  ArrayFromVector<::arrow::TimestampType, int64_t>(t_s, is_valid, s_values, &a_s);
+  ArrayFromVector<::arrow::TimestampType, int64_t>(t_ms, is_valid, ms_values, &a_ms);
+  ArrayFromVector<::arrow::TimestampType, int64_t>(t_us, is_valid, us_values, &a_us);
+  ArrayFromVector<::arrow::TimestampType, int64_t>(t_ns, is_valid, ns_values, &a_ns);
+
+  auto s1 = std::shared_ptr<::arrow::Schema>(new ::arrow::Schema({field("f_s", t_s)}));
```
cpcloud: Any reason not to use `std::make_shared` here?

src/parquet/arrow/writer.cc
```diff
@@ -533,65 +545,132 @@ Status FileWriter::Impl::WriteNullableBatch<Int96Type, ::arrow::TimestampType>(
...
+                                               const int16_t* rep_levels) {
   // Note that we can only use data_buffer_ here as we write timestamps with the fast
   // path.
   RETURN_NOT_OK(data_buffer_.Resize(array->length() * sizeof(int64_t)));
   int64_t* data_buffer_ptr = reinterpret_cast<int64_t*>(data_buffer_.mutable_data());
 
-  auto data = static_cast<const NumericArray<::arrow::TimestampType>*>(array.get());
-  auto data_ptr = reinterpret_cast<const int64_t*>(data->values()->data());
+  const auto& data = static_cast<const ::arrow::TimestampArray&>(*array);
+  auto data_ptr = reinterpret_cast<const int64_t*>(data.values()->data())
```
cpcloud: Should this use `data.raw_values()` instead of `data.values()->data()`?

src/parquet/arrow/writer.cc
```diff
@@ -533,65 +545,132 @@ Status FileWriter::Impl::WriteNullableBatch<Int96Type, ::arrow::TimestampType>(
...
+
+  if (type.unit() == TimeUnit::NANO) {
+    if (target_unit == TimeUnit::MICRO) {
+      RETURN_NOT_OK(divide_by(1000));
+    } else {
+      DCHECK_EQ(TimeUnit::MILLI, target_unit);
+      RETURN_NOT_OK(divide_by(1000000));
+    }
+  } else if (type.unit() == TimeUnit::SECOND) {
+    RETURN_NOT_OK(multiply_by(target_unit == TimeUnit::MICRO ? 1000000 : 1000));
```
cpcloud: Should these values be global constants?

src/parquet/arrow/arrow-reader-writer-test.cc
wesm: I think `make_shared` is weird about initializer lists (you would need to declare the ctor for `std::vector<std::shared_ptr<::arrow::Field>>`

src/parquet/arrow/writer.cc
wesm: We can do that. I also opened ARROW-1335 because `TimestampArray::raw_values` includes the slice offset while `PrimitiveArray::raw_values` does not (both are used in this file, I think)

src/parquet/arrow/writer.cc
wesm: What would you call them?

cpcloud: maybe:

* `kMicrosPerSecond`
* `kMillisPerSecond`

src/parquet/arrow/arrow-reader-writer-test.cc
wesm: I created ARROW-1336; I think this could be solved with a public factory function

PR #382, opened 2017-08-08
PARQUET-1075: Fix broken Coverage upload

majetideepak:
The same problem is seen in another project below. Used the same fix.
https://github.com/micropython/micropython/issues/3246


majetideepak:
This is currently listed as an issue
https://github.com/eddyxu/cpp-coveralls/issues/118

PR #383, opened 2017-08-09
PARQUET-1002: Compute statistics based on Sort Order

majetideepak:
@lomereiter,  You might also want to take a look since you previously implemented the Statistics API.


src/parquet/statistics.cc
```diff
@@ -69,6 +69,7 @@ TypedRowGroupStatistics<DType>::TypedRowGroupStatistics(
   IncrementNullCount(null_count);
   IncrementDistinctCount(distinct_count);
 
+  comparator_ = std::static_pointer_cast<CompareDefault<DType> >(Compare::Make(schema));
```
lomereiter: Should this line be just moved into `SetDescr` then?

src/parquet/util/comparison.cc
```diff
@@ -0,0 +1,80 @@
...
+
+#include "parquet/exception.h"
+#include "parquet/schema.h"
+#include "parquet/types.h"
+#include "parquet/util/comparison.h"
+
+namespace parquet {
+
+std::shared_ptr<Compare> Compare::Make(const ColumnDescriptor* descr) {
+  if (SortOrder::SIGNED == GetSortOrder(descr->logical_type(), descr->physical_type())) {
```
lomereiter: I counted 5 occurrences of this call, it probably makes sense to add a convenience method to `ColumnDescriptor`

majetideepak: Yes! makes sense.

majetideepak:
@wesm, @xhochy can you take a look at this? thanks!

wesm:
I'm sorry about the delay, I will try to review this today or worst case tomorrow

wesm:
Going to review this today

src/parquet/types.h
```diff
@@ -283,6 +296,11 @@ PARQUET_EXPORT std::string FormatStatValue(Type::type parquet_type, const char*
 
 PARQUET_EXPORT int GetTypeByteSize(Type::type t);
 
+SortOrder::type PARQUET_EXPORT DefaultSortOrder(Type::type primitive);
+
+SortOrder::type PARQUET_EXPORT GetSortOrder(LogicalType::type converted,
```
wesm: Can you put the `PARQUET_EXPORT` before the return type for consistency (we should go through and make them all consistent)?

src/parquet/util/comparison.h
```diff
@@ -20,40 +20,145 @@
...
 #include "parquet/types.h"
 
 namespace parquet {
 
-template <typename T>
-struct Compare {
-  explicit Compare(const ColumnDescriptor* descr) : type_length_(descr->type_length()) {}
-
-  inline bool operator()(const T& a, const T& b) { return a < b; }
+class PARQUET_EXPORT Compare {
```
wesm: nit: can we rename this `Comparator`?

src/parquet/util/comparison.h
```diff
@@ -20,40 +20,145 @@
...
- private:
-  int32_t type_length_;
+// The default comparison is SIGNED
+template <typename DType>
+class PARQUET_EXPORT CompareDefault : public Compare {
+ public:
+  typedef typename DType::c_type T;
+  CompareDefault() {}
+  virtual ~CompareDefault() {}
+  virtual bool operator()(const T& a, const T& b) { return a < b; }
```
wesm: override (instead of virtual)

src/parquet/util/comparison.h
```diff
@@ -20,40 +20,145 @@
...
 
 template <>
-inline bool Compare<Int96>::operator()(const Int96& a, const Int96& b) {
-  return std::lexicographical_compare(a.value, a.value + 3, b.value, b.value + 3);
-}
+class PARQUET_EXPORT CompareDefault<Int96Type> : public Compare {
+ public:
+  CompareDefault() {}
+  virtual ~CompareDefault() {}
+  virtual bool operator()(const Int96& a, const Int96& b) {
```
wesm: override

src/parquet/util/comparison.h
```diff
@@ -20,40 +20,145 @@
...
-inline bool Compare<ByteArray>::operator()(const ByteArray& a, const ByteArray& b) {
-  auto aptr = reinterpret_cast<const int8_t*>(a.ptr);
-  auto bptr = reinterpret_cast<const int8_t*>(b.ptr);
-  return std::lexicographical_compare(aptr, aptr + a.len, bptr, bptr + b.len);
-}
+class PARQUET_EXPORT CompareDefault<ByteArrayType> : public Compare {
+ public:
+  CompareDefault() {}
+  virtual ~CompareDefault() {}
+  virtual bool operator()(const ByteArray& a, const ByteArray& b) {
```
wesm: override

src/parquet/util/comparison.h
```diff
@@ -20,40 +20,145 @@
...
-  auto aptr = reinterpret_cast<const int8_t*>(a.ptr);
-  auto bptr = reinterpret_cast<const int8_t*>(b.ptr);
-  return std::lexicographical_compare(aptr, aptr + type_length_, bptr,
-                                      bptr + type_length_);
-}
+class PARQUET_EXPORT CompareDefault<FLBAType> : public Compare {
+ public:
+  explicit CompareDefault(int length) : type_length_(length) {}
+  virtual ~CompareDefault() {}
+  virtual bool operator()(const FLBA& a, const FLBA& b) {
```
wesm: override

src/parquet/file/metadata.cc
```diff
@@ -35,62 +35,20 @@ const ApplicationVersion ApplicationVersion::PARQUET_251_FIXED_VERSION =
...
-    case LogicalType::MAP:
-    case LogicalType::MAP_KEY_VALUE:
-    case LogicalType::INTERVAL:
-    case LogicalType::NONE:  // required instead of default
-      return SortOrder::UNKNOWN;
-  }
-  return SortOrder::UNKNOWN;
-}
+const ApplicationVersion ApplicationVersion::PARQUET_CPP_FIXED_STATS_VERSION =
+    ApplicationVersion("parquet-cpp version 1.2.1");
```
majetideepak: Releasing 1.2.1 sounds like a good idea. But this value should work even if we make a 1.3.0 release.

src/parquet/file/metadata.cc
```diff
@@ -577,16 +539,24 @@ class ColumnChunkMetaDataBuilder::ColumnChunkMetaDataBuilderImpl {
...
+    stats.min_value = val.min();
+    stats.__isset.min_value = val.has_min;
+    stats.__isset.max_value = val.has_max;
     stats.__isset.null_count = val.has_null_count;
     stats.__isset.distinct_count = val.has_distinct_count;
+    // If the order is SIGNED, then the old min/max values must be set too.
+    // This for backward compatibility
+    if (is_signed) {
+      stats.max = val.max();
+      stats.min = val.min();
```
majetideepak: I check and it didn't.

src/parquet/parquet_reader_writer-test.cc
```diff
@@ -0,0 +1,330 @@
+// Licensed to the Apache Software Foundation (ASF) under one
```
majetideepak: My motivation is to add end to end writer-reader tests for Parquet which we currently don't have.

src/parquet/parquet_reader_writer-test.cc
```diff
@@ -0,0 +1,330 @@
...
+  stats_[0]
+      .set_min(std::string(reinterpret_cast<const char*>(&values_[0]), sizeof(T)))
+      .set_max(std::string(reinterpret_cast<const char*>(&values_[9]), sizeof(T)));
+}
+
+// TYPE::DOUBLE
+template <>
+void TestStatistics<DoubleType>::AddNodes(std::string name) {
+  // Double physical type has only Signed Statistics
+  fields_.push_back(schema::PrimitiveNode::Make(name, Repetition::REQUIRED, Type::DOUBLE,
```
majetideepak: I will take a look.

src/parquet/parquet_version.h
```diff
@@ -19,6 +19,6 @@
 #define PARQUET_VERSION_H
 
 // define the parquet created by version
-#define CREATED_BY_VERSION "parquet-cpp version 1.1.1-SNAPSHOT"
+#define CREATED_BY_VERSION "parquet-cpp version 1.2.1-SNAPSHOT"
 
-#endif  // PARQUET_VERSION_H
+#endif // PARQUET_VERSION_H
```
majetideepak: Removing this file causes one of the Travis CI tests to fail. I am not sure why. I will look into this as well.

src/parquet/types.h
majetideepak: Will do.

src/parquet/util/comparison.h
majetideepak: Sure!

src/parquet/util/comparison.h
majetideepak: will do here and below.

src/parquet/file/metadata.cc
majetideepak: I just noticed https://github.com/apache/parquet-format/pull/55/files adds some more changes to the min-max stats format. I think we should wait until that gets merged.

majetideepak:
Thanks for the feedback @wesm! As far as I remember, the sync discussions were geared more towards UTF-8  (and other string encodings) and not much on Int96. Int96 is deprecated and is currently used only for Impala Timestamp values which require a custom sort order.

I have also implemented this patch with the idea of users be able to implement custom `Comparator` classes and plugin into the writer properties. Any custom order should be easy to implement now by simply extending the `CompareDefault**Type**` class

src/parquet/column_writer.cc
```diff
@@ -267,7 +267,9 @@ int64_t ColumnWriter::Close() {
     FlushBufferedDataPages();
 
     EncodedStatistics chunk_statistics = GetChunkStatistics();
-    if (chunk_statistics.is_set()) metadata_->SetStatistics(chunk_statistics);
+    if (chunk_statistics.is_set())
```
xhochy: Nit: Please use braces.

src/parquet/file/metadata.h
```diff
@@ -92,7 +82,7 @@ class ApplicationVersion {
   bool VersionEq(const ApplicationVersion& other_version) const;
 
   // Checks if the Version has the correct statistics for a given column
-  bool HasCorrectStatistics(Type::type primitive) const;
+  bool HasCorrectStatistics(Type::type primitive, SortOrder::type sort_order) const;
```
xhochy: As we change the interface here, we should make the new version 1.3.0. 

@majetideepak alternatively can we provide a default argument here?

wesm:
@majetideepak if you can incorporate feedback and make this not WIP, we should be able to cut the 1.3.0 RC after this is in

majetideepak:
@wesm Sorry for the delay! I somehow missed your comment. I will incorporate this feedback tomorrow morning.

src/parquet/parquet_version.h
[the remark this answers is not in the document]
majetideepak: Filed VER-1088

src/parquet/util/comparison.h
majetideepak: `CompareDefault` is the base class and will require virtual. I used override inside the derived classes.

wesm:
A follow up question on the comparator thing. Any way to do this as fully compile-time with no virtual functions (and so no exported symbols in the DLL)?

majetideepak:
I couldn't come up with an elegant design to make it fully-compile time.
The comparator depends upon the `SortOrder` and the `c_type` of the parquet physical types.
The `c_type` can be determined at compile time for a given TypedRowGroupStatistics. I don't know if we can determine the `SortOrder` at compile time.
Another motivation for virtual functions is to allow users to specify custom comparators for unknown sort orders.

src/parquet/parquet_reader_writer-test.cc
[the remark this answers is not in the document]
xhochy: I would also like to see this in `statistics-test.cc`. We can move it once we have full end-to-end tests.

majetideepak:
Fixed issue https://github.com/apache/parquet-cpp/pull/383/files#r134647005
We can go ahead without the merge of apache/parquet-format#55. That PR only impacts how the SortOrder is inferred. This PR will not require any changes.

xhochy:
Travis is failing with 

```
/Users/travis/build/apache/parquet-cpp/src/parquet/statistics-test.cc:634:13: error: variable-sized object may not be initialized
  char vals[NUM_VALUES][FLBA_LENGTH] = {"b12345", "a12345", "c12345", "d12345", "e12345",
```

Once that is fixed, feel free to merge.

majetideepak:
@xhochy Fixed the error. I don't think I have merge access yet!

wesm:
Merging. It looks like you haven't been added to the LDAP group, @julienledem can you take a look?

PR #384, opened 2017-08-29
PARQUET-1082: Change MutableBuffer to Buffer to account for deprecation in ARROW-1402

wesm:
Please do not merge this patch until after parquet-cpp 1.3.0 is released. Because these APIs have been deprecated, but not removed, the current trunk will build with both Arrow 0.6.0 and 0.7.0

### DIFF
--- a/src/parquet/arrow/reader.cc
+++ b/src/parquet/arrow/reader.cc
@@ -332,7 +332,7 @@ class PARQUET_NO_EXPORT StructImpl : public ColumnReader::Impl {
   std::shared_ptr<Field> field_;
   PoolBuffer def_levels_buffer_;
 
-  Status DefLevelsToNullArray(std::shared_ptr<MutableBuffer>* null_bitmap,
+  Status DefLevelsToNullArray(std::shared_ptr<Buffer>* null_bitmap,
                               int64_t* null_count);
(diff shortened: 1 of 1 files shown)


wesm:
@xhochy @majetideepak given proximity of Arrow 0.7.0 I suggest that we upgrade to Arrow master and merge this patch, then we can release parquet-cpp 1.3.0 that is based on 0.7.0. Sound good?

xhochy:
@wesm I'm happy to merge this once we have an Arrow 0.7.0 release. I would then directly start the Parquet C++ release vote.

majetideepak:
Sounds good. Are we including PARQUET-1002 in the 1.3.0 release? https://github.com/apache/parquet-cpp/pull/383

majetideepak:
I also noticed that the Arrow master now requires additional boost system library dependency. Will Parquet now transitively require that as well?

wesm:
This has been true for a long time, at least since

https://github.com/apache/arrow/commit/d1a9aff2937efe54fe3a5c80f7fbe19851cb71f3#diff-6725b893dfc969abac4f4ee39a3a317fR864

It's only used for exception handling from boost/filesystem.

PR #386, opened 2017-09-02
PARQUET-1085: [C++] Use namespaced macros from arrow/util/macros.h, work around UNUSED rename

wesm:
UNUSED being renamed to ARROW_UNUSED, so I work around this. I also account for LIKELY/UNLIKELY disappearing because ARROW_PREDICT_TRUE/FALSE were already in Arrow 0.6.0. We can always give these distinct names when used inside parquet-cpp but seems OK for now

### DIFF
--- a/src/parquet/exception.h
+++ b/src/parquet/exception.h
@@ -26,6 +26,11 @@
 
 #include "parquet/util/visibility.h"
 
+// PARQUET-1085
+#if !defined(ARROW_UNUSED)
+#define ARROW_UNUSED(x) UNUSED(x)
+#endif
+
 #define PARQUET_CATCH_NOT_OK(s)                    \
   try {                                            \
     (s);                                           \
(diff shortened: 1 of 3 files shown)


wesm:
cc @majetideepak @cpcloud @xhochy can someone take a quick look? This build should pass on pre-0.7.0 Arrow (and I've tested locally with 0.7.0). It is part of https://github.com/apache/arrow/pull/1030

wesm:
This build passed, but there was an idiosyncratic coveralls-related error on one of the Travis CI entries.

src/parquet/exception.h
```diff
@@ -26,6 +26,11 @@
 
 #include "parquet/util/visibility.h"
 
+// PARQUET-1085
+#if !defined(ARROW_UNUSED)
+#define ARROW_UNUSED(x) UNUSED(x)
```
xhochy: +1, backwards compability :)

PR #389, opened 2017-09-06
PARQUET-1090: Add max row group length option, fix int32 overflow

wesm:
Also fixed a race condition on googletest_ep.


majetideepak:
+1 LGTM

PR #394, opened 2017-09-11
PARQUET-1097: Account for internal API change in ARROW-1511

wesm:
See https://github.com/apache/arrow/pull/1087


wesm:
Also cherry-picking this into #397

PR #395, opened 2017-09-11
PARQUET-1098: Install util/comparison.h

wesm:
This was breaking third party builds. cc @majetideepak


wesm:
I created PARQUET-1099 to try to prevent this from happening in the future

majetideepak:
+1 Thanks for fixing this.

PR #396, opened 2017-09-12
WIP: PARQUET-1095/PARQUET-982: [C++] Read and write Arrow decimal values

cpcloud:


wesm:
I believe that if we use FIXED_LEN_BYTE_ARRAY, that we may need to pack the data into the smallest byte width that will accommodate the decimal precision. We should double check about whether Spark or other Java consumers are able to decode this data

src/parquet/arrow/reader.cc
```diff
@@ -1319,29 +1336,21 @@ Status PrimitiveImpl::NextBatch(int batch_size, std::shared_ptr<Array>* out) {
...
-      break;
-    }
+    case ::arrow::Type::FIXED_SIZE_BINARY:
+      return ReadFLBABatch<::arrow::FixedSizeBinaryType>(batch_size, type, out);
+    case ::arrow::Type::DECIMAL:
+      return ReadFLBABatch<::arrow::DecimalType>(batch_size, type, out);
     case ::arrow::Type::TIMESTAMP: {
-      ::arrow::TimestampType* timestamp_type =
-          static_cast<::arrow::TimestampType*>(field_->type().get());
+      const auto& timestamp_type = std::static_pointer_cast<::arrow::TimestampType>(type);
```
xhochy: As you already touch this line, change this to `static_cast<const ::arrow::TimestampType&>`. This should save us from some unnecessary reference counting.

src/parquet/arrow/writer.cc
```diff
@@ -679,7 +678,7 @@ Status FileWriter::Impl::TypedWriteBatch<BooleanType, ::arrow::BooleanType>(
     const int16_t* def_levels, const int16_t* rep_levels) {
   RETURN_NOT_OK(data_buffer_.Resize(array->length()));
   auto data = static_cast<const BooleanArray*>(array.get());
-  auto data_ptr = reinterpret_cast<const uint8_t*>(data->values()->data());
+  auto data_ptr = data->values()->data();
```
xhochy: Google styleguide nitpick: Only use `auto` where the type of the variable is already mentioned in the line.

wesm:
I have a somewhat invasive refactor of parquet/arrow/reader.cc ongoing for PARQUET-1100, I will try to get it to ship shape in the next couple days

cpcloud:
I'm going to wait for PARQUET-1100 before continuing on this patch.

PR #397, opened 2017-09-13
PARQUET-1104: Upgrade to Apache Arrow 0.7.0 RC0

wesm:
This will enable most of the outstanding PRs to be rebased and then merged

Close #384 
Close #394


wesm:
Cherry-picked PARQUET-1082 since it's a required fix

wesm:
@majetideepak I included a fix for the boost_system issue. I remember you mentioned this, but I'd been too busy to investigate. Can you definitely bug us very loudly when you run into a new transitive dependency? I would rather have fixed this before cutting the Arrow 0.7.0 RC but it's not a big enough problem to cancel the release over.

majetideepak:
@wesm  Will be very loud next time. Can we revert the  boost_system dependency in the next release?

wesm:
I created https://issues.apache.org/jira/browse/ARROW-1536 about fixing that, yes

majetideepak:
Filed PARQUET-1105 on the Parquet side to track it.

PR #402, opened 2017-09-22
PARQUET-1111: Fix help for verify-release-candidate

lekv:
The tool takes 2 parameters, version and rc-num. The help only printed a
single parameter.


lekv:
@xhochy - Can you have a look? Thanks :)

PR #403, opened 2017-09-25
PARQUET-1095: [C++] Read and write Arrow decimal values

cpcloud:
This depends on:
- [x] [ARROW-1607](https://github.com/apache/arrow/pull/1128)
- [x] [ARROW-1656](https://github.com/apache/arrow/pull/1184)
- [x] [ARROW-1588](https://github.com/apache/arrow/pull/1211)
- [x] Add tests for writing different sizes of values


src/parquet/arrow/schema.cc
```diff
@@ -565,12 +576,12 @@ Status FieldToNode(const std::shared_ptr<Field>& field,
       auto struct_type = std::static_pointer_cast<::arrow::StructType>(field->type());
       return StructToNode(struct_type, field->name(), field->nullable(), properties,
                           arrow_properties, out);
-    } break;
```
xhochy: Accidental deletion?

src/parquet/arrow/schema.cc
```diff
@@ -565,12 +576,12 @@ Status FieldToNode(const std::shared_ptr<Field>& field,
...
+    }
     case ArrowType::LIST: {
       auto list_type = std::static_pointer_cast<::arrow::ListType>(field->type());
       return ListToNode(list_type, field->name(), field->nullable(), properties,
                         arrow_properties, out);
-    } break;
```
xhochy: Accidental deletion?

src/parquet/arrow/schema.cc
cpcloud: No, the `break` isn't necessary because there's a `return` right?

xhochy: Ah, didn't see that in the quick review.

cpcloud:
I'm working on getting my windows VM setup with parquet-cpp, haven't forgotten about this.

src/parquet/arrow/reader.cc
```diff
@@ -874,6 +875,134 @@ struct TransferFunctor<
...
+      // Convert the fixed size binary array bytes into a Decimal128 compatible layout
+      BytesToIntegerPair(value, byte_width, high, low);
+    };
+
+    // convert each FixedSizeBinary value to valid decimal bytes
+    const int64_t null_count = fixed_size_binary_array.null_count();
+    if (null_count > 0) {
+      for (int64_t i = 0; i < length; ++i, out_ptr += type_length) {
+        if (!fixed_size_binary_array.IsNull(i)) {
+          raw_bytes_to_decimal_bytes(fixed_size_binary_array.GetValue(i), out_ptr);
```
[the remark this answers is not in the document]
cpcloud: The new buffer will have defined slots because I take a view on the bytes (`int64_t*` for high bits, `uint64_t*` for low bits) and assign values `*high`/`*low`. That happens in `BytesToIntegerPair`.

src/parquet/arrow/schema.cc
```diff
@@ -617,5 +629,73 @@ Status ToParquetSchema(const ::arrow::Schema* arrow_schema,
                          out);
 }
 
+int32_t DecimalSize(int32_t precision) {
```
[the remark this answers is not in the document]
cpcloud: Yep, will do.

src/parquet/arrow/schema.h
```diff
@@ -85,6 +85,8 @@ ::arrow::Status PARQUET_EXPORT ToParquetSchema(const ::arrow::Schema* arrow_sche
                                                const WriterProperties& properties,
                                                std::shared_ptr<SchemaDescriptor>* out);
 
+int32_t PARQUET_EXPORT DecimalSize(int32_t precision);
```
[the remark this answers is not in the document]
cpcloud: Only if dependents want to use it. We can not export it for now as it's only used internally right now.

src/parquet/arrow/writer.cc
```diff
@@ -765,24 +765,87 @@ Status FileWriter::Impl::TypedWriteBatch<FLBAType, ::arrow::FixedSizeBinaryType>
...
+
+  if (writer->descr()->schema_node()->is_required() || data.null_count() == 0) {
+    // no nulls, just dump the data
+    // todo(advancedxy): use a writeBatch to avoid this step
+    for (int64_t i = 0; i < length; i++) {
+      // bytes may be in little endian order, so swap
+      const uint8_t* raw_value = data.GetValue(i);
+      auto unsigned_64_bit = reinterpret_cast<const uint64_t*>(raw_value);
+      const uint64_t value[] = {::arrow::BitUtil::ToBigEndian(unsigned_64_bit[0]),
+                                ::arrow::BitUtil::ToBigEndian(unsigned_64_bit[1])};
```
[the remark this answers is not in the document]
cpcloud: Do you mean on the way in? These lines are performing the byte swapping on the way out.

src/parquet/arrow/schema.cc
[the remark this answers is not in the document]
cpcloud: Done

cmake_modules/ThirdpartyToolchain.cmake
```diff
@@ -366,7 +366,7 @@ if (NOT ARROW_FOUND)
     -DARROW_BUILD_TESTS=OFF)
 
   if ("$ENV{PARQUET_ARROW_VERSION}" STREQUAL "")
-    set(ARROW_VERSION "8309556c7d2b0e14df1422baa574cf2de8c1bd3b")
```
xhochy: Do we need more than 0.7.1 here?

If yes, I would like to make a parquet-cpp 1.3.1 release otherwise we can wait with this PR to be merged.

wesm: Yes, we should probably release 1.3.1 without this and then focus on getting Arrow 0.8.0 out by the end of the month. There is also https://issues.apache.org/jira/browse/PARQUET-1122 -- I am not sure if I want to block 1.3.1 over this but we will need to take care of that soon

cpcloud: This PR needs some very recent fixes in Arrow, so yes it needs more than 0.7.1. I'm fine waiting for arrow 0.8.0 then parquet-cpp 1.4.0

cpcloud:
This patch now includes explicit support for reading Decimals written in other systems as int32/int64 as per the parquet spec. The patch includes small test datasets written in spark.

wesm:
Now that ARROW-1588 is merged, we can get this fixed up and then plan to release 1.4.0 with decimal support after Arrow 0.8.0 final is out (so this would be latter half of November)?

cpcloud:
Sounds good to me. One last thing is to add tests for different byte width decimals (using random data generation).

xhochy:
For the moment, this looks good. Reping me, once this is ready to merge, than I can do a final pass.

wesm:
Aside: I think we should call `DecimalArray` instead `Decimal16Array` in Arrow to permit smaller 4/8-byte decimals in the future without breaking the parquet-cpp API. I'll open a JIRA

wesm:
see https://issues.apache.org/jira/browse/ARROW-1794

see related comments about this topic in the Kudu JIRA https://issues.apache.org/jira/browse/KUDU-721?focusedCommentId=16213209&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-16213209

cpcloud:
I'm working on ARROW-1794. Let's get that in before merging this patch.

src/parquet/arrow/arrow-reader-writer-test.cc
```diff
@@ -1889,6 +1936,61 @@ TEST(TestArrowReaderAdHoc, Int96BadMemoryAccess) {
...
+  ASSERT_EQ(expected_length, value_column->length());
+
+  auto raw_array = value_column->data();
+  ASSERT_EQ(1, raw_array->num_chunks());
+
+  auto chunk = raw_array->chunk(0);
+
+  std::shared_ptr<Array> expected_array;
+
+  ::arrow::DecimalBuilder builder(decimal_type, pool);
```
wesm: Do we also need to rename this builder?

src/parquet/arrow/reader.cc
```diff
@@ -868,12 +870,227 @@ struct TransferFunctor<
...
+  using ::arrow::BitUtil::ToLittleEndian;
+
+  for (int64_t i = 0; i < length; ++i, out_ptr += type_length) {
+    const ElementType value = values[i];
+
+    auto out_ptr_view = reinterpret_cast<uint64_t*>(out_ptr);
+    out_ptr_view[0] = ToLittleEndian(static_cast<uint64_t>(value));
+
+    // no need to byteswap here because we're either all ones or all zeros
+    out_ptr_view[1] = static_cast<uint64_t>(value < 0 ? -1 : 0);
```
wesm: The sign bit is in the same place in both cases (big vs little endian)? I'm clearly out of my depth on these details

src/parquet/arrow/reader.cc
```diff
@@ -932,6 +1149,22 @@ Status PrimitiveImpl::NextBatch(int64_t records_to_read, std::shared_ptr<Array>*
...
+      switch (descr_->physical_type()) {
+        case ::parquet::Type::INT32: {
+          TRANSFER_DATA(::arrow::DecimalType, Int32Type);
+        } break;
+        case ::parquet::Type::INT64: {
+          TRANSFER_DATA(::arrow::DecimalType, Int64Type);
+        } break;
+        case ::parquet::Type::FIXED_LEN_BYTE_ARRAY: {
+          TRANSFER_DATA(::arrow::DecimalType, FLBAType);
+        } break;
```
wesm: Do any systems that we know of use BYTE_ARRAY?

src/parquet/arrow/schema.h
```diff
@@ -85,6 +85,8 @@ ::arrow::Status PARQUET_EXPORT ToParquetSchema(const ::arrow::Schema* arrow_sche
                                                const WriterProperties& properties,
                                                std::shared_ptr<SchemaDescriptor>* out);
 
+int32_t DecimalSize(int32_t precision);
```
wesm: I presume this does not need to be exported. It could go in a private header, also

src/parquet/arrow/writer.cc
```diff
@@ -765,29 +763,82 @@ Status FileWriter::Impl::TypedWriteBatch<FLBAType, ::arrow::FixedSizeBinaryType>
...
-        writer->WriteBatch(num_levels, def_levels, rep_levels, buffer_ptr));
   }
+  PARQUET_CATCH_NOT_OK(
+      writer->WriteBatch(num_levels, def_levels, rep_levels, buffer_ptr));
+  PARQUET_CATCH_NOT_OK(writer->Close());
+  return Status::OK();
+}
+
+template <>
+Status FileWriter::Impl::TypedWriteBatch<FLBAType, ::arrow::DecimalType>(
```
wesm: Decimal128Type

src/parquet/arrow/writer.cc
```diff
@@ -765,29 +763,82 @@ Status FileWriter::Impl::TypedWriteBatch<FLBAType, ::arrow::FixedSizeBinaryType>
...
+  const auto& decimal_type = static_cast<const ::arrow::DecimalType&>(*data.type());
+  const int32_t offset =
+      decimal_type.byte_width() - DecimalSize(decimal_type.precision());
+
+  const bool does_not_have_nulls =
+      writer->descr()->schema_node()->is_required() || data.null_count() == 0;
+
+  // TODO(phillipc): Look into whether our compilers will perform loop unswitching so we
+  // don't have to keep writing two loops to handle the case where we know there are no
+  // nulls
```
wesm: It seems like MSVC didn't used to do loop unswitching but now maybe it does from MSVC 2013 onward (I guess this used to only be available in the Pro/Enterprise version of visual C++). Did a bit of googling but didn't get a definitive answer

wesm: FWIW it looks like LLVM has a heuristic about whether it will unswitch: http://llvm.org/doxygen/LoopUnswitch_8cpp_source.html. I am not sure I like that. my inclination has generally been to unswitch by hand

src/parquet/arrow/writer.cc
cpcloud: I didn't actually rename `DecimalBuilder` or `DecimalType`, just `DecimalArray` because I assumed we'd be using `DecimalBuilder` and `DecimalType` independent of the underlying storage. I think it makes sense to delimit these as well. I'll open a JIRA.

src/parquet/arrow/schema.h
cpcloud: No, this shouldn't be available to third parties.

src/parquet/arrow/reader.cc
cpcloud: Thanks for pointing this out. There are actually two bugs here, revisiting this after reading your comment and rereading my code.

1. I should be calling `FromLittleEndian` just before this line, not `ToLittleEndian` because parquet-mr writes all primitive values in little endian order. The current code works on little endian architectures but not on big endian.
2. `value` needs to be sign/zero extended if it's of type `int32_t`. That's done by simply upcasting to `int64_t`.

This particular line is performing sign/zero extension to the other 8 bytes of the 16 byte decimal value, which means it's either 64 ones if `value` is negative, or 64 zeros if it's zero or positive.

Let me know if this explanation doesn't make sense.

src/parquet/arrow/reader.cc
cpcloud: I'm not aware of any. parquet-mr *does* support this, but both hive and spark always use `FIXED_LEN_BYTE_ARRAY`. (spark uses int32 or int64, optionally, in later versions). I can implement this now, or in a follow up patch.

src/parquet/arrow/writer.cc
cpcloud: Yeah, the heuristic is based on the number of basic blocks (points where control flow can take a different path depending on the value of a conditional) and the cost of instructions that might be generated (this seems difficult hard to predict without knowing the details of the cost of particular instructions) because loop unswitching doubles the number of loops every time there's an unswitching opportunity.

Since most or all of our unswitch opportunities are based on one condition and therefore double only once, I would be extremely surprised if we ever came close to the threshold.

In any event it doesn't look like this optimization is reliably implemented in all of the compilers we want to support to start writing switched loops, I just wanted to note this here for posterity.

src/parquet/arrow/arrow-reader-writer-test.cc
cpcloud: Done in https://github.com/apache/arrow/pull/1321

src/parquet/arrow/arrow-reader-writer-test.cc
```diff
@@ -856,25 +885,43 @@ TEST_F(TestUInt32ParquetIO, Parquet_1_0_Compability) {
...
+  for (int i = 0; i < expected.length(); ++i) {
+    const bool value_is_valid = values->IsValid(i);
+    const bool expected_value_is_valid = expected.IsValid(i);
+
+    ASSERT_EQ(value_is_valid, expected_value_is_valid);
+
+    if (value_is_valid) {
+      uint32_t value = values->Value(i);
+      int64_t expected_value = expected.Value(i);
+      ASSERT_EQ(value, expected_value);
```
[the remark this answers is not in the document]
cpcloud: It looks like this is standards-compliant behavior: http://en.cppreference.com/w/cpp/language/operator_arithmetic#Conversions. Specifically, this language:

> Otherwise, if the signed operand's type can represent all values of the unsigned operand, the unsigned operand is converted to the signed operand's type

Since `UINT32_MAX <= INT64_MAX` then this is well defined behavior.

That said, I'll add an explicit cast. I think it's more readable that way. The operands to `ASSERT_EQ` should also be reversed.

src/parquet/arrow/reader.cc
wesm: makes sense at a glance

cpcloud: I will add an example in the code to make this very concrete. It's not necessarily obvious unless one is familiar with the details of sign extension (which is a consequence of using a two's complement representation).

wesm:
@cpcloud is this merge-ready once the build passes? I just opened https://issues.apache.org/jira/browse/ARROW-1836 about fixing the warning I just triaged here

cpcloud:
@wesm Yep, was just about to ping. This is good to go.

cpcloud:
@wesm I have one more change, which removes the loop from `BytesToInteger` and uses integer arithmetic operations only, even for byte sizes that are not powers of two.

wesm:
Cool. +1, will merge when build passes

cpcloud:
Thanks for fixing the warning!

wesm:
Nice work! Now we can add some Python tests in Arrow and call it a wrap

PR #413, opened 2017-10-16
PARQUET-1140: Fail on RAT errors in CI

xhochy:


xhochy:
@wesm Can you give this a quick review, I would then push out the next RC afterwards.

PR #419, opened 2017-11-21
PARQUET-970: Add Lz4 and Zstd compression codecs

advancedxy:
which are already added in [apache/parquet-format#70](https://github.com/apache/parquet-format/pull/70)


advancedxy:
Looks like the travis-ci's error is due to that libzstd.a is not involved when linking libarrow.a?

xhochy:
The update of `parquet.thrift` looks fine. 

For libztd.a, you will need at least add `zstd` to https://github.com/apache/parquet-cpp/blob/master/ci/travis_script_toolchain.sh#L43 so that it is installed from conda-forge.

There is no need for the benchmark code, we can keep it or leave it. I have no strong opinion on that.

advancedxy:
Yeah, I know the update to `parquet.thrift` should be sufficient for this change. But do we need to sync with upstream: [parquet-format/parquet.thrift](https://github.com/apache/parquet-format/blob/master/src/main/thrift/parquet.thrift)? (I checked that file, and seems there are other changes made into parquet.thrift)



>For libztd.a, you will need at least add zstd to https://github.com/apache/parquet-cpp/blob/master/ci/travis_script_toolchain.sh#L43 so that it is installed from conda-forge

I don't know arrow's building procedure very well, but from my understanding, shouldn't zstd be part of dependency of arrow, and downloaded automatically?

xhochy:
@advancedxy We can leave the rest of the changes in `parquet.thrift` for another PR. This should make merging this one here simpler.

For Arrow's build: Arrow downloads and builds the zstd static library automatically if not found on the system but will not install it again then. In this case, you can only use Arrow's shared libraries. Thus if you want to use static Arrow builds, you need to take care of supplying a static zlib build in your system. In this case, this should be done by `conda` as it defines the build environment.

advancedxy:
@xhochy Thanks for your comments. I will update corresponding changes soon.

advancedxy:
@xhochy I have rebased my commits to drop todo for parquet.thrift and add zstd and lz4-c in `travis_script_toolchain.sh`

However one question left: why Snappy or Zlib are not specified in `travis_script_toolchain.sh`

xhochy:
@advancedxy Ah, we are using Arrow's static builds via `export SNAPPY_STATIC_LIB=$ARROW_EP/snappy_ep/src/snappy_ep-install/lib/libsnappy.a` This is a possible alternative to adding it to conda. Not the nicest solution but also possible.

advancedxy:
@xhochy 
>we are using Arrow's static builds via `export SNAPPY_STATIC_LIB=$ARROW_EP/snappy_ep/src/snappy_ep-install/lib/libsnappy.a`

But that is in `travis_script_static.sh` not in `travis_script_toolchain.sh`..

Looks like travis is failing now....

advancedxy:
@xhochy Looks like the change should be applied to `travis_script_static.sh`, use `conda` to add zstd and lz4-c or use `export ZSTD_STATIC_LIB` trick?

advancedxy:
Ah, after another look at parquet-cpp's CMakeLists.txt...

Looks like we need to add lz4 and zstd to Transitive Library Linkage defined in https://github.com/apache/parquet-cpp/blob/master/CMakeLists.txt#L559

advancedxy:
@xhochy This patch is ready for review and merging.

All tests are passed and I have updated the title and first post which would be nicely formatted in commit msg.

I would reword or drop benchmark code change commit if necessary

PR #421, opened 2017-12-03
PARQUET-1167: [C++] FieldToNode function should return a status when throwing an exception

cpcloud:


cpcloud:
@wesm @xhochy Can one of you review when you get a chance? Thanks!

src/parquet/arrow/arrow-schema-test.cc
```diff
@@ -807,5 +807,17 @@ TEST_F(TestConvertArrowSchema, ParquetFlatDecimals) {
   CheckFlatSchema(parquet_fields);
 }
 
+TEST(InvalidSchema, ParquetNegativeDecimalScale) {
+  const auto& type = ::arrow::decimal(23, -2);
+  const auto& field = ::arrow::field("f0", type);
+  const auto& arrow_schema = ::arrow::schema({field});
```
wesm: I don't believe the `&` has an effect here

xhochy: Actually this could also be risky. The return values here actually return an object someone needs to own, by only storing a reference, the ownership of the objects is completely open.

PR #423, opened 2017-12-10
PARQUET-1175: Fix arrow::ArrayData method rename from ShallowCopy to Copy

wesm:


wesm:
@cpcloud or @xhochy could you review? Thanks

PR #424, opened 2017-12-11
PARQUET-859: Flatten parquet/file directory, consolidate file reader, file writer code

wesm:
I believe this makes the codebase simpler and easier to navigate. By consolidating the file reader/writer code, further refactoring and internal improvements will be less obtuse


wesm:
This also places internal column chunk writing and reading in column_reader.cc, column_writer.cc, which is a better place for that code to live semantically

PR #426, opened 2017-12-15
PARQUET-1092: Support writing chunked arrow::Table columns

wesm:
I did quite a bit of refactoring to make this easier / simpler. I think there's some additional work we could do to make the write-path code cleaner, but we should probably wait to do some of that until we implement complete nested data write and read.

I will follow up shortly with some functional tests for ARROW-232 to make sure this works end-to-end on the pyarrow side.

### DIFF
--- a/src/parquet/arrow/arrow-reader-writer-test.cc
+++ b/src/parquet/arrow/arrow-reader-writer-test.cc
@@ -24,7 +24,10 @@
 #include "gtest/gtest.h"
 
 #include <arrow/compute/api.h>
+#include <cstdint>
+#include <functional>
 #include <sstream>
+#include <vector>
 
 #include "parquet/api/reader.h"
 #include "parquet/api/writer.h"
@@ -38,6 +41,7 @@
 
 #include "arrow/api.h"
 #include "arrow/test-util.h"
+#include "arrow/type_traits.h"
 #include "arrow/util/decimal.h"
 
 using arrow::Array;
@@ -45,6 +49,7 @@ using arrow::ArrayVisitor;
 using arrow::Buffer;
 using arrow::ChunkedArray;
 using arrow::Column;
+using arrow::DataType;
 using arrow::ListArray;
 using arrow::PoolBuffer;
 using arrow::PrimitiveArray;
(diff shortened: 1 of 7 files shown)


src/parquet/arrow/arrow-reader-writer-test.cc
```diff
@@ -325,6 +330,73 @@ void WriteTableToBuffer(const std::shared_ptr<Table>& table, int num_threads,
   *out = sink->GetBuffer();
 }
 
+namespace internal {
```
xhochy: Can we move this code into Arrow after we have done the releases?

src/parquet/arrow/writer.cc
```diff
@@ -106,21 +106,19 @@ class LevelBuilder {
...
   NOT_IMPLEMENTED_VISIT(Dictionary)
-  NOT_IMPLEMENTED_VISIT(Interval)
 
   Status GenerateLevels(const Array& array, const std::shared_ptr<Field>& field,
-                        int64_t* values_offset, ::arrow::Type::type* values_type,
-                        int64_t* num_values, int64_t* num_levels,
-                        std::shared_ptr<Buffer>* def_levels,
-                        std::shared_ptr<Buffer>* rep_levels,
+                        int64_t* values_offset, int64_t* num_values, int64_t* num_levels,
+                        const std::shared_ptr<PoolBuffer>& def_levels_scratch,
```
xhochy: I don't really understand the need for `def_levels_scratch` here. Can you give me some context why this is needed?

wesm: In the nested case, the definition levels are generated using a builder. Otherwise we can reuse the same pre-allocated memory for the levels. This could all be refactored potentially to reuse a common shared buffer, maybe we can do that later

src/parquet/arrow/arrow-reader-writer-test.cc
wesm: Yes, opened https://issues.apache.org/jira/browse/ARROW-1929 so as to not forget

wesm:
@xhochy good to merge this? Then can get a passing build in the ARROW-232 patch

PR #430, opened 2018-01-17
PARQUET-1193: [CPP] Implement ColumnOrder to support min_value and max_value

majetideepak:
Changes:

1. Update parquet.thrift format
2. Add ColumnOrder Implementation
3. Make Int96 sort order UNKNOWN

### DIFF
--- a/src/parquet/statistics-test.cc
+++ b/src/parquet/statistics-test.cc
@@ -277,7 +277,7 @@ void TestRowGroupStatistics<ByteArrayType>::TestMinMaxEncode() {
   ASSERT_EQ(statistics1.max(), statistics2.max());
 }
 
-using TestTypes = ::testing::Types<Int32Type, Int64Type, Int96Type, FloatType, DoubleType,
+using TestTypes = ::testing::Types<Int32Type, Int64Type, FloatType, DoubleType,
                                    ByteArrayType, FLBAType, BooleanType>;
 
 TYPED_TEST_CASE(TestRowGroupStatistics, TestTypes);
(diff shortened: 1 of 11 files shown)


src/parquet/statistics-test.cc
```diff
@@ -397,7 +397,7 @@ TEST(CorrectStatistics, Basics) {
   ASSERT_TRUE(column_chunk4->is_stats_set());
   auto column_chunk5 = ColumnChunkMetaData::Make(
       reinterpret_cast<const uint8_t*>(&col_chunk), schema.Column(4), &version);
-  ASSERT_TRUE(column_chunk5->is_stats_set());
+  ASSERT_FALSE(column_chunk5->is_stats_set());
```
xhochy: Why is this changed needed here?

src/parquet/types.h
```diff
@@ -130,6 +130,21 @@ struct SortOrder {
   enum type { SIGNED, UNSIGNED, UNKNOWN };
 };
 
+class ColumnOrder {
+ public:
```
xhochy: From where stems the need to have a full class for the column order? From looking at the code, it seems to me like using the enum should be sufficient.

src/parquet/statistics-test.cc
majetideepak: `column_chunk5` is defined for `LogicalType::INTERVAL`, which has sort order `UNKNOWN`.
I modified this PR to not set stats for all `UNKNOWN` sort orders.

src/parquet/types.h
majetideepak: Yes, enum is sufficient for this PR. I added the full class for further extensions. The Java implementation did something similar: 
https://github.com/apache/parquet-mr/pull/435/files#diff-37183c50b2eaab3ca048f9c5954e5b52R26

PR #433, opened 2018-01-23
PARQUET-1179: Upgrade to Thrift 0.11, use std::shared_ptr instead of boost::shared_ptr

wesm:


wesm:
The macOS build is failing due to a Bison issue, and the toolchain builds fail because conda-forge hasn't upgraded to 0.11 yet

xhochy:
@wesm Do we want to first merge on `conda-forge` or here?

xhochy:
The bison issue should be fixable by installing a bison version using `brew` and adding `/user/local/opt/bison/bin` to the PATH. Other people seem to have the same issue: https://github.com/argvk/jsonxx/issues/1#issuecomment-37492928

wesm:
We can merge on conda-forge first. We may need to pin thrift-cpp in Arrow until this gets in

xhochy:
Build still fails https://ci.appveyor.com/project/wesm/parquet-cpp/build/job/6ct7coqbty6pxd6u as thrift-cpp did not yet drop in conda-forge for Windows.

xhochy:
Green build: https://ci.appveyor.com/project/wesm/parquet-cpp/build/1.0.191

thamht4190:
I still get an issue after install the latest bison 
`
/Users/dev/tham/cortex-v2/src/ThirdParty/parquet-cpp/build/thrift_ep-prefix/src/thrift_ep/compiler/cpp/src/thrift/thrifty.yy:1.1-5: invalid directive: `%code'
/Users/dev/tham/cortex-v2/src/ThirdParty/parquet-cpp/build/thrift_ep-prefix/src/thrift_ep/compiler/cpp/src/thrift/thrifty.yy:1.7-14: syntax error, unexpected identifier
`
`
Mac-CoreTeam:ThirdParty dev$ brew upgrade bison
Error: bison 3.0.4_1 already installed
`
Do you have any idea? Thanks!

thamht4190:
I tried check `bison --version`, oh, it show bison 2.3. Then I found [this article on stackoverflow](https://stackoverflow.com/questions/10778905/why-not-gnu-bison-upgrade-to-2-5-on-macosx-10-7-3/30844621#30844621). It generates another errors but it passed bison issue ;)

xhochy:
@thamht4190 You need to do `export PATH=/usr/local/opt/bison/bin:$PATH` and then a fresh build to get rid of the error. Alternatively install Thrift via homebrew or conda and use that instead of the one that is built by parquet-cpp.

PR #438, opened 2018-02-18
PARQUET-1218: More informative error message on too short pages

xhochy:


src/parquet/column_reader.cc
```diff
@@ -182,7 +182,10 @@ std::shared_ptr<Page> SerializedPageReader::NextPage() {
     // Read the compressed data page.
     buffer = stream_->Read(compressed_len, &bytes_read);
     if (bytes_read != compressed_len) {
-      ParquetException::EofException();
+      std::stringstream ss;
+      ss << "Page was smaller (" << bytes_read << ") than expected (" << compressed_len
+         << ")";
+      ParquetException::EofException("Page was smaller than expected");
```
renato2099: Shouldn't this return:
ParquetException::EofException(ss.str());
?

xhochy: yes, it should. I'll fix this now.

wesm:
Appveyor build here: https://ci.appveyor.com/project/xhochy/parquet-cpp/build/1.0.78

PR #442, opened 2018-02-19
PARQUET-1226: Fixes for CHECKIN compiler warning level with clang 5.0

wesm:
This follows analogous work from ARROW-2117

### DIFF
--- a/src/parquet/util/comparison.h
+++ b/src/parquet/util/comparison.h
@@ -38,7 +38,6 @@ class PARQUET_EXPORT CompareDefault : public Comparator {
  public:
   typedef typename DType::c_type T;
   CompareDefault() {}
-  virtual ~CompareDefault() {}
   virtual bool operator()(const T& a, const T& b) { return a < b; }
 };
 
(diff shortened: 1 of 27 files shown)


src/parquet/util/comparison.h
```diff
@@ -38,15 +38,13 @@ class PARQUET_EXPORT CompareDefault : public Comparator {
  public:
   typedef typename DType::c_type T;
   CompareDefault() {}
-  virtual ~CompareDefault() {}
```
wesm: @cpcloud I wanted to confirm the the C++ standard says that declaring empty virtual dtors in subclasses is unnecessary (per associated comments on this recently in Arrow)

cpcloud: Yep.

Found [this SO answer](https://stackoverflow.com/a/27220184/564538), then I downloaded the standard and confirmed (the answer says subsection 9, but I found it in subsection 8) 

Section 12.4/8 says:

```
If a class has a base class with a
virtual destructor, its destructor (whether user- or implicitly-declared) is virtual.
```

Coupled with the implicit definitions of a destructor if they are used (and not already defined) this means defining an empty virtual destructor is unnecessary.

PR #443, opened 2018-02-20
PARQUET-1233: Enable option to switch between stl classes and boost c…

majetideepak:
…lasses for thrift header


CMakeLists.txt
```diff
@@ -399,6 +399,11 @@ else()
 endif()
 include(ThirdpartyToolchain)
 
+if (THRIFT_VERSION VERSION_LESS "0.11.0")
+  add_definitions(-DPARQUET_THRIFT_USE_BOOST)
+  message(STATUS "Using Boost in Thrift header")
+endif()
```
wesm: Our toolchain build seems to be broken. If we wanted to be rigorous, we could pin Thrift 0.10.0 in the toolchain part of the build matrix

CMakeLists.txt
```diff
@@ -399,6 +399,11 @@ else()
 endif()
 include(ThirdpartyToolchain)
 
+if (THRIFT_VERSION VERSION_LESS "0.11.0")
```
wesm: This might need to be a build option to toggle rather than using the version number. I think in Thrift 0.11.0 there might be a way to build with `boost:shared_ptr` support

majetideepak: `FORCE_BOOST_SMART_PTR` is the option to enforce boost in Thrift 0.11.0. Can users pass that via existing `PARQUET_CXXFLAGS`? The check `if defined(PARQUET_THRIFT_USE_BOOST) || defined(FORCE_BOOST_SMART_PTR)` in `thrift.h` examines that macro.

CMakeLists.txt
majetideepak: Makes sense to add this option. Will include that.

majetideepak: Ref: https://thrift.apache.org/lib/cpp

CMakeLists.txt
majetideepak: Ref: https://thrift.apache.org/lib/cpp

wesm:
Let's make sure that this works with no extra build flags with Thrift 0.11 in the toolchain, since that's the current stable release.

wesm:
This is failing for me with thrift 0.11 in toolchain, I'm going to push a fix

wesm:
This should be working now. I think having an explicit option to indicate that you're using the old Thrift API is the cleanest solution -- we could also regex out the version number from the output of `thrift --version`

cmake_modules/ThirdpartyToolchain.cmake
```diff
@@ -259,13 +259,19 @@ include_directories(SYSTEM ${THRIFT_INCLUDE_DIR} ${THRIFT_INCLUDE_DIR}/thrift)
 message(STATUS "Thrift include dir: ${THRIFT_INCLUDE_DIR}")
 message(STATUS "Thrift static library: ${THRIFT_STATIC_LIB}")
 message(STATUS "Thrift compiler: ${THRIFT_COMPILER}")
+message(STATUS "Thrift version: ${THRIFT_VERSION}")
```
majetideepak: Should this be `THRIFT_VENDOR_VERSION`?

wesm: No, this variable is set by `FindThrift.cmake`

wesm:
@xhochy @majetideepak is this good to go?

majetideepak:
+1 LGTM

PR #444, opened 2018-02-20
PARQUET-1225: NaN values may lead to incorrect filtering under certai…

majetideepak:
1) `parquet-cpp` does not implement filtering (predicate pushdown). Clients such as Vertica, read the statistics from the metadata and implement their own filtering based on these stats.
Therefore, the read path does not require any changes. We should document that the min/max value can potentially contain NaNs.

2) I made changes to the write path to ignore the NaNs.

### DIFF
--- a/src/parquet/statistics-test.cc
+++ b/src/parquet/statistics-test.cc
@@ -659,16 +659,17 @@ TEST_F(TestStatisticsFLBA, UnknownSortOrder) {
   ASSERT_FALSE(cc_metadata->is_stats_set());
 }
 
-// PARQUET-1225: Float NaN values may lead to incorrect filtering under certain circumstances
+// PARQUET-1225: Float NaN values may lead to incorrect filtering under certain
+// circumstances
 TEST(TestStatisticsFloatNaN, NaNValues) {
   constexpr int NUM_VALUES = 10;
   NodePtr node = PrimitiveNode::Make("nan_float", Repetition::OPTIONAL, Type::FLOAT);
   ColumnDescriptor descr(node, 1, 1);
-  float values[NUM_VALUES] =
-        {std::nanf(""), -4.0f, -3.0f, -2.0f, -1.0f, std::nanf(""), 1.0f, 2.0f, 3.0f, std::nanf("")};
+  float values[NUM_VALUES] = {std::nanf(""), -4.0f, -3.0f, -2.0f, -1.0f,
+                              std::nanf(""), 1.0f,  2.0f,  3.0f,  std::nanf("")};
   float nan_values[NUM_VALUES];
-  for (int i = 0; i < NUM_VALUES; i ++) {
-       nan_values[i] = std::nanf("");
+  for (int i = 0; i < NUM_VALUES; i++) {
+    nan_values[i] = std::nanf("");
   }
 
   // Test values
@@ -709,21 +710,21 @@ TEST(TestStatisticsFloatNaN, NaNValues) {
   ASSERT_EQ(max, 3.0f);
 }
 
-// PARQUET-1225: Float NaN values may lead to incorrect filtering under certain circumstances
+// PARQUET-1225: Float NaN values may lead to incorrect filtering under certain
+// circumstances
 TEST(TestStatisticsFloatNaN, NaNValuesSpaced) {
   constexpr int NUM_VALUES = 10;
   NodePtr node = PrimitiveNode::Make("nan_float", Repetition::OPTIONAL, Type::FLOAT);
   ColumnDescriptor descr(node, 1, 1);
-  float values[NUM_VALUES] =
-        {std::nanf(""), -4.0f, -3.0f, -2.0f, -1.0f, std::nanf(""), 1.0f, 2.0f, 3.0f, std::nanf("")};
+  float values[NUM_VALUES] = {std::nanf(""), -4.0f, -3.0f, -2.0f, -1.0f,
+                              std::nanf(""), 1.0f,  2.0f,  3.0f,  std::nanf("")};
   float nan_values[NUM_VALUES];
-  for (int i = 0; i < NUM_VALUES; i ++) {
-       nan_values[i] = std::nanf("");
+  for (int i = 0; i < NUM_VALUES; i++) {
+    nan_values[i] = std::nanf("");
   }
-  std::vector<uint8_t> valid_bits(
-       BitUtil::RoundUpNumBytes(NUM_VALUES) + 1, 255);
+  std::vector<uint8_t> valid_bits(BitUtil::RoundUpNumBytes(NUM_VALUES) + 1, 255);
 
-  // Test values 
+  // Test values
   TypedRowGroupStatistics<FloatType> nan_stats(&descr);
   nan_stats.UpdateSpaced(&values[0], valid_bits.data(), 0, NUM_VALUES, 0);
   float min = nan_stats.min();
@@ -767,8 +768,8 @@ TEST(TestStatisticsDoubleNaN, NaNValues) {
   NodePtr node = PrimitiveNode::Make("nan_double", Repetition::OPTIONAL, Type::DOUBLE);
   ColumnDescriptor descr(node, 1, 1);
   TypedRowGroupStatistics<DoubleType> nan_stats(&descr);
-  double values[NUM_VALUES] =
-        {std::nan(""), std::nan(""), -3.0, -2.0, -1.0, 0.0, 1.0, 2.0, 3.0, 4.0};
+  double values[NUM_VALUES] = {std::nan(""), std::nan(""), -3.0, -2.0, -1.0,
+                               0.0,          1.0,          2.0,  3.0,  4.0};
   double* values_ptr = &values[0];
   nan_stats.Update(values_ptr, NUM_VALUES, 0);
   double min = nan_stats.min();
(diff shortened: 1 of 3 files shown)


majetideepak:
I need to add a test when all values are NaNs.

src/parquet/statistics.cc
```diff
@@ -96,6 +96,67 @@ void TypedRowGroupStatistics<DType>::Reset() {
...
+  for (int64_t i = 0; i < count; i++) {
+     if (!std::isnan(values[i])) return i;
+  }
+  return count;
+}
+
+template <>
+inline int getValueEndOffset<float>(const float* values, int64_t count) {
+  // Skip NaNs
+  for (int64_t i = (count - 1); i > 0; i--) {
```
boroknagyz: I think it should be "i >= 0" instead of "i > 0"

src/parquet/statistics.cc
```diff
@@ -96,6 +96,67 @@ void TypedRowGroupStatistics<DType>::Reset() {
...
+     if (!std::isnan(values[i])) return (i + 1);
+  }
+  return count;
```
boroknagyz: To me it seems it should be "return 0"

src/parquet/statistics.cc
```diff
@@ -96,6 +96,67 @@ void TypedRowGroupStatistics<DType>::Reset() {
...
+  return !std::isnan(*value);
+}
+
+template <>
+inline int getValueBeginOffset<double>(const double* values, int64_t count) {
+  // Skip NaNs
+  for (int64_t i = 0; i < count; i++) {
+     if (!std::isnan(values[i])) return i;
+  }
+  return 0;
```
boroknagyz: i think it should be "return count"

src/parquet/statistics.cc
```diff
@@ -96,6 +96,67 @@ void TypedRowGroupStatistics<DType>::Reset() {
...
+}
+
+template <>
+inline int getValueEndOffset<double>(const double* values, int64_t count) {
+  // Skip NaNs
+  for (int64_t i = (count - 1); i > 0; i--) {
```
boroknagyz: I think it should be `i >= 0` instead of `i > 0`
Think of the case when only the first element is a number, e.g.:
`{3.14}`, or
`{3.14, NaN, NaN, NaN, NaN, ..., NaN}`
For these inputs, this function will return 0.
`getValueBeginOffset()` will also return 0.
Usually, C++ ranges are interpreted as `[first, last)`, ie. they are open at the end. Therefore, `[0, 0)` is an empty range.
However, this solution will work, because at L178 you don't return when `begin_offset_ == end_offset`, and `minmax_element()` returns `make_pair(first, first)` if the range is empty.

But I feel like it currently works by accident, and would be better to fix this.

src/parquet/statistics.cc
```diff
@@ -107,8 +168,17 @@ void TypedRowGroupStatistics<DType>::Update(const T* values, int64_t num_not_nul
   // TODO: support distinct count?
   if (num_not_null == 0) return;
 
+  // PARQUET-1225: Handle NaNs
+  // The problem arises only if the starting/ending value(s)
```
zivanfi: Do NaN-s at the end actually cause trouble? I had the impression that only NaN-s at the beginning are problematic.

src/parquet/statistics-test.cc
```diff
@@ -659,5 +659,37 @@ TEST_F(TestStatisticsFLBA, UnknownSortOrder) {
...
+        {std::nan(""), std::nan(""), -3.0, -2.0, -1.0, 0.0, 1.0, 2.0, 3.0, 4.0};
+  double* values_ptr = &values[0];
+  nan_stats.Update(values_ptr, NUM_VALUES, 0);
+  double min = nan_stats.min();
+  double max = nan_stats.max();
+
+  ASSERT_EQ(min, -3.0);
+  ASSERT_EQ(max, 4.0);
+}
 }  // namespace test
```
zivanfi: I would suggest adding a test case for all values being NaN.

zivanfi: Ah, sorry, I just noticed you listed this yourself as something that is still left to be done.

src/parquet/statistics.cc
majetideepak: correct!

src/parquet/statistics.cc
majetideepak: They are a problem even in the end. The max value becomes NaN. This is specific to the implementation of `minmax_element`. A possible implementation is specified here http://en.cppreference.com/w/cpp/algorithm/minmax_element

src/parquet/statistics.cc
majetideepak: Fixed!

src/parquet/statistics.cc
majetideepak: Fixed!

src/parquet/statistics.cc
majetideepak: Fixed!

src/parquet/statistics.cc
```diff
@@ -96,6 +96,67 @@ void TypedRowGroupStatistics<DType>::Reset() {
   has_min_max_ = false;
 }
 
+template <typename T>
+inline int getValueBeginOffset(const T* values, int64_t count) {
```
wesm: Can we use compliant naming in these new functions (function names should be capitalized)?

src/parquet/statistics.cc
```diff
@@ -96,6 +96,67 @@ void TypedRowGroupStatistics<DType>::Reset() {
...
+template <typename T>
+inline bool notNaN (const T* value) {
+  return true;
+}
+
+template <>
+inline int getValueBeginOffset<float>(const float* values, int64_t count) {
+  // Skip NaNs
+  for (int64_t i = 0; i < count; i++) {
+     if (!std::isnan(values[i])) return i;
```
wesm: Add braces

src/parquet/statistics.cc
```diff
@@ -96,6 +96,67 @@ void TypedRowGroupStatistics<DType>::Reset() {
...
+  }
+  return count;
+}
+
+template <>
+inline int getValueEndOffset<float>(const float* values, int64_t count) {
+  // Skip NaNs
+  for (int64_t i = (count - 1); i > 0; i--) {
+     if (!std::isnan(values[i])) return (i + 1);
```
wesm: Braces

src/parquet/statistics.cc
```diff
@@ -96,6 +96,67 @@ void TypedRowGroupStatistics<DType>::Reset() {
...
+  return 0;
+}
+
+template <typename T>
+inline int getValueEndOffset(const T* values, int64_t count) {
+  return count;
+}
+
+template <typename T>
+inline bool notNaN (const T* value) {
```
wesm: Pass `const T value` instead of pointer?

src/parquet/statistics.cc
```diff
@@ -96,6 +96,67 @@ void TypedRowGroupStatistics<DType>::Reset() {
...
+  // Skip NaNs
+  for (int64_t i = (count - 1); i > 0; i--) {
+     if (!std::isnan(values[i])) return (i + 1);
+  }
+  return count;
+}
+
+template <>
+inline bool notNaN<float>(const float* value) {
+  return !std::isnan(*value);
```
wesm: Would it be better to use `return value == value` here (with the pointer -> value change per above)?

src/parquet/statistics.cc
```diff
@@ -96,6 +96,67 @@ void TypedRowGroupStatistics<DType>::Reset() {
...
+}
+
+template <>
+inline int getValueBeginOffset<double>(const double* values, int64_t count) {
+  // Skip NaNs
+  for (int64_t i = 0; i < count; i++) {
+     if (!std::isnan(values[i])) return i;
```
wesm: Braces

src/parquet/statistics.cc
```diff
@@ -96,6 +96,67 @@ void TypedRowGroupStatistics<DType>::Reset() {
...
+  }
+  return 0;
+}
+
+template <>
+inline int getValueEndOffset<double>(const double* values, int64_t count) {
+  // Skip NaNs
+  for (int64_t i = (count - 1); i > 0; i--) {
+     if (!std::isnan(values[i])) return (i + 1);
```
wesm: braces

src/parquet/statistics.cc
```diff
@@ -96,6 +96,67 @@ void TypedRowGroupStatistics<DType>::Reset() {
...
+  return count;
+}
+
+template <typename T>
+inline bool notNaN (const T* value) {
+  return true;
+}
+
+template <>
+inline int getValueBeginOffset<float>(const float* values, int64_t count) {
```
wesm: If we used `std::is_floating_point` and the functor pattern, we could avoid code duplication

```c++
template <typename T, typename Enable = void>
struct StatsHelper {
  ...
};

template <typename T>
struct StatsHelper<T, typename std::enable_if<std::is_floating_point<T>::value>::type> {
  ...
};
```

majetideepak: Thanks for this tip. Very helpful! Will include other review comments.

majetideepak:
All the changes have been made. @boroknagyz  and @wesm  please let me know if you have more feedback!

boroknagyz:
Thanks for applying the changes! To me it looks good generally.
The Update() and UpdateSpaced() functions share some common code parts (not necessarily introduced by this PR, e.g. the last if statements of the functions), maybe worth a little refactoring.

xhochy:
@majetideepak @boroknagyz @zivanfi I think with the PR we are now ready to make a new RC?

majetideepak:
I think we are ready to make a new RC. thanks!

zivanfi:
I agree, thanks for your efforts!

PR #445, opened 2018-02-28
PARQUET-1166: Add GetRecordBatchReader in parquet/arrow/reader

advancedxy:


src/parquet/arrow/writer.h
```diff
@@ -31,7 +31,6 @@ namespace arrow {
 class Array;
 class MemoryPool;
 class PrimitiveArray;
-class RowBatch;
```
advancedxy: RowBatch is never used any more and is renamed to RecordBatch

advancedxy:
Ping @wesm @xhochy, do you have any comments?

wesm:
Will review soon, have been underwater with the Arrow 0.9.0 backlog. Is there a possibility that individuals from Baidu may become more involved in Arrow development as well?

advancedxy:
> Is there a possibility that individuals from Baidu may become more involved in Arrow development as well?

I can't speak for my company. I will have to check with my manager and technical leader. However big company always wants something in return: reputation/business benefits etc. 

However, me as an individual can be more involved in Arrow development in my spare time if some help is needed.

src/parquet/arrow/reader.h
```diff
@@ -149,6 +152,13 @@ class PARQUET_EXPORT FileReader {
   ::arrow::Status ReadSchemaField(int i, const std::vector<int>& indices,
                                   std::shared_ptr<::arrow::Array>* out);
 
+  ::arrow::Status GetRecordBatchReader(const std::vector<int>& row_group_indices,
+                                       std::shared_ptr<::arrow::RecordBatchReader>* out);
+
+  ::arrow::Status GetRecordBatchReader(const std::vector<int>& row_group_indices,
+                                       const std::vector<int>& column_indices,
+                                       std::shared_ptr<::arrow::RecordBatchReader>* out);
```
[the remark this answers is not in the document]
advancedxy: Of course, will do.

src/parquet/arrow/reader.cc
```diff
@@ -152,6 +153,64 @@ class SingleRowGroupIterator : public FileColumnIterator {
   bool done_;
 };
 
+class RowGroupRecordBatchReader : public ::arrow::RecordBatchReader {
+ public:
+    explicit RowGroupRecordBatchReader(const std::vector<int>& row_group_indices,
```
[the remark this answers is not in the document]
advancedxy: will do

src/parquet/arrow/reader.cc
```diff
@@ -152,6 +153,64 @@ class SingleRowGroupIterator : public FileColumnIterator {
...
+        }
+      }
+
+      // all row groups has been consumed
+      if (next_row_group_ == row_group_indices_.size()) {
+        *out = nullptr;
+        return Status::OK();
+      }
+
+      RETURN_NOT_OK(file_reader_->ReadRowGroup(row_group_indices_[next_row_group_],
```
advancedxy: I am most concern about this one. We have to read one entire row group, but the caller may consume only the first N RecordBatches.

I am wondering that this is not optimal

advancedxy:
>I can't speak for my company. I will have to check with my manager and technical leader. However big company always wants something in return: reputation/business benefits etc.

Sorry for the delay(busy setting up cluster spark app profiling).  I checked with my manager, the general response is that: 
> with limit amount of dev resource, we will contribute back internally features if suitable,  but should not actively work on community issues.

However, I will try to figure out if there is issues I can work on in my spare time

advancedxy:
@wesm @xhochy do you have any other comments, in particular with:
1. Document wording since I am not a native speaker
2. The approach with generating table of row group first, then constructing a TableBatchReader
3. More test cases should be added?

If not, I will update the PR title, and maybe rebase commits to get a better commit history, then it should be ready for merging.

advancedxy:
ping @wesm @xhochy

xhochy:
Looks good from my side, if @wesm does not object, I'll merge tomorrow.

src/parquet/arrow/reader.h
```diff
@@ -149,6 +154,21 @@ class PARQUET_EXPORT FileReader {
...
+
+  /// \brief Return a RecordBatchReader of row groups selected from row_group_indices,
+  ///     whose columns are selected by column_indices. The ordering in row_group_indices
+  ///     and column_indices matter.
+  /// \returns error Status if either row_group_indices or column_indices contains invalid
+  ///    index
+  ::arrow::Status GetRecordBatchReader(const std::vector<int>& row_group_indices,
+                                       const std::vector<int>& column_indices,
+                                       std::shared_ptr<::arrow::RecordBatchReader>* out);
```
wesm: My main critique of these APIs is that we will want to provide for setting the number of rows to be read for each call to `ReadNext`, for example 1,000,000 rows at a time. Right now this is returning a whole row group at a time

wesm: Could you please open a JIRA about improving this code in this regard?

advancedxy: >  My main critique of these APIs is that we will want to provide for setting the number of rows to be read for each call to `ReadNext`

Ah, I do consider this when implementing this. However the [`RecordBatch`](https://github.com/apache/arrow/blob/master/cpp/src/arrow/record_batch.h#L166) interface in arrow doesn't expose that. And I'd like to hide impl details in `parquet/arrow/reader`. To enable this, I'd like to proposal new method to `RecordBatch` then.
What do you think? @wesm 

> Could you please open a JIRA about improving this code in this regard?

Will do.

PR #447, opened 2018-03-08
PARQUET-1245: Fix creating Arrow table with duplicate column names

pitrou:


cpcloud:
Thanks for doing this. Will review shortly

pitrou:
Unfortunately this doesn't seem sufficient. If I add the following test, I get an error and a crash:
```diff
diff --git a/src/parquet/arrow/arrow-reader-writer-test.cc b/src/parquet/arrow/arrow-reader-writer-test.cc
index 72e65d4..eb5a8ec 100644
--- a/src/parquet/arrow/arrow-reader-writer-test.cc
+++ b/src/parquet/arrow/arrow-reader-writer-test.cc
@@ -1669,6 +1669,27 @@ TEST(TestArrowReadWrite, TableWithChunkedColumns) {
   }
 }
 
+TEST(TestArrowReadWrite, TableWithDuplicateColumns) {
+  using ::arrow::ArrayFromVector;
+
+  auto f0 = field("duplicate", ::arrow::int8());
+  auto f1 = field("duplicate", ::arrow::int16());
+  auto schema = ::arrow::schema({f0, f1});
+
+  std::vector<int8_t> a0_values = {1, 2, 3};
+  std::vector<int16_t> a1_values = {14, 15, 16};
+
+  std::shared_ptr<Array> a0, a1;
+
+  ArrayFromVector<::arrow::Int8Type, int8_t>(a0_values, &a0);
+  ArrayFromVector<::arrow::Int16Type, int16_t>(a1_values, &a1);
+
+  auto table = Table::Make(schema,
+                           {std::make_shared<Column>(f0->name(), a0),
+                            std::make_shared<Column>(f1->name(), a1)});
+  CheckSimpleRoundtrip(table, table->num_rows());
+}
+
 TEST(TestArrowWrite, CheckChunkSize) {
   const int num_columns = 2;
   const int num_rows = 128;
```

pitrou:
Ok, the reason for the error is that a similar pattern needs fixing in `SchemaDescriptor`. Updating shortly.

src/parquet/schema.h
```diff
@@ -264,8 +264,11 @@ class PARQUET_EXPORT GroupNode : public Node {
   bool Equals(const Node* other) const override;
 
   NodePtr field(int i) const { return fields_[i]; }
+  // Get the index of a field by its name, or negative value if not found
+  // If several fields share the same name, the smallest index is returned
```
cpcloud: Couple of questions:

* I see [this language regarding the iteration order](http://en.cppreference.com/w/cpp/container/unordered_multimap) of the values for a particular key in the multimap:

> every group of elements whose keys compare equivalent (compare equal with key_eq() as the comparator) forms a contiguous subrange in the iteration order

Does the `iteration order` here mean that the values are iterated over in the order in which they were inserted?

* Why did you choose to return the first one instead of returning `-1` (or maybe `-2`) for the `std::string` overload? Do we not want to provide a way to indicate that column indexes and column names are not 1:1 in the C++ API? Maybe that already exists.

pitrou: 1) That's a good point. The fact that the container is unordered means it isn't guaranteed to retain insertion order, even for values which map to the same key (I would expect a straightforward implementation to maintain that order, though). I should probably remove the sentence above.

2) Because doing otherwise seems like it could break compatibility. Not sure how strongly you feel about it. The `std::string` overloads aren't used anymore in the parquet-cpp codebase, AFAICT.

cpcloud: > it could break compatibility

True, though IIUC wouldn't this potentially segfault if you tried to use the result to index into something?

pitrou: Why? We're still returning a valid index.

pitrou: As a side note, if @wesm wants to include this in the release, we can further API improvements to a later PR.

cpcloud: I meant that before your change, the index returned was always valid?

pitrou: Yes, it was, it just wasn't necessarily the one expected by the caller according to its semantics.

cpcloud: Right, makes sense.

src/parquet/schema.cc
```diff
@@ -720,17 +718,15 @@ int SchemaDescriptor::ColumnIndex(const std::string& node_path) const {
...
-int SchemaDescriptor::ColumnIndex(const Node& node) const {
-  int result = ColumnIndex(node.path()->ToDotString());
-  if (result < 0) {
-    return -1;
-  }
-  DCHECK(result < num_columns());
-  if (!node.Equals(Column(result)->schema_node().get())) {
-    // Same path but not the same node
-    return -1;
+int SchemaDescriptor::ColumnIndex(const Node* node) const {
```
xhochy: Why did this change from reference to pointer? We use references everywhere where the passed object cannot be null.

pitrou: Hmm... I didn't know this guideline (but out parameters are passed as pointers apparently?).
The search is done using pointer equality, so I thought passing a pointer would be more explicit (and perhaps less error-prone, in case the compiler does a temporary copy).

pitrou: By the way other methods such `Node::Equals` take a node pointer, even though passing a null pointer isn't supported. Should I still convert back to a reference?

xhochy: @wesm What is your opinion on this?

wesm: Sorry for the delay on this. I think we should consistently use const-references (unless null is a valid input), comparing pointer equality with `&value` where necessary. See commentary in https://google.github.io/styleguide/cppguide.html#Reference_Arguments

pitrou: Ok, I've turned the argument back into a const-reference.

wesm:
@cpcloud @pitrou we may want to bump parquet-cpp in conda-forge to pick up this patch at some point

PR #450, opened 2018-03-27
PARQUET-1256: Add --print-key-value-metadata option to parquet_reader tool

JacekPliszka:
This is a minor change useful for debugging.

Now parquet_reader tool has --print-key-value-metadata which when present, dump of key values of file metadata is done.

Created https://issues.apache.org/jira/browse/PARQUET-1256

### DIFF
--- a/src/parquet/printer.cc
+++ b/src/parquet/printer.cc
@@ -33,13 +33,25 @@ namespace parquet {
 #define COL_WIDTH "30"
 
 void ParquetFilePrinter::DebugPrint(std::ostream& stream, std::list<int> selected_columns,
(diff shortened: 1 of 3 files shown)


majetideepak:
@JacekPliszka Can you resolve this conflict? I will be happy to fix this if you cannot at this point.

JacekPliszka:
@majetideepak   Yes, please, if you could. Looks like I should use long int/int64 or make it different type conditionally.

wesm:
I ran clang-format and rebased

wesm:
one of the Travis entries failed due to unrelated flakiness

PR #451, opened 2018-04-06
PARQUET-1267: [C++] replace "unsafe" std::equal by std::memcmp

rip-nsk:


jbapple-cloudera:
What's the actual text of the warning?

rip-nsk:
'std::equal::_Unchecked_iterators::_Deprecate': Call to 'std::equal' with parameters that may be unsafe - this call relies on the caller to check that the passed values are correct.

jbapple-cloudera:
How is `memcmp` any safer? This sounds like an issue with the compiler, rather than the code.

rip-nsk:
std::memcmp is not safer, it just requires the same size arguments.
std::equal has (implicitly) the same requirements, but since c++14 introduces a more secure version, this one is marked as deprecate.
Both variants are valid (since we check size of arrays first).

PR #456, opened 2018-04-16
PARQUET-1274: Prevent segfault that was occurring when writing a nanosecond timestamp with arrow writer properties set to coerce timestamps and support deprecated int96 timestamps.

joshuastorck:
The bug was a due to the fact that the physical type was int64 but the WriteTimestamps function was taking a path that assumed the physical type was int96. This caused memory corruption because it was writing past the end of the array. The bug was fixed by checking that coerce timestamps is disabled when writing int96. 

A unit test was added for the regression.

### DIFF
--- a/src/parquet/arrow/arrow-reader-writer-test.cc
+++ b/src/parquet/arrow/arrow-reader-writer-test.cc
@@ -1405,7 +1405,14 @@ TEST(TestArrowReadWrite, ConvertedDateTimeTypes) {
 
 // Regression for ARROW-2802
 TEST(TestArrowReadWrite, CoerceTimestampsAndSupportDeprecatedInt96) {
-  using namespace ::arrow;
+  using ::arrow::Column;
+  using ::arrow::Field;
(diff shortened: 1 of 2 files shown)


src/parquet/arrow/arrow-reader-writer-test.cc
```diff
@@ -1403,6 +1403,56 @@ TEST(TestArrowReadWrite, ConvertedDateTimeTypes) {
   AssertTablesEqual(*ex_table, *result);
 }
 
+// Regression for ARROW-2802
+TEST(TestArrowReadWrite, CoerceTimestampsAndSupportDeprecatedInt96) {
+  using namespace ::arrow;
```
xhochy: Please list the types you use explicitly here, e.g. `using ::arrow::TimestampType`.

xhochy:
Thanks @joshuastorck 

I moved the ARROW issue over to the PARQUET tracker and also gave you karma to assign issues to yourself there, too.

PR #457, opened 2018-04-17
PARQUET-1272: Return correct row count for nested columns in ScanFileContents

xhochy:
Stumbled over this while adding lists to the `alltypes_sample` in `test_parquet.py` in Arrow.


pitrou:
Will help remove the workaround in https://github.com/apache/arrow/pull/1891 :-)

PR #458, opened 2018-04-18
PARQUET-1279: [C++] Adding use of ASSERT_NO_FATAL_FAILURE in unit tests when calling helper functions that call ASSERT_ macros

joshuastorck:
There were a number of helper functions in the unit tests that has ASSERT_ macros. However, the callers of these functions were not asserting if the helper functions had encountered an assertion. In those circumstances, the unit test will continue because the ASSERT_ macros only return in the block in which they are used. This commit adds the use of ASSERT_NO_FATAL_FAILURE to the calls of those helper functions at the scope of the unit test, which will cause the unit test to exit. This ensures that we won't encounter segfaults or unexpected behavior due to a unit test continuing execution past an asserting inside of a helper function.

### DIFF
--- a/src/parquet/arrow/arrow-reader-writer-test.cc
+++ b/src/parquet/arrow/arrow-reader-writer-test.cc
@@ -405,7 +405,8 @@ void DoSimpleRoundtrip(const std::shared_ptr<Table>& table, int num_threads,
                        const std::shared_ptr<ArrowWriterProperties>& arrow_properties =
                            default_arrow_writer_properties()) {
   std::shared_ptr<Buffer> buffer;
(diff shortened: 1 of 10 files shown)


xhochy:
@joshuastorck Can you open a JIRA for this?

joshuastorck:
Created a JIRA and updated this PR's title to associate it as a fix.

xhochy:
@joshuastorck This is for the wrong project. This here is PARQUET, not ARROW. I have moved the ticket.

xhochy:
Thanks for fixing this. This is really an unexpected behaviour I would not have thought of.

PR #459, opened 2018-04-19
PARQUET-1276: [C++] Reduce the amount of memory used for writing null decimal values

cpcloud:
Also fixes some clang 6 compile time errors


cpcloud:
cc @xhochy

src/parquet/arrow/test-util.h
```diff
@@ -69,7 +69,7 @@ typename std::enable_if<is_arrow_float<ArrowType>::value, Status>::type NonNullA
   ::arrow::test::random_real(size, 0, static_cast<c_type>(0), static_cast<c_type>(1),
                              &values);
   ::arrow::NumericBuilder<ArrowType> builder;
-  RETURN_NOT_OK(builder.Append(values.data(), values.size()));
+  RETURN_NOT_OK(builder.AppendValues(values.data(), values.size()));
```
pitrou: Sorry, I had overlooked parquet-cpp's use of this API.

src/parquet/schema.cc
```diff
@@ -113,7 +113,7 @@ PrimitiveNode::PrimitiveNode(const std::string& name, Repetition::type repetitio
 
   // PARQUET-842: In an earlier revision, decimal_metadata_.isset was being
   // set to true, but Impala will raise an incompatible metadata in such cases
-  memset(&decimal_metadata_, 0, sizeof(decimal_metadata_));
+  std::memset(&decimal_metadata_, 0, sizeof(decimal_metadata_));
```
xhochy: Don't use `std::` here. People reported problems on the usage of `std::memset` (seems like some BSD variants don't have this).

cpcloud: Thanks for the heads up, will change.

pitrou:
@cpcloud, this needs rebasing. Are you available to do it soon?

wesm:
This patch should be able to be whittled down to a small change (< 30 LOC or so) since it does work redundant with some other patches merged recently

pitrou:
Superseded by https://github.com/apache/parquet-cpp/pull/493

PR #465, opened 2018-05-17
PARQUET-979: Limit size of min, max or disable stats for long binary types

majetideepak:


src/parquet/column_writer.cc
```diff
@@ -33,6 +33,8 @@
 
 namespace parquet {
 
+static constexpr int MAX_STATS_SIZE = 4096;  // limit stats to 4k
```
xhochy: Is this defined by the Parquet standard or arbitrarily chosen?

majetideepak: I followed parquet-mr.
https://github.com/apache/parquet-mr/blob/0d55abd05b0e5027c18e60d1ac3b22998dd00951/parquet-hadoop/src/main/java/org/apache/parquet/format/converter/ParquetMetadataConverter.java#L88
I don't see this specified in the spec. We should probably add it there. I will open a JIRA to do this.

xhochy: The alternative would be to use this as sensible default and add it as an option to `WriterProperties`.

majetideepak: I added it to `ColumnProperties` since it gives more flexibility.

src/parquet/properties.h
```diff
@@ -348,19 +357,23 @@ class PARQUET_EXPORT WriterProperties {
...
-    return column_properties(path).dictionary_enabled;
+    return column_properties(path).dictionary_enabled_;
   }
 
   bool statistics_enabled(const std::shared_ptr<schema::ColumnPath>& path) const {
-    return column_properties(path).statistics_enabled;
+    return column_properties(path).statistics_enabled_;
+  }
+
+  size_t max_stats_size(const std::shared_ptr<schema::ColumnPath>& path) const {
```
xhochy: Please use `max_statistics_size`

PR #467, opened 2018-05-22
PARQUET-1308: [C++] Use Arrow thread pool, not Arrow ParallelFor, fix deprecated APIs, upgrade clang-format version. Fix record delimiting bug

pitrou:


pitrou:
Ok, CI fails because it doesn't run against Arrow master. Should we wait until the next Arrow release before we revisit this?

xhochy:
We should wait until we know how far away a new Arrow release is. At the moment, I would personally like to see a new Arrow release as soon as as possible but our release depends on getting the packaging automation working.

wesm:
@xhochy @pitrou want to rebase this as Arrow release is imminent (I would guess next week)?

pitrou:
Ok, rebased.

wesm:
Looks like this is forcing a resolution of https://issues.apache.org/jira/browse/PARQUET-1354. I can take a look later today...

wesm:
@majetideepak @xhochy @pitrou I didn't think it would be controversial but I upgraded the clang-format version used in parquet-cpp to be 6.0 to align with Arrow. Every time I open this codebase it pains me to see a half-duplicated build system

wesm:
Looks like there's an issue with Brotli on Windows

`LINK : fatal error LNK1181: cannot open input file 'C:\Miniconda35-x64\Library\lib\brotlidec-static.lib`

wesm:
Looks like we also have a test failure with latest Arrow

```
[ RUN      ] TestArrowReadWrite.ListLargeRecords
/Users/travis/build/apache/parquet-cpp/src/parquet/arrow/arrow-reader-writer-test.cc:1749: Failure
Value of: table->Equals(*chunked_table)
  Actual: false
Expected: true
[  FAILED  ] TestArrowReadWrite.ListLargeRecords (9 ms)
```

pitrou:
> Looks like we also have a test failure with latest Arrow

Annoyingly, I can't reproduce...

wesm:
Rebased

wesm:
I'm not able to reproduce this with either gcc 4.8.5 or clang-6. The failure is only happening in the Xcode 6.4 build. This is a bit of a pickle because I don't have access to a macOS machine. @xhochy can you see if you can repro since you develop on mac?

wesm:
I'm going to try to see if I can get macOS running in VMWare so I can help debug in the next few days. Will keep an eye on this issue

xhochy:
@wesm @pitrou I can reproduce and look into it.

wesm:
I am back to a machine with macOS; I will see if I can figure this out later today or tomorrow

wesm:
Well "today or tomorrow" didn't happen =) I am going to try today

wesm:
OK I have reproduced the failure locally. I'll try to diagnose the root cause

wesm:
I've got a reproduction that fails on Linux also. This should make debugging a lot easier

pitrou:
Indeed :-) Can you post the reproducer?

wesm:
Yep, I'm trying to whittle it down to be as minimal as possible

wesm:
I find this failure very odd. It's difficult to bisect to find whether it's an Arrow commit that introduced the regression. What I'm going to try to do (unless you have other ideas) is step through this unit test with the old Arrow version to try to find the point at which the results diverge in the decoding. Pretty laborious

pitrou:
Hmm, I'm confused. I managed to build parquet-cpp against apache-arrow-0.8.0 and the reproducer already fails...

pitrou:
By the way, here, the cutoff is very precisely at `num_rows = 1740` (1739 works fine, 1740 fails).

pitrou:
Update: with apache-arrow-0.8.0 and parquet-cpp changeset c064f82, the reproducer already fails.

wesm:
OK, it sounds like there is simply a bug that was undiagnosed. I'm going to spend some time on this today to get to the bottom of it

wesm:
So here's the offending value with the patch where it stands now:

```
Value 1899
  [
    [
      95,
      42,
      3,
      1,
      null,
      30,
      38,
      null,
      null,
      3,
      35,
      99,
      null,
      9
    ]
  ]
==================
  [
    [
      95
    ]
  ]
```

Only a single value of out 2000 is wrong. So my guess is there's an encoding or decoding edge case. I'll dig in and see where the problem is happening

wesm:
There's an edge case when reading the last record in a row group. You'll notice that the unit test is writing row groups with size 100 and the bug occurs on the 100th value in the row group. I'm having to jog my memory from when I wrote the record reader logic but I'll push a fix as soon as I figure it out

wesm:
Found the bug. The record delimiting logic was incorrect. I added some comments to explain what's going on in there.

This is a quite serious correctness bug so we're probably going to want to bump the conda-forge parquet-cpp version to include this fix to minimize the number of affected users

I'm going to clean up the unit test to get this into merge-ready state.

wesm:
Oof, seems I spoke too soon. There's more bugs, I'll keep working on it

wesm:
Finally! OK, I think we are good now. @cpcloud if you have a chance could you review this so we can get it in and unblock the other failing PRs?

wesm:
OK, this is passing except for the Brotli toolchain issue on Windows. merging

PR #469, opened 2018-05-29
PARQUET-1315: ColumnChunkMetaData.has_dictionary_page() should return…

majetideepak:
… bool, not int64_t


jamesclampffer:
This makes sense to me.

PR #471, opened 2018-06-21
PARQUET-1334: [C++] memory_map parameter seems missleading in parquet file opener

philhoch:
If memory_map parameter is true, normal file operation is executed, while in negative case, the according memory mapped file operation happens. Seems either be used via inverted logic or being bug.


majetideepak:
+1 LGTM. Definitely a bug that slipped through until now. Thanks for contributing this!

PR #472, opened 2018-06-21
account for total_size being 0, having no dictionary entries to allocate for

philhoch:
The call with size 0 ends up in arrows memory_pool, https://github.com/apache/arrow/blob/884474ca5ca1b8da55c0b23eb7cb784c2cd9bdb4/cpp/src/arrow/memory_pool.cc#L50, and the according allocation fails. See according documentation, https://docs.microsoft.com/en-us/cpp/c-runtime-library/reference/aligned-malloc. Only happens on Windows environment, as posix_memalign seems to handle 0 inputs in unix environments.


majetideepak:
@philhoch  can you file two JIRAs (here: https://issues.apache.org/jira/projects/PARQUET/) for the two bugs? Thanks.

majetideepak:
+1 LGTM

PR #477, opened 2018-07-17
PARQUET-1350: [C++] Use abstract ResizableBuffer instead of concrete PoolBuffer

pitrou:


cpcloud:
@xhochy This is the PR that is causing release headaches now. We now have to 1) build against parquet-cpp master, 2) cut a new release of parquet-cpp, or 3) revert the arrow commit. I'm not sure that it's benefiting the community to keep these projects separate anymore given the pace of change in both. We need some solution going forward. A monorepo is going to make the repo bigger, but we will be able to catch API breakages like this immediately.

wesm:
We will have to build against parquet-cpp master in the meantime. 

Having parquet-cpp depend on the "platform" code (IO, memory management, concurrency management, algorithms) in Arrow is the only really practical option but the awkward circular dependency. 

As commented in https://github.com/apache/arrow/pull/2326#issuecomment-408138815 moving these communities to a monorepo structure would make all of these problems go away immediately. With the release automation work in Arrow we could also release much more often

PR #479, opened 2018-07-26
PARQUET-1357: FormatStatValue truncates binary statistics on zero character

xhochy:


xhochy:
This PR also includes the clang tools update. Kept it in a separate commit to make review simpler.

PR #482, opened 2018-07-28
PARQUET-1360: Use conforming API style, variable names in WriteFileMetaData functions

wesm:
Per comments in PARQUET-1348. I also upgraded clang-format to 6.0 here so pardon the diff noise from that.

To summarize the style points:

* Don't pass `const std:shared_ptr<T>&` if `const T&` will do
* Don't pass `const std::shared_ptr<T>&` if `T*` will do
* I prefer to only use class static function members for alternate constructors
* Out arguments after in arguments
* `lower_with_underscores` for variable names


rgruener:
Thanks! I would have done these changes myself had I known about (or spent more time to look at) the style standards.

wesm:
No problemo, the style issues could be better documented (all we really have is https://github.com/apache/parquet-cpp#developer-notes).

PR #484, opened 2018-08-13
PARQUET-1372:  Add an API to allow writing RowGroups based on size

majetideepak:
I split the changes into multiple commits to ease the review.
Used the example program to test the new API.
I will add unit tests once we converge on the API after review. 
Thanks to @anatolishein for collaborating with the API design.

### DIFF
--- a/examples/low-level-api/CMakeLists.txt
+++ b/examples/low-level-api/CMakeLists.txt
@@ -15,6 +15,8 @@
 # specific language governing permissions and limitations
 # under the License.
 
+include_directories(SYSTEM . )
+
 if (PARQUET_BUILD_EXECUTABLES)
   add_executable(reader-writer reader-writer.cc)
   add_executable(reader-writer2 reader-writer2.cc)
--- a/examples/low-level-api/CMakeLists.txt
+++ b/examples/low-level-api/CMakeLists.txt
@@ -15,11 +15,11 @@
 # specific language governing permissions and limitations
 # under the License.
 
-include_directories(SYSTEM . )
-
 if (PARQUET_BUILD_EXECUTABLES)
   add_executable(reader-writer reader-writer.cc)
   add_executable(reader-writer2 reader-writer2.cc)
+  target_include_directories(reader-writer PRIVATE .)
+  target_include_directories(reader-writer2 PRIVATE .)
   target_link_libraries(reader-writer parquet_static)
   target_link_libraries(reader-writer2 parquet_static)
 endif()
--- a/examples/low-level-api/CMakeLists.txt
+++ b/examples/low-level-api/CMakeLists.txt
@@ -17,5 +17,7 @@
 
 if (PARQUET_BUILD_EXECUTABLES)
   add_executable(reader-writer reader-writer.cc)
+  add_executable(reader-writer2 reader-writer2.cc)
   target_link_libraries(reader-writer parquet_static)
+  target_link_libraries(reader-writer2 parquet_static)
 endif()
--- a/examples/low-level-api/reader-writer.cc
+++ b/examples/low-level-api/reader-writer.cc
@@ -26,7 +26,8 @@
  * This example describes writing and reading Parquet Files in C++ and serves as a
  * reference to the API.
  * The file contains all the physical data types supported by Parquet.
- * This example uses the RowGroupWriter API that supports writing RowGroups optimized for memory consumption
+ * This example uses the RowGroupWriter API that supports writing RowGroups optimized for
+ *memory consumption
  **/
 
 /* Parquet is a structured columnar file format
@@ -70,8 +71,7 @@ int main(int argc, char** argv) {
         parquet::ParquetFileWriter::Open(out_file, schema, props);
 
     // Append a RowGroup with a specific number of rows.
-    parquet::RowGroupWriter* rg_writer =
-        file_writer->AppendRowGroup();
+    parquet::RowGroupWriter* rg_writer = file_writer->AppendRowGroup();
 
     // Write the Bool column
     parquet::BoolWriter* bool_writer =
--- a/examples/low-level-api/reader-writer.cc
+++ b/examples/low-level-api/reader-writer.cc
@@ -123,7 +123,7 @@ int main(int argc, char** argv) {
 
     // Append a RowGroup with a specific number of rows.
     parquet::RowGroupWriter* rg_writer =
-        file_writer->AppendRowGroup(NUM_ROWS_PER_ROW_GROUP);
+        file_writer->AppendRowGroup();
 
     // Write the Bool column
     parquet::BoolWriter* bool_writer =
--- a/examples/low-level-api/reader-writer.cc
+++ b/examples/low-level-api/reader-writer.cc
@@ -18,19 +18,15 @@
 #include <cassert>
 #include <fstream>
 #include <iostream>
-#include <list>
 #include <memory>
 
-#include <arrow/io/file.h>
-#include <arrow/util/logging.h>
-
-#include <parquet/api/reader.h>
-#include <parquet/api/writer.h>
+#include <reader_writer.h>
 
 /*
  * This example describes writing and reading Parquet Files in C++ and serves as a
  * reference to the API.
  * The file contains all the physical data types supported by Parquet.
+ * This example uses the RowGroupWriter API that supports writing RowGroups optimized for memory consumption
  **/
 
 /* Parquet is a structured columnar file format
@@ -46,55 +42,7 @@
  **/
 
 constexpr int NUM_ROWS_PER_ROW_GROUP = 500;
-constexpr int FIXED_LENGTH = 10;
 const char PARQUET_FILENAME[] = "parquet_cpp_example.parquet";
-
-using parquet::Repetition;
-using parquet::Type;
-using parquet::LogicalType;
-using parquet::schema::PrimitiveNode;
-using parquet::schema::GroupNode;
-
-static std::shared_ptr<GroupNode> SetupSchema() {
-  parquet::schema::NodeVector fields;
-  // Create a primitive node named 'boolean_field' with type:BOOLEAN,
-  // repetition:REQUIRED
-  fields.push_back(PrimitiveNode::Make("boolean_field", Repetition::REQUIRED,
-                                       Type::BOOLEAN, LogicalType::NONE));
-
-  // Create a primitive node named 'int32_field' with type:INT32, repetition:REQUIRED,
-  // logical type:TIME_MILLIS
-  fields.push_back(PrimitiveNode::Make("int32_field", Repetition::REQUIRED, Type::INT32,
-                                       LogicalType::TIME_MILLIS));
-
-  // Create a primitive node named 'int64_field' with type:INT64, repetition:REPEATED
-  fields.push_back(PrimitiveNode::Make("int64_field", Repetition::REPEATED, Type::INT64,
-                                       LogicalType::NONE));
-
-  fields.push_back(PrimitiveNode::Make("int96_field", Repetition::REQUIRED, Type::INT96,
-                                       LogicalType::NONE));
-
-  fields.push_back(PrimitiveNode::Make("float_field", Repetition::REQUIRED, Type::FLOAT,
-                                       LogicalType::NONE));
-
-  fields.push_back(PrimitiveNode::Make("double_field", Repetition::REQUIRED, Type::DOUBLE,
-                                       LogicalType::NONE));
-
-  // Create a primitive node named 'ba_field' with type:BYTE_ARRAY, repetition:OPTIONAL
-  fields.push_back(PrimitiveNode::Make("ba_field", Repetition::OPTIONAL, Type::BYTE_ARRAY,
-                                       LogicalType::NONE));
-
-  // Create a primitive node named 'flba_field' with type:FIXED_LEN_BYTE_ARRAY,
-  // repetition:REQUIRED, field_length = FIXED_LENGTH
-  fields.push_back(PrimitiveNode::Make("flba_field", Repetition::REQUIRED,
-                                       Type::FIXED_LEN_BYTE_ARRAY, LogicalType::NONE,
-                                       FIXED_LENGTH));
-
-  // Create a GroupNode named 'schema' using the primitive nodes defined above
-  // This GroupNode is the root node of the schema tree
-  return std::static_pointer_cast<GroupNode>(
-      GroupNode::Make("schema", Repetition::REQUIRED, fields));
-}
 
 int main(int argc, char** argv) {
   /**********************************************************************************
--- a/examples/low-level-api/reader-writer2.cc
+++ b/examples/low-level-api/reader-writer2.cc
@@ -1,0 +1,482 @@
+// Licensed to the Apache Software Foundation (ASF) under one
+// or more contributor license agreements. See the NOTICE file
+// distributed with this work for additional information
+// regarding copyright ownership. The ASF licenses this file
+// to you under the Apache License, Version 2.0 (the
+// "License"); you may not use this file except in compliance
+// with the License. You may obtain a copy of the License at
+//
+// http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing,
+// software distributed under the License is distributed on an
+// "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+// KIND, either express or implied. See the License for the
+// specific language governing permissions and limitations
+// under the License.
+
+#include <cassert>
+#include <fstream>
+#include <iostream>
+#include <list>
+#include <memory>
+
+#include <arrow/io/file.h>
+#include <arrow/util/logging.h>
+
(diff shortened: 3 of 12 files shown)


examples/low-level-api/reader-writer2.cc
```diff
@@ -0,0 +1,482 @@
...
+// software distributed under the License is distributed on an
+// "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+// KIND, either express or implied. See the License for the
+// specific language governing permissions and limitations
+// under the License.
+
+#include <cassert>
+#include <fstream>
+#include <iostream>
+#include <list>
```
jamesclampffer: nit: Doesn't look like you're using anything from list in here.  Maybe I am missing something implicit.

src/parquet/file_writer.h
```diff
@@ -79,6 +83,44 @@ class PARQUET_EXPORT RowGroupWriter {
   std::unique_ptr<Contents> contents_;
 };
 
+// RowGroupWriter implementation that allows bounding a RowGroup size to a certain value
+// Columns can be written in an order
+// All the values are compressed and stored in memory
+// Values are written to the file on Close
+
+class PARQUET_EXPORT RowGroupWriter2 {
```
jamesclampffer: A name like "FixedSizeRowGroupWriter", really anything other than <something>2, would be a little more intuitive in the public API.

src/parquet/file_writer.h
```diff
@@ -79,6 +83,44 @@ class PARQUET_EXPORT RowGroupWriter {
...
+ public:
+  // Forward declare a virtual class 'Contents' to aid dependency injection and more
+  // easily create test fixtures
+  // An implementation of the Contents class is defined in the .cc file
+  struct Contents {
+    virtual ~Contents() = default;
+    virtual int num_columns() const = 0;
+    virtual int64_t num_rows() const = 0;
+    virtual int64_t current_compressed_bytes() const = 0;
+    virtual ColumnWriter* get_column(int i) const = 0;
```
jamesclampffer: If the ColumnWriter can indirectly touch the RowGroupWriter's internal state it'd be best to return a `const ColumnWriter*` from the const accessor.

src/parquet/file_writer.h
```diff
@@ -102,6 +144,8 @@ class PARQUET_EXPORT ParquetFileWriter {
 
     virtual RowGroupWriter* AppendRowGroup() = 0;
 
+    virtual RowGroupWriter2* AppendRowGroup2() = 0;
+
```
jamesclampffer: Consider a subclass for writing fixed size rowgroups that overrides `AppendRowGroup` or take a param in the `ParquetFileWriter` ctor that determines how it wants to pack rowgroups.  That way whatever client code is pushing tuples in doesn't have to care about the on-disk layout after the writer is created.

src/parquet/file_writer.h
AnatoliShein: Agreed about the name. Something like ContinuousRowGroupWriter might work a bit better though since we now allow writing to each column multiple times, and this way suffix "2" can be replaced in all other locations with something like "_cont". "Repeated" might work too.

examples/low-level-api/reader-writer2.cc
wesm: Could be result of copy-paste, IWYU could confirm

examples/low-level-api/reader-writer2.cc
```diff
@@ -0,0 +1,480 @@
...
+    }
+  } catch (const std::exception& e) {
+    std::cerr << "Parquet read error: " << e.what() << std::endl;
+    return -1;
+  }
+
+  std::cout << "Parquet Writing and Reading Complete" << std::endl;
+
+  return 0;
+}
```
wesm: Is there a way to reuse code between this file and the other read/write example?

src/parquet/column_writer.cc
```diff
@@ -294,6 +295,8 @@ ColumnWriter::ColumnWriter(ColumnChunkMetaDataBuilder* metadata,
       num_buffered_encoded_values_(0),
       rows_written_(0),
       total_bytes_written_(0),
+      current_compressed_bytes_(0),
+      flush_on_close_(flush_on_close),
```
wesm: Would it be better to make this part of `WriterProperties`?

src/parquet/column_writer.cc
```diff
@@ -425,10 +430,20 @@ void ColumnWriter::WriteDataPage(const CompressedDataPage& page) {
...
+      // Important that we buffer all pages for RowGroupWriter2
+      if (num_buffered_values_ > 0) {
+        AddDataPage();
+      }
+    } else if (has_dictionary_ && !fallback_) {
+      flush_on_close_ = false;
       WriteDictionaryPage();
     }
 
+    flush_on_close_ = false;
```
wesm: This seems a tad hacky to me; I think there is a missing use of abstraction here which is the management of pages to be written to the file. The Parquet Java library has notions that provide for multiple implementations of this:

https://github.com/apache/parquet-mr/tree/master/parquet-column/src/main/java/org/apache/parquet/column/page

We don't need to necessarily do it in this patch, but using the existing PageWriter (or something similar if what is there now doesn't have the right features) abstraction with APIs to provide for

a) an implementation that accumulates pages in memory
b) an implementation that writes pages immediately

At some point we will want to make page writes asynchronous for better write performance so this refactoring will probably be desired at some point anyway

src/parquet/column_writer.cc
```diff
@@ -519,7 +537,13 @@ void TypedColumnWriter<Type>::WriteDictionaryPage() {
 
   DictionaryPage page(buffer, dict_encoder->num_entries(),
                       properties_->dictionary_index_encoding());
-  total_bytes_written_ += pager_->WriteDictionaryPage(page);
+
+  if (flush_on_close_) {
+    current_compressed_bytes_ += page.size() + sizeof(format::DictionaryPageHeader);
+    saved_dictionary_page_.push_back(std::move(page));
```
wesm: This aligns with my comments above

src/parquet/column_writer.h
```diff
@@ -117,6 +118,10 @@ class PARQUET_EXPORT ColumnWriter {
 
   int64_t rows_written() const { return rows_written_; }
 
+  // Only considers the size of the compressed pages + page header
+  // Some values might be still buffered an not written to a page yet
+  int64_t current_compressed_bytes() const { return current_compressed_bytes_; }
```
wesm: A couple notes

* This value is only being updated when `flush_on_close` is true, which might be surprising
* The purpose of this method from the comment is only clear with additional context about other parts of the implementation. Should this say "Total size of column chunk serialized so far" -- perhaps this should give this information regardless of whether we are flushing immediately or at the end

src/parquet/column_writer.h
```diff
@@ -209,6 +220,8 @@ class PARQUET_EXPORT ColumnWriter {
 
   std::vector<CompressedDataPage> data_pages_;
 
+  std::vector<DictionaryPage> saved_dictionary_page_;
```
wesm: Should the pages all be kept in the same collection?

src/parquet/file_writer.h
wesm: Agree that `RowGroupWriter2` is not a good name. 

Why do we need a new class? It's not clear to me that it's needed

src/parquet/file_writer.h
```diff
@@ -79,6 +83,44 @@ class PARQUET_EXPORT RowGroupWriter {
...
+  ColumnWriter* get_column(int i) const;
+
+  void Close();
+
+  int num_columns() const;
+
+  int64_t num_rows() const;
+
+  // Only considers the size of the compressed pages + page header in all the columns
+  // Some values might be still buffered an not written to a page yet
```
wesm: See note above re: this comment

src/parquet/file_writer.h
wesm: Agreed. I think these details should be handled in the `WriterProperties` and the API of `RowGroupWriter` should be able to indicate to the caller whether it is "safe" to continue writing, or whether the current row group needs to be terminated and flushed

src/parquet/types.h
```diff
@@ -114,13 +114,9 @@ struct Compression {
 };
 
 struct Encryption {
-  enum type {
-    AES_GCM_V1 = 0,
-    AES_GCM_CTR_V1 = 1
-  };
+  enum type { AES_GCM_V1 = 0, AES_GCM_CTR_V1 = 1 };
```
wesm: clang-format artifact?

src/parquet/file_writer.h
majetideepak: The current `RowGroupWriter` class has `virtual ColumnWriter* NextColumn() = 0;` which does not fit well with the use case here which is to be able to `write to a column chunk any number of times (can cycle around all column chunks any number of times) and also any order`. 
The current `RowGroupWriter` requires us to write a column chunk at one go and move to the immediate next column. 
We need the following external API to support this use case.
`virtual ColumnWriter* get_column(i) = 0;`
The new `RowGroupWriter2` has different semantics in this aspect.

majetideepak: I am not tied to this name. Will change to `FixedSizeRowGroupWriter`

src/parquet/file_writer.h
majetideepak: will do!

src/parquet/file_writer.h
majetideepak: This will restrict the clients from being able to loosely or tightly bound a RowGroup size.
Example: Clients might want to write a row at a time or a batch at a time or do both adaptively to achieve their target RowGroup size.

examples/low-level-api/reader-writer2.cc
majetideepak: The two examples are quite different. The other example writes a single row group with a predetermined number of rows. I can move the schema creation part to a common header.

src/parquet/column_writer.cc
majetideepak: It will depend on what RowGroupWriter API we provide to the users. If parquet-cpp takes the responsibility of bounding RowGroup sizes (which I think is not in its scope based on the example above), then this can be.

src/parquet/column_writer.cc
majetideepak: The PageWriter class is responsible for writing compressed data pages out. But the goal of either writing compressed Pages or accumulating them is more efficient if its part of the `ColumnWriter`.

src/parquet/column_writer.h
majetideepak: Good point! I had `total_bytes_written_` in mind for the immediate flushing case.
We also keep pages in memory when the dictionary values are being encoded. Probably I will specify that the pages are only held in memory.

src/parquet/column_writer.h
majetideepak: Will do!

src/parquet/types.h
majetideepak: yes!

examples/low-level-api/reader-writer2.cc
majetideepak: Will check and fix!

src/parquet/file_writer.cc
```diff
@@ -125,21 +191,43 @@ class RowGroupSerializer : public RowGroupWriter::Contents {
...
-    if (num_rows_ < 0) {
-      num_rows_ = current_rows;
-      metadata_->set_num_rows(current_rows);
-    } else if (num_rows_ != current_rows) {
-      std::stringstream ss;
-      ss << "Column " << current_column_index_ << " had " << current_rows
-         << " while previous column had " << num_rows_;
-      throw ParquetException(ss.str());
+    // verify when only one column is written at a time
+    if (!row_group_by_size_ && column_writers_.size() > 0 && column_writers_[0]) {
```
majetideepak: This if-else is one of the reasons to use a new `RowGroupWriter` class in my previous version of this.

src/parquet/file_writer.cc
```diff
@@ -96,23 +125,60 @@ class RowGroupSerializer : public RowGroupWriter::Contents {
...
         PageWriter::Open(sink_, properties_->compression(column_descr->path()), col_meta,
                          properties_->memory_pool());
-    current_column_writer_ = ColumnWriter::Make(col_meta, std::move(pager), properties_);
-    return current_column_writer_.get();
+    column_writers_[0] = ColumnWriter::Make(col_meta, std::move(pager), properties_);
+    return column_writers_[0].get();
+  }
+
+  ColumnWriter* get_column(int i) override {
+    if (!row_group_by_size_) {
```
majetideepak: This restriction is the other misfit in using a single `RowGroupWriter` API

majetideepak:
@wesm, @xhochy  I made changes based on the feedback.  I can quickly add unit tests if we agree on the API.

examples/low-level-api/CMakeLists.txt
```diff
@@ -15,7 +15,11 @@
 # specific language governing permissions and limitations
 # under the License.
 
+include_directories(SYSTEM . )
```
xhochy: Please use https://cmake.org/cmake/help/v3.3/command/target_include_directories.html so that this does not spread to the main library.

src/parquet/file_writer.h
```diff
@@ -50,8 +54,16 @@ class PARQUET_EXPORT RowGroupWriter {
     virtual int64_t num_rows() const = 0;
 
     virtual ColumnWriter* NextColumn() = 0;
+    // to be used only when row_group_by_size = true
+    virtual ColumnWriter* get_column(int i) = 0;
```
xhochy: I would just name this `column()`

src/parquet/file_writer.h
```diff
@@ -69,11 +81,17 @@ class PARQUET_EXPORT RowGroupWriter {
 
   int num_columns() const;
 
+  // to be used only when row_group_by_size = true
+  ColumnWriter* get_column(int i);
```
xhochy: `column()`

examples/low-level-api/CMakeLists.txt
wesm: +1. See example here: https://github.com/apache/arrow/blob/master/cpp/src/arrow/python/CMakeLists.txt#L121

examples/low-level-api/reader-writer.cc
```diff
@@ -18,19 +18,15 @@
...
-
-#include <parquet/api/reader.h>
-#include <parquet/api/writer.h>
+#include <reader_writer.h>
 
 /*
  * This example describes writing and reading Parquet Files in C++ and serves as a
  * reference to the API.
  * The file contains all the physical data types supported by Parquet.
+ * This example uses the RowGroupWriter API that supports writing RowGroups optimized for memory consumption
```
wesm: Seems maybe clang-format isn't hitting this file

src/parquet/file_writer.h
```diff
@@ -100,7 +118,7 @@ class PARQUET_EXPORT ParquetFileWriter {
     /// \note Deprecated since 1.3.0
     RowGroupWriter* AppendRowGroup(int64_t num_rows);
 
-    virtual RowGroupWriter* AppendRowGroup() = 0;
+    virtual RowGroupWriter* AppendRowGroup(bool row_group_by_size = false) = 0;
```
wesm: Adding a new method `AppendFixedSizeRowGroup()` would result in more readable code than `AppendRowGroup(true)`

src/parquet/file_writer.h
```diff
@@ -50,8 +54,16 @@ class PARQUET_EXPORT RowGroupWriter {
     virtual int64_t num_rows() const = 0;
 
     virtual ColumnWriter* NextColumn() = 0;
+    // to be used only when row_group_by_size = true
```
wesm: Clarify in a comment for `NextColumn()` that `ColumnWriter*` objects become invalid when `NextColumn` is called unless writing fixed size row groups

src/parquet/file_writer.h
```diff
@@ -39,6 +39,10 @@ class GroupNode;
 
 }  // namespace schema
 
+// RowGroupWriter implementation that optimizes memory requirement
+// All columns must be written one after the other
+// Writing to a new column prevents modification to the previous column
```
wesm: Add `\brief` here and use `///` so these comments will show up in doxygen. We can make sure the Parquet docs show up well in http://arrow.apache.org/docs/cpp/ after the merge

src/parquet/file_writer.cc
```diff
@@ -47,47 +49,74 @@ void RowGroupWriter::Close() {
...
+  return contents_->total_bytes_written();
+}
+
 int RowGroupWriter::current_column() { return contents_->current_column(); }
 
 int RowGroupWriter::num_columns() const { return contents_->num_columns(); }
 
 int64_t RowGroupWriter::num_rows() const { return contents_->num_rows(); }
 
+inline void throwRowsMisMatchError(int col, int64_t prev, int64_t curr) {
```
wesm: `ThrowRowsMismatchError`

wesm:
I suggest maybe calling this `AppendBufferedRowGroup` or something

PR #487, opened 2018-08-16
PARQUET-1382: [C++] Prepare for arrow::test namespace removal

pitrou:
Also tried to make Travis-CI builds faster.


pitrou:
There are a lot of warnings now in the OS X build, such as:
```
ld: warning: direct access in apache::thrift::protocol::TCompactProtocolFactoryT<apache::thrift::transport::TMemoryBuffer>::getProtocol(std::__1::shared_ptr<apache::thrift::transport::TTransport>) to global weak symbol typeinfo for apache::thrift::transport::TTransport means the weak symbol cannot be overridden at runtime. This was likely caused by different translation units being compiled with different visibility settings.
```

Also, ListLargeRecords failed here @wesm :
https://travis-ci.org/apache/parquet-cpp/jobs/416711723#L5106

pitrou:
apt is very slow on Travis-CI today :/

pitrou:
The AppVeyor toolchain build apparently failed linking with brotli:
https://ci.appveyor.com/project/ApacheSoftwareFoundation/parquet-cpp/build/1.0.762

(is brotli actually useful? There are better compression options nowadays...)

wesm:
Brotli is in the format. We might drop it at some point or make it optional in the build

pitrou:
Ok, our AppVeyor CI script was forcing an old Brotli version, so I'm trying to bump it up to see whether that fixes things.

pitrou:
Ok, so the AppVeyor build is fixed now. The Travis-CI build still has the OS X failure. What should we do? Do you know how to investigate that bug?

wesm:
I have about 2 hours to spend on it today, I'll write in PARQUET-1308 as I make progress figuring it out

wesm:
Just rebased this (there were some small conflicts, hopefully I fixed them correctly). Should hopefully get a green build now

PR #488, opened 2018-08-17
PARQUET-1384: fix clang build error for bloom_filter-test.cc

cjjnjust:


wesm:
@cjjnjust I notice that you are developing in the master branch. In addition to making things harder for you (hard to work on multiple patches at once), it makes things a bit harder for PR maintainers (git requires slightly different commands to be able to track branches having the same name as some of your own brances). Could you use named branches for your next patches?

wesm:
thanks!

PR #489, opened 2018-08-17
PARQUET-1385: Do not run TestBloomFilter.FPPTest when valgrind is in use

wesm:
This test will still be run in other entries of Travis CI where valgrind is not being used


pitrou:
The number of iterations is adjusted in #487, so I don't think this PR is necessary.

wesm:
OK, thanks. Closing

PR #491, opened 2018-08-17
PARQUET-1369: Disregard column sort order if statistics max/min are equal

rgruener:


majetideepak:
+1 LGTM

rgruener:
Fixed, sorry I forgot to run `make lint` before submitting

src/parquet/metadata.h
```diff
@@ -84,6 +85,7 @@ class ApplicationVersion {
 
   // Checks if the Version has the correct statistics for a given column
   bool HasCorrectStatistics(Type::type primitive,
+                            const parquet::format::Statistics& statistics,
```
rgruener: Since this is actually a thrift definition, it is not allowed to be included here apparently. I can move the check for `max_equals_min` out of this but that doesnt seem like the best thing organizationally. Any suggestions here?

wesm: You may have to create an accessor class to hide the Thrift structures from the public API. That's what we've done with the rest of the Thrift metadata

rgruener: I got around it by not passing the thrift structure as an argument anymore but just whether max/min are equivalent. This seems not ideal since the tests dont cover extracting this value and this logic seems like a good fit for being within `HasCorrectStatistics` so I will probably look into an accessor class. Is there a good example you can point me to (will be faster than me scouring the codebase looking for a good one)?

wesm: You'll want to use the private implementation ("PIMPL") pattern. Here's an example of one: https://github.com/apache/arrow/blob/master/cpp/src/arrow/memory_pool.h#L108. There are others in this codebase, too

src/parquet/metadata.cc
```diff
@@ -124,8 +155,13 @@ class ColumnChunkMetaData::ColumnChunkMetaDataImpl {
   //    Eg: UTF8
   inline bool is_stats_set() const {
     DCHECK(writer_version_ != nullptr);
-    return column_->meta_data.__isset.statistics &&
-           writer_version_->HasCorrectStatistics(type(), descr_->sort_order());
+    if (!column_->meta_data.__isset.statistics) {
+      return false;
+    }
+    std::unique_ptr<MinMaxStatistics> statistics = MinMaxStatistics::Make(
+            reinterpret_cast<const uint8_t*>(&column_->meta_data.statistics));
```
majetideepak: You need to use `TypedRowGroupStatistics` to get type based equality. You can simply solve your problem by creating that object (via `MakeColumnStats()`) here.
You will have to re-organize `is_stats_set` and `statistics()` to avoid calling `MakeColumnStats()` twice.

src/parquet/metadata-test.cc
```diff
@@ -219,12 +220,31 @@ TEST(ApplicationVersion, Basics) {
...
+  ASSERT_TRUE(version1.HasCorrectStatistics(Type::BYTE_ARRAY, min_max_statistics,
+                                            SortOrder::SIGNED));
+  ASSERT_FALSE(version1.HasCorrectStatistics(Type::BYTE_ARRAY, min_max_statistics,
+                                             SortOrder::UNSIGNED));
+  ASSERT_TRUE(version3.HasCorrectStatistics(Type::FIXED_LEN_BYTE_ARRAY,
+                                            min_max_statistics, SortOrder::SIGNED));
+
+  // Check that the old statistics are correct if min and max are the same
+  // regardless of sort order
+  statistics.__set_min("a");
```
majetideepak: please add tests for other types as well.

src/parquet/metadata.cc
rgruener: @majetideepak So I implemented that but that complicated things since for the new statistics it always says min/max is set (see https://github.com/apache/parquet-cpp/blob/master/src/parquet/metadata.cc#L57)

This makes it seem like min and max are equivalent when they are not set (since they will both be empty strings)

rgruener:
@majetideepak So I implemented that but that complicated things since for the new statistics it always says min/max is set (see https://github.com/apache/parquet-cpp/blob/master/src/parquet/metadata.cc#L57)

This makes it seem like min and max are equivalent when they are not set (since they will both be empty strings). Any suggestions on resolving that? I don't really see a clean way to handle this while still using that class without doing a bunch of refactoring

majetideepak:
@rgruener  Can't we extend the following check to see if min/max are set as well?
```
    if (!column_->meta_data.__isset.statistics) {
      return false;
    }
```

rgruener:
Do we want to do that? Statistics would still be valid if min/max is not set but the num nulls is for example

majetideepak:
If it is possible that the stats are set but not the new `min_value` and `max_value`, I think we should change that line ( https://github.com/apache/parquet-cpp/blob/master/src/parquet/metadata.cc#L57) from `true` to `metadata.statistics.__isset.max_value || metadata.statistics.__isset.min_value`. I can't recall why that is set to `true` always.

wesm:
This PR needs to be moved to https://github.com/apache/arrow -- please let us know if you require assistance

wesm:
Can I help with moving this patch over?

wesm:
Here's a ported commit, open a new PR with this https://github.com/wesm/arrow/commit/676f792b6ee6c0eb2bad0637c6d6a304813e1324

rgruener:
This has been moved to https://github.com/apache/arrow/pull/2655

PR #492, opened 2018-08-19
PARQUET-1392: Read multiple RowGroups at once into an Arrow table

xhochy:
Decided to go with the more simplistic approach and only introduce a convenience API for now. Once is merged, I'll do some work that at least primitive arrays are read into a single Array in this path.

### DIFF
--- a/src/parquet/arrow/reader.cc
+++ b/src/parquet/arrow/reader.cc
@@ -571,12 +571,10 @@ Status FileReader::Impl::ReadRowGroups(const std::vector<int>& row_groups,
                                        std::shared_ptr<Table>* table) {
   // TODO(PARQUET-1393): Modify the record readers to already read this into a single,
   // continuous array.
-  std::vector<std::shared_ptr<Table>> tables;
+  std::vector<std::shared_ptr<Table>> tables(row_groups.size(), nullptr);
 
   for (size_t i = 0; i < row_groups.size(); ++i) {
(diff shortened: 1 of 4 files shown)


src/parquet/arrow/reader.cc
```diff
@@ -562,6 +566,31 @@ Status FileReader::Impl::ReadTable(std::shared_ptr<Table>* table) {
...
+Status FileReader::Impl::ReadRowGroups(const std::vector<int>& row_groups,
+                                       const std::vector<int>& indices,
+                                       std::shared_ptr<Table>* table) {
+  // TODO(PARQUET-1393): Modify the record readers to already read this into a single,
+  // continuous array.
+  std::vector<std::shared_ptr<Table>> tables;
+
+  for (size_t i = 0; i < row_groups.size(); ++i) {
+    std::shared_ptr<Table> rg_table;
+    RETURN_NOT_OK(ReadRowGroup(row_groups[i], indices, &rg_table));
```
majetideepak: a small optimization here is to use
`std::vector<std::shared_ptr<Table>> tables(row_group.size(), nullptr)` above and
use `&tables[i]` here. Afaik certain tools like Spark generate files with many small RowGroups.

majetideepak: I just saw your TODO. So this change might be less relevant.

xhochy: Still, added it as it saves some cycles.

PR #495, opened 2018-08-23
PARQUET-1160: [C++] Implement BYTE_ARRAY-backed Decimal reads

thaining:
This change enables support for DECIMALs backed by BYTE_ARRAYs
on disk.  It does this by creating a TransferFunctor routine that
transforms a ByteArrayType to an ::arrow::Decimal128Type.

The routine does this by:
1. Creating an arrow::BinaryArray from the RecordReader's builder
2. Allocating a buffer for the arrow::Decimal128Array
3. Converting the big-endian bytes in each BinaryArray entry to two integers
   representing the high and low bits of each decimal value.


src/parquet/arrow/reader.cc
```diff
@@ -1220,6 +1220,68 @@ struct TransferFunctor<::arrow::Decimal128Type, FLBAType> {
...
+        if ((record_len < 0) || (record_len > type_length)) {
+            return Status::Invalid("Invalid BYTE_ARRAY size");
+        }
+
+        auto out_ptr_view = reinterpret_cast<uint64_t*>(out_ptr);
+        out_ptr_view[0] = 0;
+        out_ptr_view[1] = 0;
+
+
+        if ((null_count > 0) && !binary_array.IsNull(i)) {
```
majetideepak: Nit: The statements in the two `if blocks` are the same. Should we combine them with an `||`?

thaining: I think I was being too literal in trying to model the structure of  TransferFunctor<::arrow::Decimal128Type, FLBAType>, since that has a for loop when (null_count > 0) and another where it doesn't. 

You are correct, an || is more compact.  I will update.  Thank you.

xhochy:
@thaining Travis reports some code style issues you could probably get rid of by running `make format`. Can you take care of that?

thaining:
Fixed the indenting and column width changes in reader.cc suggested by 'make format' with clang-format 5.0.2.   For arrow-reader-writer-test.cc, the program only suggested reordering of the using statements, which I had not touched.

thaining:
@xhochy, a unit test conforming to the others for Spark was also included in the last batch of commits.

wesm:
This PR needs to be moved to https://github.com/apache/arrow -- please let us know if you require assistance

thaining:
> This PR needs to be moved to https://github.com/apache/arrow -- please let us know if you require assistance

Looks like I do require assistance.  The change squashes down to a clean commit, but I do not have write access to the repo.  Can you advise me on next steps?  Thanks.

Ted

wesm:
I rolled up your changes as this commit https://github.com/wesm/arrow/commit/4f68fc3e46920f599ff6e5a0f37af5eb8cdc24e7. You can open a PR starting from there

thaining:
Thanks for the help.   I was a little confused, and that gave me the push I needed.  I've created PR https://github.com/apache/arrow/pull/2646

